Issue #1, opened 2018-03-17
Implement ADT which specifies TOML


ChShersh:
According to this document:

* https://github.com/toml-lang/toml

PR #2, opened 2018-03-17
[#1] Implement basic version of TOML

ChShersh:
This PR implements basic AST which represents TOML configuration.

It's not type-safe enough. Probably some type magic should be used, like, `GADT` and `DataKinds`. But this will make parsing harder. And also it will make data types much scarier... But I think at some point of time this should be done. Maybe even in this PR.

For now I want to discuss whether this data type is enough to represent whole TOML so it's also should be convenient to work with this type.

I recommend to build documentation locally and see how it looks like.


src/Toml/Type.hs
```diff
@@ -0,0 +1,146 @@
...
+    { unTableId :: NonEmpty Text
+    } deriving (Eq, Ord, Hashable)
+
+-- TODO: describe how some TOML document will look like with this type
+{- | Represents TOML configuration value. -}
+data TOML = TOML
+    { tomlPairs       :: HashMap Key     Value
+    , tomlTables      :: HashMap TableId TOML
+    , tomlTableArrays :: HashMap TableId (NonEmpty TOML)
+    -- TODO: I don't really like the above structure, probably something better should be used
```
vrom911: That's good point, but I don't have anything in mind for the moment..

ChShersh: The first version of parser won't support array of tables anyways, so we can safely comment this line. And maybe we will come up with better design later...

PR #6, opened 2018-03-29
[#4] Add basic pretty-printer

### LINKED ISSUE #4: Implement basic pretty-printer
Implement `TOML -> Text` function which renders TOML data type into textual representation.

vrom911:
Resolves #4 
I've created `playground` where now you can see the result of printing some example `toml` file. You can `stack exec play-tomland` for that.
Also I didn't touch `tomlTableArrays` for the moment as was discussed.

### DIFF
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -1,0 +1,46 @@
+module Main where
+
+import Data.HashMap.Strict (HashMap, fromList)
+import Data.Time (fromGregorian)
+
+import Toml.Printer (toml2Text)
+import Toml.Type (DateTime (..), Key (..), TOML (..), TableId (..), Value (..))
+
+import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.Text as Text
+
+main :: IO ()
+main = putStrLn $ Text.unpack $ toml2Text myToml
+
+
+myToml :: TOML
+myToml = TOML (fromList
+    [ (Key "a", Bool True)
+    , (Key "list", Array [String "one", String "two"])
+    , (Key "time", Array [Date $ Day (fromGregorian 2018 3 29)])
+    ] ) myInnerToml mempty
+
+myInnerToml :: HashMap TableId TOML
+myInnerToml = fromList $ [ ( TableId (NonEmpty.fromList ["table", "name", "1"])
+                           , TOML (fromList [ (Key "aInner", Int 1)
+                                            , (Key "listInner", Array [Bool True, Bool False])
+                                            ]) myInnerInnerToml mempty
+                           )
+                         , ( TableId (NonEmpty.fromList ["table", "name", "2"])
+                           , TOML (fromList [ (Key "2Inner", Int 42)
+                                            ]) mempty mempty
+                           )
+                         ]
+
+myInnerInnerToml :: HashMap TableId TOML
+myInnerInnerToml =
+    fromList $ [ ( TableId (NonEmpty.fromList ["table", "name", "1", "1"])
+                 , TOML (fromList [ (Key "aInner", Int 1)
+                                  , (Key "listInner", Array [Bool True, Bool False])
+                                  ]) mempty mempty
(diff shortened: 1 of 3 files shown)


src/Toml/Printer.hs
```diff
@@ -0,0 +1,56 @@
...
+import Toml.Type (DateTime (..), Key (..), TOML (..), TableId (..), Value (..))
+
+import qualified Data.HashMap.Strict as HashMap
+import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.Text as Text
+
+tab :: Int -> Text
+tab n = Text.cons '\n' (Text.replicate (2*n) " ")
+
+toml2Text :: TOML -> Text
```
ChShersh: Since this function is exported let's add Haddock documentation to it. Maybe with some small example from playground... What do you think?

examples/Playground.hs
```diff
@@ -0,0 +1,46 @@
...
+    ] ) myInnerToml mempty
+
+myInnerToml :: HashMap TableId TOML
+myInnerToml = fromList $ [ ( TableId (NonEmpty.fromList ["table", "name", "1"])
+                           , TOML (fromList [ (Key "aInner", Int 1)
+                                            , (Key "listInner", Array [Bool True, Bool False])
+                                            ]) myInnerInnerToml mempty
+                           )
+                         , ( TableId (NonEmpty.fromList ["table", "name", "2"])
+                           , TOML (fromList [ (Key "2Inner", Int 42)
```
ChShersh: I guess, it's not really convenient to you to specify TOML in a such way manually :disappointed: I will create issue for this, don't worry! We will have convenient EDSL for specifying TOML manually.

examples/Playground.hs
```diff
@@ -0,0 +1,46 @@
...
+import Data.Time (fromGregorian)
+
+import Toml.Printer (toml2Text)
+import Toml.Type (DateTime (..), Key (..), TOML (..), TableId (..), Value (..))
+
+import qualified Data.List.NonEmpty as NonEmpty
+import qualified Data.Text as Text
+
+main :: IO ()
+main = putStrLn $ Text.unpack $ toml2Text myToml
```
ChShersh: You can use `putStrLn` from `Data.Text.IO` module to avoid `unpack`.

src/Toml/Printer.hs
```diff
@@ -0,0 +1,56 @@
...
+
+toml2TextInd :: Int -> TOML -> Text
+toml2TextInd i TOML{..} = keyValue2Text i tomlPairs  <> "\n"
+                       <> tables2Text   i tomlTables
+
+keyValue2Text :: Int -> HashMap Key Value -> Text
+keyValue2Text i = Text.concat . map kvText . HashMap.toList
+  where
+    kvText :: (Key, Value) -> Text
+    kvText (Key{..}, v) = tab i <> unKey <> " = " <> valText v
```
ChShersh: Very minor, but I find `(Key k, v)` prettier than using `unKey` later.

src/Toml/Printer.hs
```diff
@@ -1,3 +1,5 @@
+{- | Contains functional for pretty printing @toml@ types. -}
```
ChShersh: Maybe `functions` instead of `functional`?

src/Toml/Printer.hs
```diff
@@ -12,12 +14,43 @@ import qualified Data.HashMap.Strict as HashMap
...
+[example.owner]
+  name = "Kowainik"
+
+@
+-}
 toml2Text :: TOML -> Text
 toml2Text = toml2TextInd 0
 
+-- | Converts 'TOML' into 'Text' with the given indent.
 toml2TextInd :: Int -> TOML -> Text
```
ChShersh: I guess we can export this function as well. In future we will have `PrintOptions` data type with `defaultOptions`.

Issue #7, opened 2018-03-29
Implement type-safe version of TOML ast


ChShersh:
So, currently TOML `Value` data type allows to have values of different type inside `Array`. I propose to solve this problem by introducing new `TValue` data type (or renaming current `Value` to `UValue` aka _untyped value) with GADT and `DataKinds` feature. Basically:

```haskell
data TomlType = TBool | ...

data Value (t :: TomlType) where
    Bool :: Bool -> Value TBool
    ...
```

Basically, approach starting from this slide should be used:

* http://slides.com/fp-ctd/lecture-15#/8

I guess we should have `UValue` and `typeCheck :: UValue -> Maybe Value` function. Not sure though...

Also, several problems need to be solved. For example, this type also should have typed and untyped version but it's not obvious how to modify it... Let's discuss and see whether it makes sense!

```haskell
data TOML = TOML
    { tomlPairs       :: HashMap Key     Value
    , tomlTables      :: HashMap TableId TOML
```

PR #9, opened 2018-03-29
[#7] Implement type-safe Value

vrom911:


src/Toml/Type.hs
```diff
@@ -35,26 +43,29 @@ newtype TableId = TableId
...
 data TOML = TOML
-    { tomlPairs       :: HashMap Key     Value
+    { tomlPairs       :: HashMap Key     AnyValue
     , tomlTables      :: HashMap TableId TOML
     , tomlTableArrays :: HashMap TableId (NonEmpty TOML)
     -- TODO: I don't really like the above structure, probably something better should be used
     }
 
+-- Needed for GADT parameterization
+data TomlType = TBool | TInt | TFloat | TString | TDate | TArray
```
ChShersh: I came up with `TomlType` but personally I don't like this name.. Let's use something like `TValue` or `VType` or `ValueType` instead.

src/Toml/Type.hs
```diff
@@ -144,3 +167,32 @@ lt2 = 00:32:00.999999
...
+typeCheck (UString s) = Just $ AnyValue $ String s
+typeCheck (UDate d)   = Just $ AnyValue $ Date d
+typeCheck (UArray a)  = case a of
+    []     -> Just $ AnyValue $ Array []
+    (x:xs) -> do
+        AnyValue v <- typeCheck x
+        AnyValue . Array <$> checkElem v xs
+  where
+    checkElem :: Value t -> [UValue] -> Maybe [Value t]
+    checkElem x []     = Just [x]
```
ChShersh: Let's call this variable `v` instead of `x` to have consistent naming.

PR #28, opened 2018-04-07
[WIP] Basic parse print property test

vrom911:


src/Toml/Parser.hs
```diff
@@ -52,7 +52,7 @@ integer :: Parser Integer
 integer = L.signed sc (lexeme L.decimal)
 
 float :: Parser Double
-float = lexeme L.float
+float = L.signed sc $ lexeme L.float
```
ChShersh: :+1:

test/Test/Toml/Gen.hs
```diff
@@ -0,0 +1,107 @@
...
+
+-- TODO: Arrays and Date.
+-- | Generates random value of 'AnyValue' type.
+genAnyValue :: MonadGen m => m AnyValue
+genAnyValue = do
+  randB <- Gen.bool
+  randI <- toInteger <$> Gen.int (Range.constantBounded @Int)
+  randF <- Gen.double $ Range.constant @Double (-1000000.0) 1000000.0
+  randS <- Gen.text (Range.constant 0 256) Gen.alphaNum
+  Gen.element
```
ChShersh: I think it's better to use [`choice`](https://hackage.haskell.org/package/hedgehog-0.5.3/docs/Hedgehog-Gen.html#v:choice) combinator here. It can increase speed of tests. Because it looks to me that now this generator will generate every value when we need only one.

test/Test/Toml/Gen.hs
```diff
@@ -0,0 +1,107 @@
...
+genPiece = Piece <$> Gen.text (Range.constant 1 50) Gen.alphaNum
+
+genKey :: MonadGen m => m Key
+genKey = Key <$> Gen.nonEmpty (Range.constant 1 10) genPiece
+
+genKeyAnyValue :: MonadGen m => m (Key, AnyValue)
+genKeyAnyValue = do
+    key <- genKey
+    val <- genAnyValue
+    pure (key, val)
```
ChShersh: I think for such simple generators it's better to use shorter implementations with `liftA2` or similar.

test/Test/Toml/Parsing/Property.hs
```diff
@@ -0,0 +1,31 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE ExplicitForAll      #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE GADTs               #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications    #-}
```
ChShersh: Are all these extensions needed?

test/Test/Toml/Parsing/Property.hs
```diff
@@ -0,0 +1,31 @@
...
+import Toml.Parser (parse)
+import Toml.Printer (prettyToml)
+
+import Test.Toml.Gen (genToml)
+
+propertyTests :: [TestTree]
+propertyTests = [parsePrintTest]
+
+parsePrintTest :: TestTree
+parsePrintTest = testProperty "parse . prettyPrint == id" prop_parsePrint
```
ChShersh: Maybe we can write `testProperty bla-bla-bla` right in `propertyTests` list to make such files shorter? :thinking: 
What do you think? Also no need to invent extra names.

test/Test/Toml/PrefixTree/Property.hs
```diff
@@ -90,7 +38,7 @@ prop_InsertLookup =  property $ do
 
 prop_InsertInsert :: Property
 prop_InsertInsert =  property $ do
-    t <- forAll genPrefixMap
+    t <- forAll $ genPrefixMap genVal
```
ChShersh: Looks like you don't need to make `genPrefixMap` generator polymorphic since you call it only with `genVal` for now.

Issue #29, opened 2018-04-09
Implement our own 'PrefixTree'


ChShersh:
After thinking a little bit more on how to store `Key` components I came to conclusion that we really want to store component on each level of `Toml` data type instead of storing whole key.

Currently we have the following data types:

```haskell
newtype Key = Key { unKey :: NonEmpty Text }

data TOML = TOML
    { tomlPairs  :: HashMap Key AnyValue
    , tomlTables :: HashMap Key TOML
    }
```

So each level can contain whole key. This is not really convenient actually... Because if you want to insert some key into such map, you need to scan whole `HashMap` in order to split by greatest common prefix. This sad and not really convenient...

On the other hand, storing only single key component per level simplifies things a lot!

So we actually want to have something like this:

```haskell
data TOML = TOML
    { tomlPairs  :: HashMap Key AnyValue
    , tomlTables :: HashMap Text TOML  -- here 'Text' is only single component of 'Key'
    }
```

But with such scheme it's harder to implement pretty-printing for tables like this:

```toml
[foo.bar]
baz = 3
```

So we should analyze nested `Toml`s to not print table in this way:

```toml
[foo]
  [bar]  # we don't want this
    baz = 3
```

Also, during parsing we need to split every table key by component in insert them in this `HashMap` in some smart way. This is also undesirable. 

Currently I'm having hard problems with these issues: 
* #25
* #10 

So let's implement our own data structure! Basically what we want is to have something like monomorphic for `Key` `HashMap` but which joins nodes by common prefix.

I think that data structure can look like this:

```haskell
data PrefixTree a
    = Leaf (NonEmpty Text) a
    | Branch (NonEmpty Text) (PrefixTree a) (PrefixTree a)

type PrefixMap a = Maybe (PrefixTree a)
```

And we need to have the following basic functions and instances:

```haskell
single :: Key -> a -> PrefixTree a
insert :: Key -> a -> PrefixTree a -> PrefixTree a
lookup :: Key -> PrefixTree a -> Maybe a
delete :: Key -> PrefixTree a -> Maybe (PrefixTree a)

instance Semigroup (PrefixTree a)  -- or PartialSemigroup
```

ChShersh:
Okay, so this structure doesn't work:

```haskell
data PrefixTree a
    = Leaf (NonEmpty Text) a
    | Branch (NonEmpty Text) (PrefixTree a) (PrefixTree a)
```

### Problems

There're multiple reasons for this:

[...]

PR #30, opened 2018-04-20
Add basic PrefixTree

vrom911:
First attempts of `PrefixTree` (see #29 )

### DIFF
--- a/src/Toml/PrefixTree.hs
+++ b/src/Toml/PrefixTree.hs
@@ -1,9 +1,15 @@
 module Toml.PrefixTree
        ( PrefixTree (..)
-       , single
-       , insert
-       , lookup
-       , delete
+       , singleT
+       , insertT
+       , lookupT
+       , deleteT
+
+         -- * Types
+       , Piece (..)
+       , Key (..)
+       , Prefix
+       , KeysDiff (..)
        ) where
 
 import Prelude hiding (lookup)
@@ -30,12 +36,12 @@ type Prefix = Key
 -- | Map of layer names and corresponding 'PrefixTree's.
 type PrefixMap a = HashMap Piece (PrefixTree a)
 
--- | Data structure to represent @toml@.
+-- | Data structure to represent table tree for @toml@.
 data PrefixTree a
-    = Leaf Key a
-    | Branch { bcommonPref :: Prefix         -- ^ greatest common prefix
-             , bVal        :: (Maybe a)      -- ^ value by key = prefix
-             , bPrefixMap  :: (PrefixMap a)  -- ^ suffixes of prefix
+    = Leaf !Key !a
+    | Branch { bCommonPref :: !Prefix         -- ^ greatest common prefix
+             , bVal        :: !(Maybe a)      -- ^ value by key = prefix
+             , bPrefixMap  :: !(PrefixMap a)  -- ^ suffixes of prefix
              }
 
 data KeysDiff
@@ -44,13 +50,13 @@ data KeysDiff
       -- | Keys don't have any common part.
     | NoPrefix
       -- | The first key is the prefix for the second one.
-    | FstIsPref { diff :: Key}
+    | FstIsPref { diff :: !Key}
       -- | The second key is the prefix for the first one.
-    | SndIsPref { diff :: Key}
+    | SndIsPref { diff :: !Key}
       -- | Key have same prefix.
-    | Diff { pref    :: Key
-           , diffFst :: Key
-           , diffSnd :: Key
+    | Diff { pref    :: !Key
+           , diffFst :: !Key
+           , diffSnd :: !Key
            }
     deriving (Show, Eq)
 
@@ -58,11 +64,11 @@ toKey :: Piece -> [Piece] -> Key
 toKey x xs = Key $ x :| xs
 
 keysDiff :: Key -> Key -> KeysDiff
-keysDiff (Key (x :| xs)) (Key (y :| ys)) =
-    if x == y
-    then listSame xs ys []
-    else NoPrefix
+keysDiff (Key (x :| xs)) (Key (y :| ys))
+    | x == y    = listSame xs ys []
+    | otherwise = NoPrefix
   where
+    listSame :: [Piece] -> [Piece] -> [Piece] -> KeysDiff
     listSame [] []     _ = Equal
     listSame [] (s:ss) _ = FstIsPref $ toKey s ss
     listSame (f:fs) [] _ = SndIsPref $ toKey f fs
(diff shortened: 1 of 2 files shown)


src/Toml/PrefixTree.hs
```diff
@@ -0,0 +1,101 @@
...
+    if x == y
+    then listSame xs ys []
+    else NoPrefix
+  where
+    listSame [] []     _ = Equal
+    listSame [] (s:ss) _ = FstIsPref $ toKey s ss
+    listSame (f:fs) [] _ = SndIsPref $ toKey f fs
+    listSame (f:fs) (s:ss) pr =
+        if f == s
+        then listSame fs ss (pr ++ [f])
```
ChShersh: Appending element to the end of this list if very slow operation, unfortunately... Especially when such appends accumulates. Calling `reverse` function only once is a better solution but still very slow. Since this function will be called on every `lookup` and even multiple times per lookup, it's better to optimize it. Though, it's not immediately obvious to me how to do, this requires some techical details... Moreover, I don't think this is really required. In most cases `Prefix` will be just singleton map. So this `++` will be called really rare. 

Otherwise it  looks completely correct :slightly_smiling_face:  So let's leave it as it is!

src/Toml/PrefixTree.hs
```diff
@@ -0,0 +1,101 @@
...
+           , diffSnd :: Key
+           }
+    deriving (Show, Eq)
+
+toKey :: Piece -> [Piece] -> Key
+toKey x xs = Key $ x :| xs
+
+keysDiff :: Key -> Key -> KeysDiff
+keysDiff (Key (x :| xs)) (Key (y :| ys)) =
+    if x == y
```
ChShersh: Maybe it's better to use guards here?

src/Toml/PrefixTree.hs
```diff
@@ -0,0 +1,101 @@
...
+-- | Creates a 'PrefixTree' of one key-value element.
+single :: Key -> a -> PrefixTree a
+single = Leaf
+
+-- | Inserts key-value element into the given 'PrefixTree'.
+insert :: Key -> a -> PrefixTree a -> PrefixTree a
+insert = undefined
+
+-- | Looks up the value at a key in the 'PrefixTree'.
+lookup :: Key -> PrefixTree a -> Maybe a
```
ChShersh: In `TOML` we will store `PrefixMap`, not `PrefixTree`. So for every function `foo` we should have two versions. I propose to add suffix `T` for functions with work with `PrefixTree` (`T` is for `T`ree). So let's have:

```haskell
lookup  :: Key -> PrefixMap  a -> Maybe a
lookupT :: Key -> PrefixTree a -> Maybe a
```

src/Toml/PrefixTree.hs
```diff
@@ -0,0 +1,101 @@
...
+lookup lk (Leaf k v) = if lk == k then Just v else Nothing
+lookup lk (Branch pref mv prefMap) =
+    case keysDiff pref lk of
+        Equal -> mv
```
ChShersh: I like this implementation! With `keysDiff` function looks really nice! :blush:

src/Toml/PrefixTree.hs
```diff
@@ -0,0 +1,101 @@
...
+        NoPrefix -> Nothing
+        Diff _ _ _ -> Nothing
+        SndIsPref _ -> Nothing
+        FstIsPref (Key (r :| rs)) -> case HashMap.lookup r prefMap of
+            Nothing -> Nothing
+            Just pt -> case rs of
+                           []     -> bVal pt
```
ChShersh: `bVal` function is partial: it fails at runtime on `Leaf`. So it's not safe to call it here.

src/Toml/PrefixTree.hs
```diff
@@ -0,0 +1,101 @@
...
+lookup lk (Leaf k v) = if lk == k then Just v else Nothing
+lookup lk (Branch pref mv prefMap) =
+    case keysDiff pref lk of
+        Equal -> mv
+        NoPrefix -> Nothing
+        Diff _ _ _ -> Nothing
+        SndIsPref _ -> Nothing
+        FstIsPref (Key (r :| rs)) -> case HashMap.lookup r prefMap of
+            Nothing -> Nothing
+            Just pt -> case rs of
```
ChShersh: Here you can use `Monad` instance for `Maybe` to avoid manuall pattern-matching :slightly_smiling_face:

src/Toml/PrefixTree.hs
```diff
@@ -0,0 +1,101 @@
...
+                           []     -> bVal pt
+                           (x:xs) -> lookup (toKey x xs) pt
+
+
+-- | Remove a key and its value from the 'PrefixTree'. Returns the deleted value if
+-- operation succeeded or 'Nothing' in other case.
+delete :: Key -> PrefixTree a -> Maybe (PrefixTree a)
```
ChShersh: Usually `delete` function always returns just the structure, without `Maybe`. If key is not deleted than map just not changed. This makes it easier to write code like this:

```haskell
delete a $ delete b $ delete c someMap
```

src/Toml/PrefixTree.hs
```diff
@@ -0,0 +1,109 @@
...
+lookupT lk (Branch pref mv prefMap) =
+    case keysDiff pref lk of
+        Equal -> mv
+        NoPrefix -> Nothing
+        Diff _ _ _ -> Nothing
+        SndIsPref _ -> Nothing
+        FstIsPref (Key (r :| rs)) -> HashMap.lookup r prefMap >>= \pt ->
+            case rs of
+                []     -> getVal pt
+                (x:xs) -> lookupT (toKey x xs) pt
```
ChShersh: I think this case is implemented slightly incorrect. It's hard to track all these prefixes properly... But I think the main problem here is that you're trying to optimize things and for key `x :| xs` store `x` in `HashMap` head and then `xs` in the rest of the tree. Unfortunately, this won't work. Consider this mapping:

```toml
a.x = 1
a.y = 2
```

The only way to store this in our `PrefixTree` is the following:

```haskell
Branch
  { bCommonPrefix = "a" :| []
  , bVal = Nothing
  , bPrefixMap =
       [ "x" → Leaf ("x" :| []) 1
       , "b" → Leaf ("b" :| []) 2
       ]
  }
```

So yes, we will have some data duplication. We will store head of each key in `HashMap` twice: as a key in `HashMap` and as a first component of `Prefix` in `PrefixTree` value. This `HashMap` is needed only for performance reasons (that's why we don't store list of `PrefixTree`s). And this cost is not really that big. But it simplifies implementation a lot.

src/Toml/PrefixTree.hs
```diff
@@ -0,0 +1,109 @@
...
+insertT :: Key -> a -> PrefixTree a -> PrefixTree a
+insertT = undefined
+
+-- | Looks up the value at a key in the 'PrefixTree'.
+lookupT :: Key -> PrefixTree a -> Maybe a
+lookupT lk (Leaf k v) = if lk == k then Just v else Nothing
+lookupT lk (Branch pref mv prefMap) =
+    case keysDiff pref lk of
+        Equal -> mv
+        NoPrefix -> Nothing
```
ChShersh: Though this can be written shorter like:

```haskell
Equal -> mv
FstIsPref -> ...
_ -> Nothing
```

I really like that you explicitly enumerated every constructor! This makes this already difficult logic simpler to read and understand and reason about!

src/Toml/PrefixTree.hs
```diff
@@ -83,7 +83,33 @@ singleT = Leaf
 
 -- | Inserts key-value element into the given 'PrefixTree'.
 insertT :: Key -> a -> PrefixTree a -> PrefixTree a
-insertT = undefined
+insertT newK newV (Leaf k v) =
+    case keysDiff k newK of
+        Equal -> Leaf k newV
+        NoPrefix -> error "Impossible happened"
```
ChShersh: Better to write something like `"Algorithm error: can't be equal prefixes in insertT:Leaf case"`.

ChShersh:
This looks amazing! I really see how pattern-matching works well.

PR #33, opened 2018-04-30
Add PrefixMap functions, refactor

vrom911:


src/Toml/PrefixTree.hs
```diff
@@ -81,16 +85,20 @@ keysDiff (Key (x :| xs)) (Key (y :| ys))
...
 insertT newK newV (Leaf k v) =
     case keysDiff k newK of
         Equal -> Leaf k newV
         NoPrefix -> error "Algorithm error: can't be equal prefixes in insertT:Leaf case"
-        FstIsPref rK@(Key r) -> Branch k (Just v)
-          $ HashMap.fromList [(NonEmpty.head r, Leaf rK newV)]
-        SndIsPref lK@(Key l) -> Branch newK (Just newV)
-          $ HashMap.fromList [(NonEmpty.head l, Leaf lK v)]
+        FstIsPref rK@(Key (r :| _)) ->
+            Branch k (Just v) $ HashMap.singleton r (Leaf rK newV)
```
ChShersh: You can use `single` function here.

src/Toml/PrefixTree.hs
```diff
@@ -122,6 +136,6 @@ lookupT lk (Branch pref mv prefMap) =
         SndIsPref _         -> Nothing
         FstIsPref k@(Key r) -> HashMap.lookup (NonEmpty.head r) prefMap >>= lookupT k
```
ChShersh: Would be great to reuse `lookup` here

src/Toml/PrefixTree.hs
```diff
@@ -99,18 +107,24 @@ insertT newK newV (Branch pref mv prefMap) =
     case keysDiff pref newK of
         Equal    -> Branch pref (Just newV) prefMap
         NoPrefix -> error "Algorithm error: can't be equal prefixes in insertT:Branch case"
-        FstIsPref rK@(Key r) -> let piece = NonEmpty.head r in
+        FstIsPref rK@(Key (piece :| _)) ->
             Branch pref mv $
               case HashMap.lookup piece prefMap of
                   Nothing -> HashMap.insert piece (Leaf rK newV) prefMap
                   Just a  -> HashMap.insert piece (insertT rK newV a) prefMap
```
ChShersh: I think you can reuse `insert` function here to avoid some code duplication.

PR #34, opened 2018-05-01
[#32] Add InsertLookup property test for PrefixTree

### LINKED ISSUE #32: Property tests for 'PrefixTree' and 'PrefixMap'
Our `PrefixTree` is not really the most simple data structure... Would be great to have some property tests for it. Specifically:

- [x] `PrefixMap` generator. Ideally using [`recursive`](https://hackage.haskell.org/package/hedgehog-0.5.3/docs/Hedgehog-Gen.html#v:recursive) from `hedgehog` library.
- [x] **InsertLookup**: `lookup k (insert k v t) ≡ Just v`
- [x] **InsertInsert**: `insert x a . insert x b ≡ insert x a` (can be tested as `lookup a (insert x a $ insert x b t) === Just a`)

ChShersh:
Resolves #32 

Implements `InsertLookup` property test for `PrefixTree`. I can't think of more properties for now... Maybe more of them can be discovered in future.

I've also added some unit tests.

### DIFF
--- a/src/Toml/PrefixTree.hs
+++ b/src/Toml/PrefixTree.hs
@@ -16,6 +16,7 @@ module Toml.PrefixTree
          -- * Types
        , Piece (..)
        , Key (..)
+       , pattern (:||)
        , Prefix
        , KeysDiff (..)
        ) where
@@ -77,6 +78,7 @@ data PrefixTree a
              , bVal        :: !(Maybe a)      -- ^ value by key = prefix
              , bPrefixMap  :: !(PrefixMap a)  -- ^ suffixes of prefix
              }
+    deriving (Show, Eq)
 
 data KeysDiff
       -- | Keys are equal
--- a/test/Test/Toml/PrefixTree.hs
+++ b/test/Test/Toml/PrefixTree.hs
@@ -1,0 +1,13 @@
+module Test.Toml.PrefixTree
+       ( prefixTreeTests
+       ) where
+
+import Test.Tasty (TestTree, testGroup)
+
+import Test.Toml.PrefixTree.Property (propertyTests)
+import Test.Toml.PrefixTree.Unit (unitTests)
+
+prefixTreeTests :: IO TestTree
+prefixTreeTests = do
+   uTests <- unitTests
+   pure $ testGroup "Prefix Tree tests" $ uTests : propertyTests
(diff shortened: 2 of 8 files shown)


src/Toml/PrefixTree.hs
```diff
@@ -34,6 +36,9 @@ import qualified Data.List.NonEmpty as NonEmpty
 newtype Piece = Piece { unPiece :: Text }
     deriving (Show, Eq, Hashable)
 
+instance IsString Piece where
+    fromString = Piece . fromString
```
vrom911: I just made the same thing :slightly_smiling_face:  I just derived :)

test/Test/Toml/PrefixTree.hs
```diff
@@ -0,0 +1,13 @@
+module Test.Toml.PrefixTree
```
vrom911: I really like this structure! :+1:  It makes code easier to read with this test separation

test/Test/Toml/PrefixTree/Property.hs
```diff
@@ -0,0 +1,98 @@
...
+genPrefixMap = do
+    entries <- Gen.list (Range.linear 0 10) genEntry
+    kvps    <- forM entries $ \(piece, key) -> do
+        tree <- genPrefixTree key
+        pure (piece, tree)
+
+    pure $ HashMap.fromList kvps
+
+genPrefixTree :: forall m . MonadGen m => Key -> m (PrefixTree V)
+genPrefixTree key = Gen.recursive
```
vrom911: That's nice to see this library feature to work for us! :clap:

test/Test/Toml/PrefixTree/Property.hs
```diff
@@ -0,0 +1,98 @@
...
+prop_insertLookup =  property $ do
+    prefMap <- forAll genPrefixMap
+    key     <- forAll genKey
+    val     <- forAll genVal
+
+    Prefix.lookup key (Prefix.insert key val prefMap) === Just val
+
+    -- DEBUG: ensures that trees of depth at least 5 are generated
+    -- assert $ depth prefMap < 5
+
```
vrom911: Another property I can think of is
```
lookup a (insert a y $ insert a x t) === Just y
```
to check that it uses the latest value. How do you think?

src/Toml/PrefixTree.hs
ChShersh: Yeah, I should derive it as well...

test/Test/Toml/PrefixTree/Property.hs
ChShersh: @vrom911 That's an excellent property! I can add it as well.

PR #35, opened 2018-05-01
[#31] Change TOML to use PrefixMap

### LINKED ISSUE #31: Replace 'HashMap' in 'TOML' with our 'PrefixTree'
This involves:

- [x] Chane `TOML` type
- [x] Add `insert` and `lookup` functions for `PrefixMap` (basically just refactoring current implementation)
- [x] Change parser (removing merging, using `insert` from `PrefixMap`)
- [x] Change renderer

vrom911:
Resolves #31

### DIFF
--- a/src/Toml/Parser.hs
+++ b/src/Toml/Parser.hs
@@ -9,21 +9,18 @@ module Toml.Parser
 import Control.Applicative (Alternative (..), liftA2)
 import Control.Applicative.Combinators (between, manyTill, sepBy)
 import Control.Monad (void)
-import Data.Function (on)
-import Data.HashMap.Lazy (HashMap)
-import Data.List.NonEmpty (NonEmpty (..))
-import Data.Ord (comparing)
 import Data.Semigroup ((<>))
 import Data.Text (Text)
 import Data.Void (Void)
 import Text.Megaparsec (Parsec, parseErrorPretty', try)
 import Text.Megaparsec.Char (alphaNumChar, anyChar, char, space, space1)
 
-import Toml.Type (AnyValue, Key (..), TOML (..), UValue (..), typeCheck)
+import Toml.PrefixTree (Key (..), Piece (..), insert)
+import Toml.Type (AnyValue, TOML (..), UValue (..), typeCheck)
 
 import qualified Control.Applicative.Combinators.NonEmpty as NC
 import qualified Data.HashMap.Lazy as HashMap
-import qualified Data.List as List (isPrefixOf, sortBy)
+import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.Text as Text
 import qualified Text.Megaparsec as Mega (parse)
 import qualified Text.Megaparsec.Char.Lexer as L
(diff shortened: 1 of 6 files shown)


src/Toml/Parser.hs
```diff
@@ -81,7 +78,7 @@ keyComponentP :: Parser Text
 keyComponentP = bareKeyP <|> (quote <$> stringP)
 
 keyP :: Parser Key
-keyP = Key <$> NC.sepBy1 keyComponentP (char '.')
+keyP = Key <$> NonEmpty.map Piece <$> NC.sepBy1 keyComponentP (char '.')
```
ChShersh: I think it's better to change type of `keyComponentP` to return `Piece`.

src/Toml/Parser.hs
```diff
@@ -122,55 +119,23 @@ keyValP = do
...
+    makeToml :: [(Key, AnyValue)] -> TOML
+    makeToml kv = TOML (HashMap.fromList kv) mempty
 
 tomlP :: Parser TOML
 tomlP = do
     kvs    <- many keyValP
     tables <- many tableHeaderP
     pure TOML { tomlPairs  = HashMap.fromList kvs
-              , tomlTables = merge $ List.sortBy (comparing thName) tables
+              , tomlTables = foldr (\TableHeader{..} -> insert thName thKeys) mempty tables
```
ChShersh: It's usually better to use `foldl'` function. I think you can use `foldl'` and implement this lambda in `where`. Or maybe you can somehow use `fromList` function here by remove `TableHeader` data type and replacing it with just a pair...

src/Toml/Parser.hs
```diff
@@ -122,55 +119,23 @@ keyValP = do
...
-{- Groups by prefix. So the following list:
-
-["a", "a.x", "a.y", "b.1", "b.2", "c"]
-
-will be converted in the following list of non empty lists:
-
-[ "a" :| ["a.x", "a.y"], "b.1" :| [], "b.2" :| [], "c" :| [] ]
-
--}
-groupByPrefix :: [TableHeader] -> [NonEmpty TableHeader]
```
ChShersh: It's so nice to see this ugly code removed!

src/Toml/Printer.hs
```diff
@@ -81,15 +82,29 @@ prettyKeyValue i = Text.concat . map kvText . HashMap.toList
     showText :: Show a => a -> Text
     showText = Text.pack . show
 
-prettyTables :: Int -> HashMap Key TOML -> Text
-prettyTables i = Text.concat . map prettyTable . HashMap.toList
+prettyTables :: Int -> Text -> PrefixMap TOML -> Text
+prettyTables i pref = Text.concat . map prettyTable . HashMap.elems
```
ChShersh: Code in this function looks not very elegant... But I don't see better way to implement it. Need to think more, it's not so obvious...

src/Toml/PrefixTree.hs
```diff
@@ -139,3 +158,9 @@ lookupT lk (Branch pref mv prefMap) =
 -- | Looks up the value at a key in the 'PrefixMap'.
 lookup :: Key -> PrefixMap a -> Maybe a
 lookup k@(p :|| _) prefMap = HashMap.lookup p prefMap >>= lookupT k
+
+fromList :: [(Key, a)] -> PrefixMap a
```
ChShersh: I think this function requires some documentation.

src/Toml/Parser.hs
vrom911: That actually cool idea! Will work on it :wink:

PR #36, opened 2018-05-03
[#25] Implement basic bidirectional Toml conversion

### LINKED ISSUE #25: Add ability to convert between 'Toml' and user custom data types
What we want is to have some way to describe correspondence between user data types and TOML representation. Consider this Haskell data type:

```haskell
data Node = Node { root :: Bool, labels :: [Int] }
```

Would be great to have an ability to specify correspondence in some way like this:

```haskell
nodeToml :: TomlM Node
nodeToml = do
    root   <- bool "root"
    labels <- array @Int "labels"
    pure Node{..}
```

And then we can have two functions:

```haskell
decode :: Toml -> TomlM a -> Either DecodeError a
encode :: a -> TomlM a -> Toml
```

### Difficult questions

1. How this data type should be encoded in TOML?

Like this:

```toml
node.root = true
node.labels = [ 1, 2, 3 ]
```

or like this:

```toml
[node]
root = true
labels = [ 1, 2, 3 ]
```

I think we should give an ability to specify which way user want.

2. How to encode sum types?

I propose the following strategy: if you have `foo :: Either Int String` then it should be either

```toml
foo.Left = 3
```

or

```toml
foo.Right = "bar"
```

If you have both keys, this should be a type error. And, again, give ability to specify `foo` as table. And this scheme easily can be extended if constructors have multiple fields.

3. What to do with arrays of different values?

In TOML it's possible to have:

```toml
foo = [ [1, 2, 3], [true, false] ]
```

But Haskell doesn't really have a standard data structure for such case... So should we care about supporting this case or not?..

ChShersh:
Resolves #25 

Implementation is not full. It doesn't support all features. But it works!

What is not implemented:

- [ ] Arrays
- [ ] Tables
- [ ] Pretty error messages

But this should be relatively easy to add.

### DIFF
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -3,7 +3,7 @@ module Main where
 import Data.Text (Text)
 import Data.Time (fromGregorian)
 
-import Toml.Bi (BiToml, bool, double, encode, int, str, unsafeDecode, (.=))
+import Toml.Bi (BiToml, (.=))
 import Toml.Parser (ParseException (..), parse)
 import Toml.PrefixTree (Key (..), Piece (..), PrefixMap, fromList)
 import Toml.Printer (prettyToml)
@@ -12,6 +12,7 @@ import Toml.Type (AnyValue (..), DateTime (..), TOML (..), Value (..))
 import qualified Data.HashMap.Strict as HashMap
 import qualified Data.List.NonEmpty as NonEmpty
 import qualified Data.Text.IO as TIO
+import qualified Toml
 
 data Test = Test
   { testB :: Bool
@@ -22,10 +23,10 @@ data Test = Test
 
 testT :: BiToml Test
 testT = Test
- <$> bool   (key ["testB"]) .= testB
- <*> int    (key ["testI"]) .= testI
- <*> double (key ["testF"]) .= testF
- <*> str    (key ["testS"]) .= testS
+ <$> Toml.bool   (key ["testB"]) .= testB
+ <*> Toml.int    (key ["testI"]) .= testI
+ <*> Toml.double (key ["testF"]) .= testF
+ <*> Toml.str    (key ["testS"]) .= testS
 
 main :: IO ()
 main = do
@@ -40,9 +41,9 @@ main = do
 
     TIO.putStrLn "=== Testing bidirectional conversion ==="
     biFile <- TIO.readFile "examples/biTest.toml"
-    case encode testT biFile of
+    case Toml.encode testT biFile of
         Left _     -> putStrLn "Some error"
-        Right test -> TIO.putStrLn $ unsafeDecode testT test
+        Right test -> TIO.putStrLn $ Toml.unsafeDecode testT test
 
 key :: [Text] -> Key
 key = Key . NonEmpty.fromList . map Piece
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -3,6 +3,7 @@ module Main where
 import Data.Text (Text)
 import Data.Time (fromGregorian)
 
+import Toml.Bi (BiToml, bool, double, encode, int, str, unsafeDecode, (.=))
 import Toml.Parser (ParseException (..), parse)
 import Toml.PrefixTree (Key (..), Piece (..), PrefixMap, fromList)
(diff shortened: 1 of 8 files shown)


examples/Playground.hs
```diff
@@ -12,6 +13,20 @@ import qualified Data.HashMap.Strict as HashMap
...
+  , testF :: Double
+  , testS :: Text
+  }
+
+testT :: BiToml Test
+testT = Test
+ <$> bool   (key ["testB"]) .= testB
+ <*> int    (key ["testI"]) .= testI
+ <*> double (key ["testF"]) .= testF
+ <*> str    (key ["testS"]) .= testS
```
ChShersh: This is how it looks like. With `IsString` instance for `Key` can be even better! But it requires to add extra validation to `Key` in conversion level.

src/Toml/Bi/Combinators.hs
```diff
@@ -0,0 +1,146 @@
...
+    fromDouble :: Value f -> Maybe Double
+    fromDouble (Float f) = Just f
+    fromDouble _         = Nothing
+
+-- | Parser for string values.
+str :: Key -> BiToml Text
+str = bijectionMaker "String" fromString String
+  where
+    fromString :: Value f -> Maybe Text
+    fromString (String s) = Just s
```
ChShersh: Unfortunately, patterns are not first class objects in Haskell at the moment, so I need to write this function manually each time...

examples/Playground.hs
```diff
@@ -23,6 +38,12 @@ main = do
         Left (ParseException e) -> TIO.putStrLn e
         Right toml              -> TIO.putStrLn $ prettyToml toml
 
+    TIO.putStrLn "=== Testing bidirectional conversion ==="
+    biFile <- TIO.readFile "examples/biTest.toml"
+    case encode testT biFile of
```
vrom911: Probably should consider using qualified imports in here for readability

src/Toml/Bi/Monad.hs
```diff
@@ -0,0 +1,87 @@
...
+       , (.=)
+       ) where
+
+{- | Monad for bidirectional Toml conversion. Contains pair of functions:
+
+1. How to read value of type @a@ from immutable environment context @r@?
+2. How to store value of type @a@ in stateful context @w@?
+
+In practice instead of @r@ we will use some @Reader Toml@ and instead of @w@ we will
+use @State Toml@. This with the bunch of utility functions allows to have single
```
vrom911: Missing word, probably : This *approach* with the..

src/Toml/Bi/Monad.hs
```diff
@@ -0,0 +1,87 @@
...
+described in more details in [related blog post](https://blog.poisson.chat/posts/2016-10-12-bidirectional-serialization.html).
+
+-}
+data Bijection r w c a = Bijection
+    { -- | Extract value of type @a@ from monadic context @r@.
+      biRead  :: r a
+
+      -- | Store value of type @c@ inside monadic context @w@ and returning
+      -- value of type @a@. Type of this function actually should be @a -> w ()@ but with
+      -- such type it's impossible to have 'Monad' ant other instances.
```
vrom911: typo: an*d*

src/Toml/Bi/Monad.hs
```diff
@@ -0,0 +1,87 @@
...
+
+1. How to get field from object?
+2. How to write this field to toml?
+
+In code this should be used like this:
+
+@
+data Foo = Foo { fooBar :: Int, fooBaz :: String }
+
+foo :: BiTomo Foo
```
vrom911: Typo: BiTom*l*

src/Toml/Bi/Combinators.hs
```diff
@@ -0,0 +1,149 @@
...
+    _ <- result
+
+    pure $ prettyToml toml
+
+fromRight :: b -> Either a b -> b
+fromRight b (Left _)  = b
+fromRight _ (Right b) = b
+
+-- | Unsafe version of 'decode' function if you're sure that you decoding
+-- structure is correct.
```
vrom911: your decoding *of* structure

PR #45, opened 2018-05-23
[#44] Add bidirectional converter for Maybe

### LINKED ISSUE #44: Introduce converter for 'Maybe'
It's not possible now to parse absent field. `Maybe` converter should solve this problem.

ChShersh:
Resolves #44

### DIFF
--- a/src/Toml.hs
+++ b/src/Toml.hs
@@ -9,11 +9,13 @@ import qualified Toml
 module Toml
     ( module Toml.Bi
     , module Toml.Parser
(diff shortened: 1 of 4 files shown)


src/Toml.hs
```diff
@@ -9,11 +9,13 @@ import qualified Toml
 module Toml
     ( module Toml.Bi
     , module Toml.Parser
+    , module Toml.PrefixTree
```
vrom911: Thank you!

PR #49, opened 2018-05-23
[#47] Introduce mdimap combinator

### LINKED ISSUE #47: Non-bijectional 'dimapBijection'
`dimapBijection` works good for things like `Num` and `newtype`s but if you have functions like:

```haskell
parseMyType :: Text -> Maybe MyType
showMyType  :: MyType -> Text
```

you will have some problems with dimapping just pair of functions for `str` converter 😞 

So new combinator should be created. Also, since `dimapBijection` is used very often let's make it name shorter (just `dimap`) and let's call new function `mdimap` (stands for _maybe dimap_).

ChShersh:
Resolves #47

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Change log
 tomland uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
+0.2.0
+=====
+* Switch names for `decode` and `encode` functions.
+* [#47](https://github.com/kowainik/tomland/issues/47):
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ Change log
 tomland uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
+0.2.0
+=====
+* Switch names for `decode` and `encode` functions.
```
vrom911: 😏

PR #51, opened 2018-05-26
[#37] Add table support to bi

### LINKED ISSUE #37: Support tables in bidirectional conversion

ChShersh:
Resolves #37

### DIFF
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -35,6 +35,7 @@ module Toml.Bi.Combinators
        , str
        , arrayOf
        , maybeP
+       , table
 
          -- * Value parsers
        , Valuer (..)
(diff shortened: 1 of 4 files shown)


src/Toml/Bi/Combinators.hs
```diff
@@ -170,7 +147,7 @@ mdimap toString toMaybe bi = Bijection
   , biWrite = \s -> do
         retS <- biWrite bi $ toString s
         case toMaybe retS of
-            Nothing -> throwError $ DecodeParsing "Unable to convert" -- TODO
+            Nothing -> error $ "Given pair of functions for 'mdimap' doesn't satisfy roundtrip property"
```
vrom911: Probably highlight this behavior in the haddock and say what property should be satisfied for these functions?

PR #52, opened 2018-05-28
[#39] Pretty exceptions

### LINKED ISSUE #39: Implement pretty 'Show' instance for bi-exceptions

ChShersh:
Resolves #39

### DIFF
--- a/src/Toml/Bi/Code.hs
+++ b/src/Toml/Bi/Code.hs
@@ -1,0 +1,72 @@
+module Toml.Bi.Code
+       ( -- * Types
+         BiToml
+       , Env
+       , St
+
+         -- * Exceptions
+       , DecodeException (..)
+       , prettyException
+
+         -- * Encode/Decode
+       , decode
+       , encode
+       ) where
+
+import Control.Monad.Except (ExceptT, runExceptT)
+import Control.Monad.Reader (Reader, runReader)
+import Control.Monad.State (State, execState)
+import Data.Bifunctor (first)
+import Data.Semigroup ((<>))
+import Data.Text (Text)
+import Data.Foldable (toList)
+
+import Toml.Bi.Monad (Bi, Bijection (..))
+import Toml.Parser (ParseException (..), parse)
+import Toml.PrefixTree (Key (..), unPiece)
+import Toml.Printer (prettyToml)
+import Toml.Type (TOML (..), ValueType, showType)
+
+import qualified Data.Text as Text
+
+-- | Type of exception for converting from 'Toml' to user custom data type.
+data DecodeException
+    = KeyNotFound Key  -- ^ No such key
+    | TableNotFound Key  -- ^ No such table
+    | TypeMismatch Key Text ValueType -- ^ Expected type vs actual type
+    | ParseError ParseException  -- ^ Exception during parsing
+    deriving (Eq, Show)  -- TODO: manual pretty show instances
+
+-- | Converts 'DecodeException' into pretty human-readable text.
+prettyException :: DecodeException -> Text
+prettyException = \case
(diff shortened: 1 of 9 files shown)


src/Toml/Bi/Code.hs
```diff
@@ -0,0 +1,72 @@
...
+
+import qualified Data.Text as Text
+
+-- | Type of exception for converting from 'Toml' to user custom data type.
+data DecodeException
+    = KeyNotFound Key  -- ^ No such key
+    | TableNotFound Key  -- ^ No such table
+    | TypeMismatch Key Text ValueType -- ^ Expected type vs actual type
+    | ParseError ParseException  -- ^ Exception during parsing
+    deriving (Eq, Show)  -- TODO: manual pretty show instances
```
vrom911: Isn't `show` is just `toString . prettyException` now?

src/Toml/Bi/Combinators.hs
```diff
@@ -250,9 +213,13 @@ maybeP converter key = let bi = converter key in Bijection
...
     handleNotFound :: DecodeException -> Env (Maybe a)
-    handleNotFound (KeyNotFound _)   = pure Nothing
-    handleNotFound (TableNotFound _) = pure Nothing
-    handleNotFound e                 = throwError e
+    handleNotFound e@(KeyNotFound name)
+        | name == key = pure Nothing
+        | otherwise = throwError e
+    handleNotFound e@(TableNotFound name)
+        | name == key = pure Nothing
+        | otherwise = throwError e
```
vrom911: Probably we could some how get rid of code duplication.. But I can't see the better way for this, so let keep it as it is. At least it's readable!

src/Toml/Bi/Combinators.hs
```diff
@@ -272,7 +239,7 @@ table bi key = Bijection input output
...
         a <$ modify (\(TOML vals tables) -> TOML vals (Prefix.insert key newToml tables))
 
-    rethrowNotFound :: DecodeException -> Env a
-    rethrowNotFound (KeyNotFound name)   = throwError $ IncompleteTable key name
-    rethrowNotFound (TableNotFound name) = throwError $ IncompleteTable key name
-    rethrowNotFound e                    = throwError e
+    addTableName :: DecodeException -> Env a
+    addTableName (KeyNotFound name)   = throwError $ KeyNotFound (key <> name)
+    addTableName (TableNotFound name) = throwError $ TableNotFound (key <> name)
+    addTableName e                    = throwError e
```
vrom911: Probably it's better to call it `tableNameHandler` or something like this just to see that it's for error handling?

ChShersh: Good catch! I also forgot about one error here...

src/Toml/Bi/Combinators.hs
ChShersh: I think it's possible. Will try to rewrite it.

src/Toml/Bi/Code.hs
ChShersh: Personally I like to keep default definition of `Show` just to be able to see structure of data type if needed. But I don't mind writing `Show` using `prettyException`. Though, it requires more code than just generating `Show` instance...

src/Toml/Bi/Combinators.hs
```diff
@@ -213,13 +213,9 @@ maybeP converter key = let bi = converter key in Bijection
     }
   where
     handleNotFound :: DecodeException -> Env (Maybe a)
-    handleNotFound e@(KeyNotFound name)
-        | name == key = pure Nothing
+    handleNotFound e
+        | e `elem` [KeyNotFound key, TableNotFound key] = pure Nothing
```
vrom911: Wow! This is cool solution! 🍾

PR #55, opened 2018-06-03
[#8] Create EDSL for tomland

### LINKED ISSUE #8: Create EDSL for manually specifying TOML data types
Sometimes you don't want to specify TOML in file, you want to have configuration embedded to code. Currently you can specify everything manually using `HashMap`s and constructors:

```haskell
myToml :: TOML
myToml = TOML (fromList
    [ (Key "a", Bool True)
    , (Key "list", Array [String "one", String "two"])
    , (Key "time", Array [Date $ Day (fromGregorian 2018 3 29)])
    ] ) myInnerToml mempty

myInnerToml :: HashMap TableId TOML
myInnerToml = fromList $ [ ( TableId (NonEmpty.fromList ["table", "name", "1"])
                           , TOML (fromList [ (Key "aInner", Int 1)
                                            , (Key "listInner", Array [Bool True, Bool False])
                                            ]) myInnerInnerToml mempty
                           )
                         , ( TableId (NonEmpty.fromList ["table", "name", "2"])
                           , TOML (fromList [ (Key "2Inner", Int 42)
                                            ]) mempty mempty
                           )
                         ]
```

But this is ugly and not convenient. What would be great is to have some EDSL so the above can be configured like this:

```haskell
myToml :: TOML
myToml = configure $ do
  "a"    :=: True
  "list" :=: ["one", "two"]
  "time" :=: Day (fromGregorian 2018 3 29) -- maybe something more convenient here
  table "table.name.1" $ do
    "aInner"    :=: 1
    "listInner" :=: [True, False]
    ...
```

The last one looks more convenient and simpler to me.

vrom911:
Resolves #8

### DIFF
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -4,7 +4,7 @@ import Data.Text (Text)
 import Data.Time (fromGregorian)
 
 import Toml.Bi (BiToml, (.=))
-import Toml.Edsl (table, toml, (=:))
+import Toml.Edsl (mkToml, table, (=:))
 import Toml.Parser (ParseException (..), parse)
 import Toml.Printer (prettyToml)
 import Toml.Type (DateTime (..), TOML (..), Value (..))
@@ -48,7 +48,7 @@ main = do
     content <- TIO.readFile "test.toml"
     case parse content of
         Left (ParseException e) -> TIO.putStrLn e
-        Right tml               -> TIO.putStrLn $ prettyToml tml
+        Right toml              -> TIO.putStrLn $ prettyToml toml
 
     TIO.putStrLn "=== Testing bidirectional conversion ==="
     biFile <- TIO.readFile "examples/biTest.toml"
@@ -57,17 +57,17 @@ main = do
         Right test -> Toml.encode testT test
 
 myToml :: TOML
-myToml = toml $ do
+myToml = mkToml $ do
     "a" =: Bool True
-    "list" =: Array [String "one", String "two"]
+    "list" =: Array ["one", "two"]
     "time" =: Array [Date $ Day (fromGregorian 2018 3 29)]
     table "table.name.1" $ do
-        "aInner" =: Int 1
+        "aInner" =: 1
         "listInner" =: Array [Bool True, Bool False]
         table "1" $ do
-            "aInner11" =: Int 11
-            "listInner11" =: Array [Int 0, Int 1]
+            "aInner11" =: 11
+            "listInner11" =: Array [0, 1]
         table "2" $
-            "Inner12" =: Int 12
+            "Inner12" =: "12"
     table "table.name.2" $
-        "Inner2" =: Int 42
+        "Inner2" =: 42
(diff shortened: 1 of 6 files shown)


src/Toml/Edsl.hs
```diff
@@ -0,0 +1,26 @@
...
+
+toml :: Env -> TOML
+toml env = execState env $ TOML mempty mempty
+
+(=:) :: Key -> Value a -> Env
+(=:) k v = modify $ \tml -> tml {tomlPairs = HashMap.insert k (AnyValue v) (tomlPairs tml)}
+
+table :: Key -> Env -> Env
+table k env = modify $ \tml -> tml
+    { tomlTables = Prefix.insert k (execState env (TOML mempty mempty)) (tomlTables tml)
```
ChShersh: `execState env (TOML mempty mempty)` looks like `toml` function you defined above!

examples/Playground.hs
```diff
@@ -58,35 +57,17 @@ main = do
...
-    , ( "table.name.1.2"
-      , TOML (HashMap.fromList [("Inner1.2", AnyValue $ Int 42)]) mempty
-      )
-    ]
+myToml = toml $ do
+    "a" =: Bool True
+    "list" =: Array [String "one", String "two"]
+    "time" =: Array [Date $ Day (fromGregorian 2018 3 29)]
+    table "table.name.1" $ do
+        "aInner" =: Int 1
```
ChShersh: I think I know how you can write `key =: 1` instead of `ke =: Int 1`. Since we have GADT for `Value`, you can implement `Num` instance for `Value 'TInt`. And this will guarantee that it will be converted to `Value` using `Int` constructor. Also all pattern matchings will be exhaustive inside instance implementation because of that!

Same for `IsString` instance for `Value`. Not sure about `IsList` intsance, it probably will be impossible to implement... And, unfortunately, there's no `IsBool` type class in Haskell :disappointed:

src/Toml/Edsl.hs
```diff
@@ -0,0 +1,26 @@
...
+
+import Control.Monad.State (State, execState, modify)
+
+import Toml.PrefixTree (Key)
+import Toml.Type (AnyValue (..), TOML (..), Value (..))
+
+import qualified Data.HashMap.Strict as HashMap
+import qualified Toml.PrefixTree as Prefix
+
+type Env = State TOML ()
```
ChShersh: Maybe it's better to name this type somehow else? Since it EDSL for creating toml, we could name it like `TomlDsl` or `Tdsl` or `TDSL` or something like that.

examples/Playground.hs
```diff
@@ -49,7 +48,7 @@ main = do
     content <- TIO.readFile "test.toml"
     case parse content of
         Left (ParseException e) -> TIO.putStrLn e
-        Right toml              -> TIO.putStrLn $ prettyToml toml
+        Right tml               -> TIO.putStrLn $ prettyToml tml
```
ChShersh: Maybe `toml` is not the best name since if you export `toml` function you can't use `toml` as variable name? Maybe it's better to call it `mkToml` or `declareTolm` or `defineToml` to `constructToml` or `createToml` or something else? What do you think?

Issue #40, opened 2018-05-20
Support full-featured string parser


ChShersh:
Relates to #16 

See details here: https://github.com/kowainik/tomland/pull/12#discussion_r178982571

PR #63, opened 2018-06-17
[#15] Add parsing of multiline strings

### LINKED ISSUE #15: Add parsing of multiline strings
Multiline strings are quite useful. But it's not easy to parse them...

ghallak:
Resolves #15 

There are a few things from the TOML specification that I wasn't not sure how to handle:

1. > TOML parsers should feel free to normalize newline to whatever makes sense for their platform.

    Is it the compiler's responsibility to handle this, or the programmer's?

2. > All of the escape sequences that are valid for basic strings are also valid for multi-line basic strings.

    Not sure what this could mean!

3. > Any Unicode character may be used except those that must be escaped: backslash and the control characters (U+0000 to U+001F, U+007F)

   Does this mean that the characters that must be escaped cannot be used, or does it mean that they should be escaped when they are used?
    And if it means that they can be used, then I'm not sure what to do with control characters after escaping them!

4. > Control characters other than tab are not permitted in a literal string.

    What should happen when there is a control character inside the literal string?


test/Test/Toml/Parsing/Unit.hs
```diff
@@ -293,6 +319,18 @@ spec_Parser = do
       --  stringFailOn (squote "\nabc")
       --  stringFailOn (squote "ab\r\nc")
       --  stringFailOn (squote "abc\n")
+    context "when the string is a multi-line literal string" $ do
+      let squote3 = quoteWith "'''"
+
+      it "can parse multi-line strings surrounded by three single quotes" $ do
+        parseText (squote3 "first line \nsecond.\n   3\n") "first line \nsecond.\n   3\n"
```
ChShersh: Let's use [`neat-interpolation`](http://hackage.haskell.org/package/neat-interpolation) package for testing. So instead of writing string like this:

```haskell
squote3 "first line \nsecond.\n   3\n"
```

You will just write like this (using `{-# LANGUAGE QuasiQuotes #-}`:

```haskell
[text|
'''first line
second.
    3
'''
|]
```

This makes it more visible what actually parsed. Tests will become bigger, I understand. But we will think how to split tests later.

src/Toml/Parser.hs
```diff
@@ -79,12 +80,66 @@ basicStringP :: Parser Text
...
+textP = multilineStringP <|> multilineLiteralStringP <|> literalStringP <|> basicStringP
 
 -- adds " or ' to both sides
 quote :: Text -> Text -> Text
 quote q t = q <> t <> q
 
+nonControlChar :: Parser Text
+nonControlChar = Text.singleton <$> satisfy (not . isControl)
+
+multilineStringP :: Parser Text
```
ChShersh: Let's call this parser `multilineBasicStringP`

src/Toml/Parser.hs
```diff
@@ -79,12 +80,66 @@ basicStringP :: Parser Text
...
+    allowedChar :: Parser Text
+    allowedChar = lineEndingBackslash <|> escapeSequence <|> nonControlChar <|> eol
+
+    lineEndingBackslash :: Parser Text
+    lineEndingBackslash = Text.empty <$ try (char '\\' >> eol >> space)
+
+    escapeSequence :: Parser Text
+    escapeSequence = char '\\' >> anyChar
+                 >>= \case
+                      'b'  -> return "\b"
```
ChShersh: Oh, I see that you're already done issue #40. And this is very cool! So it's better to refactor code to make `basicStringP` parser also support these characters. But let's do this refactoring under separate PR, let's keep this PR simpler.

src/Toml/Parser.hs
```diff
@@ -79,12 +80,66 @@ basicStringP :: Parser Text
...
+                      't'  -> return "\t"
```
ChShersh: It's better to use `pure` instead of `return`, because it's shorter 😏

src/Toml/Parser.hs
```diff
@@ -79,12 +80,66 @@ basicStringP :: Parser Text
...
+                      'n'  -> return "\n"
+                      'f'  -> return "\f"
+                      'r'  -> return "\r"
+                      '"'  -> return "\""
+                      '\\' -> return "\\"
+                      'u'  -> parseHexUnicode 4
+                      'U'  -> parseHexUnicode 8
+                      _    -> fail ""
```
ChShersh: Instead of `""` it's better to write like this:

```haskell
c -> fail $ "Invalid escape sequence: " <> "\\" <> singleton c
```

src/Toml/Parser.hs
```diff
@@ -79,12 +80,66 @@ basicStringP :: Parser Text
...
+      where
+        parseHexUnicode :: Int -> Parser Text
+        parseHexUnicode n = count n hexDigitChar
+                        >>= \x -> case (toUnicode . toInt) x of
+                              Just c  -> return (Text.singleton c)
+                              Nothing -> fail ""
```
ChShersh: Same here for more descriptive error.

src/Toml/Parser.hs
```diff
@@ -79,12 +80,66 @@ basicStringP :: Parser Text
...
+          where
+            toInt xs = (read :: String -> Int) ("0x" ++ xs)
```
ChShersh: We already have parser for hexadecimal number. I think we can reuse it here. If it's too complicated, then at least it's better to enable paranoia mode and use `readMaybe` function.

src/Toml/Parser.hs
```diff
@@ -79,12 +80,66 @@ basicStringP :: Parser Text
...
+                      'r'  -> return "\r"
+                      '"'  -> return "\""
+                      '\\' -> return "\\"
+                      'u'  -> parseHexUnicode 4
+                      'U'  -> parseHexUnicode 8
+                      _    -> fail ""
+      where
+        parseHexUnicode :: Int -> Parser Text
+        parseHexUnicode n = count n hexDigitChar
+                        >>= \x -> case (toUnicode . toInt) x of
```
ChShersh: `toUnicode $ toInt x` contains less noise than `(toUnicode . toInt) x`

src/Toml/Parser.hs
ChShersh: Also, it's better to write top-level type singatures for functions inside `where` instead of specifying types internally. So this is better:

```haskell
toInt :: String -> Int
toInt xs = ...
```

src/Toml/Parser.hs
```diff
@@ -79,12 +80,66 @@ basicStringP :: Parser Text
...
+                              Just c  -> return (Text.singleton c)
+                              Nothing -> fail ""
+          where
+            toInt xs = (read :: String -> Int) ("0x" ++ xs)
+            toUnicode x
```
ChShersh: Looks weird that this function is not anywhere on Haddock 🤔 But I didn't found it either. I can recommend asking on SO.

src/Toml/Parser.hs
```diff
@@ -79,12 +80,66 @@ basicStringP :: Parser Text
...
+
+    allowedChar :: Parser Text
+    allowedChar = lineEndingBackslash <|> escapeSequence <|> nonControlChar <|> eol
+
+    lineEndingBackslash :: Parser Text
+    lineEndingBackslash = Text.empty <$ try (char '\\' >> eol >> space)
+
+    escapeSequence :: Parser Text
+    escapeSequence = char '\\' >> anyChar
+                 >>= \case
```
ChShersh: I think you can join this line with line above to make code contain less lines.

src/Toml/Parser.hs
```diff
@@ -79,12 +80,66 @@ basicStringP :: Parser Text
...
 quote :: Text -> Text -> Text
 quote q t = q <> t <> q
 
+nonControlChar :: Parser Text
+nonControlChar = Text.singleton <$> satisfy (not . isControl)
+
+multilineStringP :: Parser Text
+multilineStringP = lexeme $ fmap mconcat $ quotes3
+                                        >> optional eol
+                                        >> allowedChar `manyTill` quotes3
```
ChShersh: These three lines are almost the same for `multilineBasicStringP` and `multilineLiteralStringP`. It's better then to create parser `multilineP` which takes parser as an argument. This will help to remove code duplication.

ChShersh:
> I wasn't sure if it is the compiler's responsibility to handle this, or the programmer's, so I couldn't add tests for it.

Using `eol` parser is a correct choice in this situation.

src/Toml/Parser.hs
ghallak: That's why I wanted to merge #71 and close #40 before working on multiline strings.

But since we're working on #15 in this PR, what do you suggest as a change here? Should I make the implementation of `multilineStringP` wrong? or should I fix the implementation of `basicStringP`?

src/Toml/Parser.hs
ghallak: That's a good reason to use `pure`, that I haven't thought of before.

src/Toml/Parser.hs
ghallak: The parser for hexadecimal numbers that we have is not able to parse a specific number of digits (in this case we need to parser either 4 or 8 digits), so I implemented this `toInt` function.

The `read` or `readMaybe` functions will read an `Int` in decimal representation if it's not prefixed with `0x` or `0X`.

What I need here is a function that works like `read`, but it should treat the string as a hexadecimal `Int`, even if it's not prefixed with `0x`, that's why I implemented `toInt`.

Can you suggest something else that will do the job?

test/Test/Toml/Parsing/Unit.hs
ghallak: Instead of `parseText (squote3 "first line \nsecond.\n   3\n") "first line \nsecond.\n   3\n"`, I've tried to do this:
```haskell
parseText
[text|
'''first line
second.
    3
'''
|]
[text|
first line
second.
   3

|]
```
Now I'm getting a long error message but it starts with:
```
 • Couldn't match expected type ‘Data.Text.Internal.Text -> b0’
               with actual type ‘Data.Text.Internal.Text’
```
I'm gonna have to figure out this error before making the other changes. Any ideas about the cause of the error?

src/Toml/Parser.hs
ghallak: So, there should be a function that works like [`chr`](https://hackage.haskell.org/package/base-4.11.1.0/docs/Data-Char.html#v:chr) but fails if its argument is not a Unicode scalar value? I can ask about that.

test/Test/Toml/Parsing/Unit.hs
ChShersh: @ghallak Could you provide complete error message with actual code? Did you enable `QuasiQuotes`? Maybe you need to add extra spaces to arguments, not start them from beginning of line, like this:

```haskell
parseText
    [text|
    ...
    |]
    [text|
    ...
    |]
```

src/Toml/Parser.hs
ChShersh: Let's keep `basicStringP` and `multilineBasicStringP` as it is for now, and you will refactor this under your PR #71. So #71 will be a very easy PR 🙂

src/Toml/Parser.hs
ChShersh: @ghallak Yeah, you could ask about that. Good idea!

src/Toml/Parser.hs
```diff
@@ -89,10 +89,11 @@ quote q t = q <> t <> q
...
 
-multilineStringP :: Parser Text
-multilineStringP = lexeme $ fmap mconcat $ quotes3
-                                        >> optional eol
-                                        >> allowedChar `manyTill` quotes3
+multilineP :: Parser Text -> Parser Text -> Parser [Text]
+multilineP quotes allowedChar = quotes >> optional eol >> allowedChar `manyTill` quotes
+
+multilineBasicStringP :: Parser Text
+multilineBasicStringP = lexeme $ fmap mconcat $ multilineP quotes3 allowedChar
```
ChShersh: You can move `lexeme $ fmap mconcat $` into `multilineP` as well to remove even more code duplication.

src/Toml/Parser.hs
```diff
@@ -80,7 +80,7 @@ basicStringP :: Parser Text
 basicStringP = lexeme $ Text.pack <$> (char '"' *> anyChar `manyTill` char '"')
 
 textP :: Parser Text
-textP = multilineStringP <|> multilineLiteralStringP <|> literalStringP <|> basicStringP
+textP = multilineBasicStringP <|> multilineLiteralStringP <|> literalStringP <|> basicStringP
```
ChShersh: All these parsers are not backtracking. Which means that if you try to parse string like this:

```haskell
"foo"
```

It will try `multilineBasicStringP` first. This parser will consume single `"` and will fail because here `quotes >> optional eol` parser `quotes` is not under `try`. So my intuition that `textP` will fail on `"foo"` unless you add `try` to `quotes` parser. But if this currently works then it's a mystery to me 🤷‍♂️

src/Toml/Parser.hs
```diff
@@ -89,10 +89,11 @@ quote q t = q <> t <> q
 nonControlChar :: Parser Text
 nonControlChar = Text.singleton <$> satisfy (not . isControl)
 
-multilineStringP :: Parser Text
-multilineStringP = lexeme $ fmap mconcat $ quotes3
-                                        >> optional eol
-                                        >> allowedChar `manyTill` quotes3
+multilineP :: Parser Text -> Parser Text -> Parser [Text]
+multilineP quotes allowedChar = quotes >> optional eol >> allowedChar `manyTill` quotes
```
ChShersh: Okay, let's introduce even more renaming. Let's add `P` suffix to all parsers, even arguments and parsers under `where`. So it will be `quotesP`, `allowedCharP`, `notControlCharP`, etc.

src/Toml/Parser.hs
ChShersh: @ghallak Yeah, I see that we can't reuse `hexP` parser here. It's completely okay.

Okay, usage of `read` should be completely safe here. Let's not enable paranoia mode then. So you can write this function like this:

```haskell
hexToInt :: String -> Int
hexToInt xs = read $ "0x" ++ xs
```

src/Toml/Parser.hs
```diff
@@ -103,24 +104,23 @@ multilineStringP = lexeme $ fmap mconcat $ quotes3
...
+                      c    -> fail $ "Invalid escape sequence: " <> "\\" <> [c]
       where
         parseHexUnicode :: Int -> Parser Text
         parseHexUnicode n = count n hexDigitChar
-                        >>= \x -> case (toUnicode . toInt) x of
-                              Just c  -> return (Text.singleton c)
-                              Nothing -> fail ""
+                        >>= \x -> case toUnicode $ toInt x of
+                              Just c  -> pure (Text.singleton c)
+                              Nothing -> fail $ "Invalid unicode character: " <> "\\" <> (if n == 4 then "u" else "U") <> x
```
ChShersh: Very good error message!

test/Test/Toml/Parsing/Unit.hs
ghallak: Adding extra spaces to the arguments fixed the problem, but now I notice that:
```haskell
[text|
1
|]
```
is the same as:
```haskell
[text|
1


|]
```
so it's ignoring trailing newlines, and that will make it not possible to rewrite some of the tests.

Also, is there a way to insert a control character like `\b` for example in a `[text| ... |]` block? that's also required to rewrite some of the tests.

ChShersh: @ghallak Ah, shit, `neat-interploation` actually doesn't fit here... Okay, let's forget this idea and keep tests as it is now. I will think of better solution for this problem later. There's a workaround for trailing empty lines. But I dunno about control characters...

ghallak: What's the workaround for trailing empty lines? :thinking:

ChShersh: The idea is to have variable like `endLine = "\n"` (empty string probably will work as well) and then write like this:

```haskell
[text|


$endLine
|]
```

Also works if needed to keep indentation.

src/Toml/Parser.hs
ghallak: You can read the note here about [`try`](https://hackage.haskell.org/package/megaparsec-6.5.0/docs/Text-Megaparsec.html#v:try):

It says:
>Please note that as of Megaparsec 4.4.0, string backtracks automatically (see tokens), so it does not need try.

I'm using `string` in both `multilineBasicStringP` and `multilineLiteralStringP` and I'm counting on it to backtrack.

ChShersh: Ah, okay, `string` parser is an exception.

Issue #54, opened 2018-05-30
Support sum types in bidirectional conversion


ChShersh:
Let's say, we have a type like this:

```haskell
data Either l r = Left l | Right r
```

It should be possible to write bidirectional converter for such type type. Perfectly with the help of `Alternative` type class. I think in TOML this should look like this:

```toml
foo.Left = 42
bar.Right = true
```

Specifying both constructors should result in parsing error.

PR #65, opened 2018-06-21
[#60] Replace Valuer with Prism

### LINKED ISSUE #60: Rename 'Valuer' to 'SimplePrism'
And add more combinators (for something like `Either` if needed).

ChShersh:
Resolves #60 

I don't like `unsafeArray` part. But can't do nothing with it.

### DIFF
--- a/src/Toml/Prism.hs
+++ b/src/Toml/Prism.hs
@@ -53,7 +53,6 @@ instance Cat.Category Prism where
         { preview = preview objectPrism >=> preview fieldPrism
         , review = review objectPrism . review fieldPrism
         }
-    -- TODO: is naming too verbose?
 
 -- | Creates prism for 'AnyValue'.
 mkAnyValuePrism :: (forall t . Value t -> Maybe a)
--- a/src/Toml/Prism.hs
+++ b/src/Toml/Prism.hs
@@ -1,0 +1,104 @@
+{-# LANGUAGE DataKinds      #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE Rank2Types     #-}
+
+-- | Naive implementation of data-prism approach.
+
+module Toml.Prism
+       ( -- * Prism idea
+         Prism (..)
+       , match
+       , mkAnyValuePrism
+
+         -- * Value prisms
+       , _Bool
+       , _Integer
+       , _Double
+       , _String
+       , _Array
+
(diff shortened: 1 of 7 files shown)


src/Toml/Prism.hs
```diff
@@ -0,0 +1,104 @@
...
+instance Cat.Category Prism where
+    id :: Prism object object
+    id = Prism { preview = Just, review = id }
+
+    (.) :: Prism field subfield -> Prism object field -> Prism object subfield
+    fieldPrism . objectPrism = Prism
+        { preview = preview objectPrism >=> preview fieldPrism
+        , review = review objectPrism . review fieldPrism
+        }
+    -- TODO: is naming too verbose?
```
vrom911: Nope! It's better for understanding and it's not that noisy

ChShersh:
@vrom911 That's actually a good questions. Drawbacks of this implementation are pretty clear. But benefits are not so obvious so it's a good thing to elaborate more on them.

### +

1. **Idea.** I'm just personally happier in applying existing concept (or Haskell programming pattern in another words) instead of using some ad-hoc thing.
2. **Interface unification.** In future I want to support parsing of sum types (see issue #54). I imagine the following interface (but probably should be much better):

```haskell
eitherP :: BiToml (Either Smth Other)
eitherP = prismToml _Left  smthP
      <|> prismToml _Right otherP
```

So `Prism` gives us same interface for sum types and for array values. Which means that users need to learn less different things to use our library.

3. **Composability.** Since we have valid `Category` instance for `Prism` (and we couldn't have it for `Valuer`) we can actually compose our prisms to write more complex parsers using smaller ones as building blocks. So if we have types like this:

```haskell
data User = User { name :: String }

data Member = U User | A Admin
```

we could write parser like this:

```haskell
memberUserT :: BiToml Member
memberUserT = prismParser (_U . _User) string
```

And it TOML this should look like this:

```
someKey.U = "Vasya"
```

The idea is that we can reuse our components. Example might be not good since final interface is not specified. But I hope this gives some idea.

### –

1. **unsafeArray.** Should be more careful with `unsafeArray` function.
2. **Uniqueness.** This implementation of `Prism` is non-conventional and doesn't exist on Hackage anywhere. I'm not a category theory expert so I can't even claim that it's correct enough (though it looks pretty correct to me). And this `Prism` is not compatible with other `lens` ecosystem.

Issue #20, opened 2018-04-04
Add parsing of hex, oct and bin integer numbers


ChShersh:
TOML allows to specify numbers with `0x`, `0o` and `0b` prefixes. `megaparsec` has primitives for such things. We should enhance our integer parser.

PR #67, opened 2018-06-21
[#26] Implement unit testing for TOML parsing

### LINKED ISSUE #26: Implement mini-framework for unit testing TOML parsing
Pretty-printing doesn't require testing. Only single, that property-one. But we need to have some convenient way to test different parsing parts of TOML language. It would be great if we could parse everything from TOML repository. Need to think how to do this in convenient way...

Here are the things we need to test currently:

- [x] `Bool`
- [x] `Integer`
    - [x] Binary
    - [x] Octal
    - [x] Decimal
    - [x] Hexadecimal
- [ ] `Double`
- [x] `Array`
- [ ] Keys
- [ ] Key-value
- [ ] Table
- [ ] Some small complete toml

ghallak:
Resolves #26 

Tests are implemented for parsing the following types:

- `Integer`
- `Boolean`
- `Array`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@ The change log is available [on GitHub][2].
   Create EDSL for easier TOML data type writing.
 * [#10](https://github.com/kowainik/tomland/issues/10):
   Add `Semigroup` and `Monoid` instances for `PrefixTree` and `TOML`.
-  Add property tests on laws
+  Add property tests on laws.
+* [#20](https://github.com/kowainik/tomland/issues/20):
+  Add parsing of hexadecimal, octal, and binary integer numbers.
 * [#26](https://github.com/kowainik/tomland/issues/26):
   Implement unit tests for TOML parsers.
   Allow terminating commas inside an array.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,14 +4,6 @@ Change log
 tomland uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
-0.3.1
-=====
-
-* [#26](https://github.com/kowainik/tomland/issues/26):
-  Implement unit tests for `Integer`, `Boolean`, and `Array` parsers.
-  Allow terminating commas inside an array.
-  Allow comments before and after any value inside an array.
-
 0.3
 =====
 
@@ -20,6 +12,11 @@ The change log is available [on GitHub][2].
 * [#10](https://github.com/kowainik/tomland/issues/10):
   Add `Semigroup` and `Monoid` instances for `PrefixTree` and `TOML`.
   Add property tests on laws
+* [#26](https://github.com/kowainik/tomland/issues/26):
+  Implement unit tests for TOML parsers.
+  Allow terminating commas inside an array.
+  Allow comments before and after any value inside an array.
+  Allow keys to be literal strings.
 * **Breaking change:** [#60](https://github.com/kowainik/tomland/issues/60):
   Replace `Valuer` with `Prism`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@ Change log
 tomland uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
+0.3.1
+=====
+
+* [#26](https://github.com/kowainik/tomland/issues/26):
+  Implement unit tests for `Integer`, `Boolean`, and `Array` parsers.
+  Allow terminating commas inside an array.
+  Allow comments before and after any value inside an array.
+
 0.3
 =====
 
--- a/src/Toml/Parser.hs
+++ b/src/Toml/Parser.hs
@@ -10,7 +10,7 @@ module Toml.Parser
 
 -- I hate default Prelude... Do I really need to import all this stuff manually?..
 import Control.Applicative (Alternative (..))
-import Control.Applicative.Combinators (between, manyTill, sepBy)
+import Control.Applicative.Combinators (between, choice, manyTill, sepEndBy1, skipMany)
 import Control.Monad (void)
 import Data.Char (digitToInt)
 import Data.List (foldl')
@@ -18,7 +18,7 @@ import Data.Semigroup ((<>))
 import Data.Text (Text)
 import Data.Void (Void)
 import Text.Megaparsec (Parsec, parseErrorPretty', try)
-import Text.Megaparsec.Char (alphaNumChar, anyChar, char, oneOf, space, space1)
+import Text.Megaparsec.Char (alphaNumChar, anyChar, char, oneOf, space1)
 
 import Toml.PrefixTree (Key (..), Piece (..), fromList)
 import Toml.Type (AnyValue, TOML (..), UValue (..), typeCheck)
@@ -109,18 +109,35 @@ boolP = False <$ text "false"
 -- dateTimeP = error "Not implemented!"
 
 arrayP :: Parser [UValue]
-arrayP = lexeme $ between (char '[' *> space) (char ']') (valueP `sepBy` spComma)
+arrayP = lexeme $ between (char '[' *> sc) (char ']') elements
   where
+    elements :: Parser [UValue]
+    elements = (choice (fmap (\p -> p `sepEndBy1` spComma) ps) <|> pure []) <* skipMany (text ",")
+    ps :: [Parser UValue]
+    ps = [uBoolP, uFloatP, uIntP, uStringP, uArrayP]
     spComma :: Parser ()
-    spComma = char ',' *> space
+    spComma = char ',' *> sc
+
+uBoolP :: Parser UValue
+uBoolP = UBool <$> boolP
+
+uFloatP :: Parser UValue
+uFloatP = UFloat <$> try float
+
+uIntP :: Parser UValue
+uIntP = UInt <$> integerP
+
+uStringP :: Parser UValue
+uStringP = UString <$> (literalStringP <|> stringP)
+
+-- uDateP :: Parser UValue
+-- uDateP = UDate <$> dateTimeP
+
+uArrayP :: Parser UValue
+uArrayP = UArray <$> arrayP
 
 valueP :: Parser UValue
-valueP = UBool   <$> boolP
-     <|> UFloat  <$> try float
-     <|> UInt    <$> integerP
-     <|> UString <$> (literalStringP <|> stringP)
---     <|> UDate   <$> dateTimeP
-     <|> UArray  <$> arrayP
+valueP = uBoolP <|> uFloatP <|> uIntP <|> uStringP <|> uArrayP
 
 -- TOML
 
--- a/src/Toml/Parser.hs
+++ b/src/Toml/Parser.hs
@@ -3,9 +3,15 @@
 module Toml.Parser
        ( ParseException (..)
        , parse
-       , integerP
-       , boolP
        , arrayP
+       , boolP
+       , floatP
+       , integerP
+       , keyP
+       , keyValP
+       , stringP
+       , tableHeaderP
+       , tomlP
        ) where
 
 -- I hate default Prelude... Do I really need to import all this stuff manually?..
--- a/src/Toml/Parser.hs
+++ b/src/Toml/Parser.hs
@@ -53,8 +53,8 @@ text = L.symbol sc
 text_ :: Text -> Parser ()
 text_ = void . text
 
-float :: Parser Double
-float = L.signed sc $ lexeme L.float
+floatP :: Parser Double
+floatP = L.signed sc $ lexeme L.float
 
 ----------------------------------------------------------------------------
 -- TOML parser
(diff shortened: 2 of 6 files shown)


src/Toml/Parser.hs
```diff
@@ -106,18 +109,35 @@ boolP = False <$ text "false"
 -- dateTimeP = error "Not implemented!"
 
 arrayP :: Parser [UValue]
-arrayP = lexeme $ between (char '[' *> space) (char ']') (valueP `sepBy` spComma)
+arrayP = lexeme $ between (char '[' *> sc) (char ']') elements
   where
+    elements :: Parser [UValue]
+    elements = (choice (fmap (\p -> p `sepEndBy1` spComma) ps) <|> pure []) <* skipMany (text ",")
```
ChShersh: You don't need to implement parser in a such way to parse elements of single type. This is a part of typechecking:

https://github.com/kowainik/tomland/blob/9d44db6f56088c38d76a443b74defbd5ebd5245b/src/Toml/Parser.hs#L128-L131

And it was done in a such way to provide better error messages in future. With catching this on parser stage you can't give good error message without making parser scary. So it's okay if array parse can parse `"[true, 42]"`.

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -0,0 +1,146 @@
+{-# LANGUAGE FlexibleContexts #-}
+
+module Test.Toml.Parsing.Unit where
+
+import Test.Tasty.Hspec (Spec, context, describe, it)
+
+import Data.Text (singleton)
+import Test.Hspec.Megaparsec
```
ChShersh: All imports should be explicit.

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -0,0 +1,146 @@
+{-# LANGUAGE FlexibleContexts #-}
+
+module Test.Toml.Parsing.Unit where
+
+import Test.Tasty.Hspec (Spec, context, describe, it)
+
```
ChShersh: This empty line can be removed.

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -0,0 +1,146 @@
...
+      arrayFailOn = failOn arrayP
+      boolFailOn  = failOn boolP
+
+  describe "arrayP" $ do
+    it "can parse arrays" $ do
+      parseArray "[]" []
+      parseArray "[1]" [UInt 1]
+      parseArray "[1, 2, 3]" [UInt 1, UInt 2, UInt 3]
+      parseArray "[1.2, 2.3, 3.4]" [UFloat 1.2, UFloat 2.3, UFloat 3.4]
+      parseArray "['x', 'y']" [UString (singleton 'x'), UString (singleton 'y')]
```
ChShersh: `singleton 'x'` is just `"x"`.

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -0,0 +1,146 @@
...
+      parseArray "[[1], [2]]" [UArray [UInt 1], UArray [UInt 2]]
+      -- TODO: Add a test for array of UDate after implementing issue #18
```
ChShersh: Instead of writing this `TODO` you can use [`xit`](https://hackage.haskell.org/package/hspec-2.5.1/docs/Test-Hspec.html#v:xit) function `hspec` to mark this test as pending.

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -0,0 +1,146 @@
...
+  let parseX p given expected = parse p "" given `shouldParse` expected
+      failOn p given          = parse p "" `shouldFailOn` given
+
+  let parseArray  = parseX arrayP
+      parseBool   = parseX boolP
+      parseInt    = parseX integerP
+      arrayFailOn = failOn arrayP
+      boolFailOn  = failOn boolP
+
+  describe "arrayP" $ do
```
ChShersh: These array tests are really cool!

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -0,0 +1,146 @@
...
+        parseInt "-0" 0
+      it "can parse both the minimum and maximum numbers in the 64 bit range" $ do
+        parseInt "-9223372036854775808" (-9223372036854775808)
+        parseInt "9223372036854775807" 9223372036854775807
+      -- TODO: After implementing issue #64
+      --it "does not parse numbers with leading zeros" $ do
+      --  parseInt "0123" 0
+      --  parseInt "-023" 0
+      -- TODO: After implementing issue #17
+      --it "can parse numbers with underscores between digits" $ do
```
ChShersh: Again, those comments can be put under `xit`.

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -0,0 +1,146 @@
...
+        parseInt "0x00000" 0
+        parseInt "0x012af" 0x12af
+      it "does not parse negative numbers" $ do
+        parseInt "-0xfff" 0
+      it "does not parse numbers with non-valid hexadecimal digits" $ do
+        parseInt "0xfgh" 0xf
+      it "can parse numbers when hex digits are lowercase" $ do
+        parseInt "0xabcdef" 0xabcdef
+      it "can parse numbers when hex digits are uppercase" $ do
+        parseInt "0xABCDEF" 0xabcdef
```
ChShersh: Could you write capital letters in Haskell value here as well?

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -0,0 +1,146 @@
...
+      parseBool "false" False
+    it "fails if `true` or `false` are not all lowercase" $ do
+      boolFailOn "True"
+      boolFailOn "False"
+      boolFailOn "TRUE"
+      boolFailOn "FALSE"
+      boolFailOn "tRuE"
+      boolFailOn "fAlSE"
+
+  describe "integerP" $ do
```
ChShersh: Integer tests are really good as well!

ChShersh:
Also, I almost forgot: could you please update changelog as well?

src/Toml/Parser.hs
ghallak: The main reason for creating these extra functions (`uBoolP`, `uFloatP`, `uIntP`, `uStringP`, `uArrayP`), is to marke `arrayP` parse elements of a single type.

If it's okay for `arrayP` to parse elements of different types, then adding those functions would unnecessary, and I can go back to what we have in `master`:
https://github.com/kowainik/tomland/blob/9d44db6f56088c38d76a443b74defbd5ebd5245b/src/Toml/Parser.hs#L114-L120

test/Test/Toml/Parsing/Unit.hs
ghallak: @ChShersh Using `"x"` instead of `singleton 'x'` won't pass the test, because `UString` takes an argument of type `Text` and not `String`.
https://github.com/kowainik/tomland/blob/9d44db6f56088c38d76a443b74defbd5ebd5245b/src/Toml/Type.hs#L241

test/Test/Toml/Parsing/Unit.hs
ghallak: @ChShersh  Using [`xit`](https://hackage.haskell.org/package/hspec-2.5.1/docs/Test-Hspec.html#v:xit) will make the test fail with the following message: `# PENDING: No reason given`, and I was trying to avoid failed tests.

Should I use `xit` anyway?

src/Toml/Parser.hs
```diff
@@ -106,18 +109,35 @@ boolP = False <$ text "false"
...
+    ps :: [Parser UValue]
+    ps = [uBoolP, uFloatP, uIntP, uStringP, uArrayP]
     spComma :: Parser ()
-    spComma = char ',' *> space
+    spComma = char ',' *> sc
+
+uBoolP :: Parser UValue
+uBoolP = UBool <$> boolP
```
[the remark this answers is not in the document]
ChShersh: @ghallak Yeah, adding this test would be good 👍 This whole `lexeme`/`symbol` stuff in `megaparsec` is slightly confusing to me.

src/Toml/Parser.hs
```diff
@@ -106,18 +109,35 @@ boolP = False <$ text "false"
 -- dateTimeP = error "Not implemented!"
 
 arrayP :: Parser [UValue]
-arrayP = lexeme $ between (char '[' *> space) (char ']') (valueP `sepBy` spComma)
+arrayP = lexeme $ between (char '[' *> sc) (char ']') elements
```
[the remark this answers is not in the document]
ChShersh: @ghallak Ah, I see, so `sc` also consumes comments, not only spaces. Nice.

src/Toml/Parser.hs
ChShersh: @ghallak It's completely okay to allow arrays of different elements during `arrayP` parser. Because same-type validation is performed later in parsing.

test/Test/Toml/Parsing/Unit.hs
ChShersh: @ghallak Well, it's not actually failing, it's pending. Using `xit` is better because after every test we will see those tests. And we can just enable them later very easily once we implement feature itself.

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ Change log
 tomland uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
+0.3.1
```
ChShersh: Could you please move this to the end of version `0.3` in changelog? Version `0.3` is not released on Hackage yet. So it makes sense to put everything new there.

Also, you should add issue #20 to that changelog.

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -0,0 +1,308 @@
...
+        parseFloat "-0.01" (-0.01)
+      it "can parse a number which consists of an integral part, and an exponent part" $ do
+        parseFloat "5e+22" 5e+22
+        parseFloat "1e6" 1e6
+        parseFloat "-2E-2" (-2E-2)
+      it "can parse a number which consists of an integral, a fractional, and an exponent part" $ do
+        parseFloat "6.626e-34" 6.626e-34
+      it "can parse sign-prefixed zero" $ do
+        parseFloat "+0.0" 0.0
+        parseFloat "-0.0" (-0.0)
```
ChShersh: Very nice floating point tests!

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -0,0 +1,308 @@
...
+      it "can parse numbers when hex digits are uppercase" $ do
+        parseInt "0xABCDEF" 0xABCDEF
+      it "can parse numbers when hex digits are in both lowercase and uppercase" $ do
+        parseInt "0xAbCdEf" 0xAbCdEf
+        parseInt "0xaBcDeF" 0xaBcDeF
+
+  describe "keyP" $ do
+    context "when the key is a bare key" $ do
+      it "can parse keys which contain ASCII letters, digits, underscores, and dashes" $ do
+        parseKey "key"       (makeKey ["key"])
```
ChShersh: Actually, `makeKey` function is not needed since there exist `IsString` instance for key. So these tests could be written just like

```haskell
parseKey "key" "key"
```

But types of first and second arguments are different, so that's actually okay! If you prefer explicit style more, you can instead write:

```haskell
parseKey "key" ("key" :: Key)
```

ghallak: For this case:
```haskell
parseKey "site.\"google.com\"" "site.\"google.com\""
```
I get the following:
```
expected: Key {unKey = Piece {unPiece = "site"} :| [Piece {unPiece = "\"google"},Piece {unPiece = "com\""}]}
but got: Key {unKey = Piece {unPiece = "site"} :| [Piece {unPiece = "\"google.com\""}]}
```
`\"google.com\"` should be a single `Piece` and should not be split.

I've seen that this issue in `IsString` implementation was addressed here:
https://github.com/kowainik/tomland/blob/533e2d5efccb8e38922d38f116cded3c35a81be0/src/Toml/PrefixTree.hs#L67-L71

What do you suggest to do in this case?

ChShersh: @ghallak Ah, right, `IsString` instance is not supposed to work with quoted keys.

ghallak: So should I keep using `makeKey` ?

ChShersh: @ghallak It's on your choice now. In that case I'm okay if you will use `makeKey` to make tests look more consistent. But you also can use `makeKey` only for quoted keys.

ghallak: @ChShersh Will go with consistency this time. Using `makeKey` only for dotted keys which have quoted keys as part of them, will make things look strange.

PR #69, opened 2018-06-27
[#66] Introduce consistent naming

### LINKED ISSUE #66: Introduce consistent renaming + some refactoring
Current naming is not perfect and not consistent. At the beginning I tried to give names according to toml specification. But now I see that it's better to be closer to Haskell. Which brings to the following naming convention:

1. Constructors of `Value` and `ValueType` changes to `Integer`, `Text`, `Double` from `Int`, `String`, `Float`. And rename `ValueType` to `TValue`.
2. Each parser should have `P` suffix.
3. Each prism should be capitalised and start with underscore.
4. Each bidirectional converter should be name of constructor without any suffices. When name clashes with `Prelude` name it should have `T` suffix (convention for all `BiToml` configurations).

Also, I want to change module structure a bit to make it more modular:

Split `Toml.Types` module into `Toml.Types.Value`, `Toml.Types.AnyValue`, `Toml.Types.TOML`, `TOML.Types.UValue`. This change is also motivated by fact that `Toml.Types` module becomes very big.

Let me know, what do you think!

ChShersh:
Resolves #66


src/Toml/Type/Value.hs
```diff
@@ -0,0 +1,210 @@
...
+arr1 = [ 1, 2, 3 ]
+arr2 = [ "red", "yellow", "green" ]
+arr3 = [ [ 1, 2 ], [3, 4, 5] ]
+arr4 = [ "all", 'strings', """are the same""", '''type''']
+arr5 = [ [ 1, 2 ], ["a", "b", "c"] ]
+
+arr6 = [ 1, 2.0 ] # INVALID
+@
+    -}
+    Array  :: [Value t] -> Value 'TArray
```
ghallak: Since this is implemented as a Haskell list, should it be changed to `List` instead of `Array`?

src/Toml/Parser.hs
```diff
@@ -127,10 +127,10 @@ arrayP = lexeme $ between (char '[' *> sc) (char ']') elements
     spComma = char ',' *> sc
 
 valueP :: Parser UValue
-valueP = UBool   <$> boolP
-     <|> UFloat  <$> try floatP
-     <|> UInt    <$> integerP
-     <|> UString <$> stringP
+valueP = UBool <$> boolP
```
ghallak: It would be nice if the `<$>` operators are aligned here, as they were before.

test/Test/Toml/Gen.hs
```diff
@@ -54,13 +54,13 @@ genAnyValue :: MonadGen m => m AnyValue
 genAnyValue = do
   let randB = Gen.bool
   let randI = toInteger <$> Gen.int (Range.constantBounded @Int)
-  let randF = Gen.double $ Range.constant @Double (-1000000.0) 1000000.0
+  let randD = Gen.double $ Range.constant @Double (-1000000.0) 1000000.0
   let randS = Gen.text (Range.constant 0 256) Gen.alphaNum
```
ghallak: Shouldn't this be `randT` now instead of `randS`?

src/Toml/Type/Value.hs
ChShersh: I would like to keep it as `Array` by the following reasons:

1. In Haskell types like `Int`, `String` and `Float` are very common and well-known and are different from `Integer`, `Text` and `Double` (which are also well-known Haskell types). This leads to confusion in using data types. But there's no such type like `Array` (at least commonly-used). So it's okay to be closer to toml specification here.
2. In future internal representation of `Array` constructor might change to something like `Vector` if we discover some serious performance problems. So it no longer will be `List` and having `Array` will help with backwards compatibility. 
3. `Text`, `Integer` and `Double` in Haskell are semantically exactly equivalent to toml specification. While lists are not semantically equivalent. In toml you can have an array like `[ [1, 2], [true, false] ]` but in Haskell you can't. So having `Array` name for variable hints user that it's different from Haskell lists.

PR #71, opened 2018-06-28
[#40] Support full-featured string parser

### LINKED ISSUE #40: Support full-featured string parser

ghallak:
Resolves #40

### DIFF
--- a/src/Toml/Parser.hs
+++ b/src/Toml/Parser.hs
@@ -64,30 +64,52 @@ text_ = void . text
 -- TOML parser
 ----------------------------------------------------------------------------
 
--- Keys
-
-bareKeyP :: Parser Text
-bareKeyP = lexeme $ Text.pack <$> bareStrP
-  where
-    bareStrP :: Parser String
-    bareStrP = some $ alphaNumChar <|> char '_' <|> char '-'
-
-literalStringP :: Parser Text
-literalStringP = lexeme $ Text.pack <$> (char '\'' *> anyChar `manyTill` char '\'')
-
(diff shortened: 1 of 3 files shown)


src/Toml/Parser.hs
```diff
@@ -64,31 +64,53 @@ text_ = void . text
...
+          | otherwise                    = Nothing
+
+basicStringP :: Parser Text
+basicStringP = lexeme $ mconcat <$> (char '"' *> (escapeSequenceP <|> nonControlCharP) `manyTill` char '"')
+
+literalStringP :: Parser Text
+literalStringP = lexeme $ Text.pack <$> (char '\'' *> nonEolCharP `manyTill` char '\'')
+  where
+    nonEolCharP :: Parser Char
+    nonEolCharP = satisfy (\c -> c /= '\n' && c /= '\r')
```
ghallak: I don't like this line, I want this parser to fail only when [`eol`](https://hackage.haskell.org/package/megaparsec-6.5.0/docs/Text-Megaparsec-Char.html#v:eol) succeeds.

@ChShersh I couldn't think of an easier way to do that, do you have any ideas?

ChShersh: @ghallak No, I don't know better way. You could ask on `megaparsec` issue tracker about this. Usually parser combinator libraries have inversions of parsers, like this:

```haskell
-- | Succeeds only when given parser fails.
notP :: Parser a -> Parser (Token s)
```

PR #74, opened 2018-07-14
[#18] Add dates parsing

### LINKED ISSUE #18: Add dates parsing
Parsing dates is hard :disappointed: But we really need them.

ghallak:
Resolves #18

### DIFF
--- a/src/Toml/Parser.hs
+++ b/src/Toml/Parser.hs
@@ -20,11 +20,12 @@ import Control.Applicative (Alternative (..))
 import Control.Applicative.Combinators (between, count, manyTill, optional, sepEndBy, skipMany)
 import Control.Monad (void)
 import Data.Char (chr, digitToInt, isControl)
+import Data.Fixed (Pico)
 import Data.List (foldl')
 import Data.Semigroup ((<>))
 import Data.Text (Text)
-import Data.Time (Day, LocalTime, TimeOfDay, ZonedTime)
-import Data.Time.Format.ISO8601 (iso8601ParseM)
+import Data.Time (LocalTime (..), ZonedTime (..), fromGregorianValid, makeTimeOfDayValid,
+                  minutesToTimeZone)
 import Data.Void (Void)
 import Text.Megaparsec (Parsec, parseErrorPretty', try)
 import Text.Megaparsec.Char (alphaNumChar, anyChar, char, digitChar, eol, hexDigitChar, oneOf,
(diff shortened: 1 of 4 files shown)


test/Test/Toml/Parsing/Unit.hs
```diff
@@ -20,28 +22,38 @@ spec_Parser = do
...
+      keyValFailOn   = failOn keyValP
+      textFailOn     = failOn textP
 
       doubleSatisfies = parseXSatisfies doubleP
 
       quoteWith q t = q <> t <> q
       squote        = quoteWith "'"
       dquote        = quoteWith "\""
 
+      makeDay year month day            = Day $ fromGregorian year month day
```
ghallak: @ChShersh I have tried here to do: `makeDay = Day . fromGregorian`, but it doesn't work because `fromGregorian` takes accepts 3 arguments.

Is there a workaround that allows doing `makeDay = Day . fromGregorian` ?

ChShersh: @ghallak There exist trick for variadic composition operator. We implemented it in `universum`:

* https://github.com/serokell/universum/blob/master/src/Universum/VarArg.hs

But with usual composition operator you can't do that. So `makeDay year month day = Day $ fromGregorian year month day`

ChShersh:
Okay, this PR looks quite bid and complicated. I will try to review it tomorrow...

src/Toml/Parser.hs
```diff
@@ -193,10 +261,10 @@ arrayP = lexeme $ between (char '[' *> sc) (char ']') elements
 
 valueP :: Parser UValue
 valueP = UBool    <$> boolP
+     <|> UDate    <$> dateTimeP
```
ChShersh: Since `dateTimeP` parser is going to be very slow, it's better to move it at the end of `valueP` function.

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -20,28 +22,38 @@ spec_Parser = do
...
+      makeHours hour minute second      = Hours $ TimeOfDay hour minute second
+      makeLocal (Day day) (Hours hours) = Local $ LocalTime day hours
+      makeLocal _          _            = undefined
```
ChShersh: I don't see why pattern-matching is not exhaustive here... Anyway, it's better to avoid `undefined`. Or at least put `error`

ghallak: Since `LocalTime` accepts two arguments of type `Day` and `TimeOfDay` (or `Local` accepts `Day` and `Hours`), there can't be a result if `makeLocal` accepts arguments of different types, `Zoned` and `Local` for example.

I can still switch to using `error` instead of `undefined`.

src/Toml/Parser.hs
ghallak: If `integerP` is tried before `dateTimeP` it will parse a part of the date as an integer, such in this case: "12:24:14" (`integerP` will parse `12`).

I was thinking that `doubleP` should also be tried before `dateTimeP`, but I can't think of such case now.

So it's possible to move `dateTimeP` to the end of `valueP` as long as `integerP` is after it.

ChShersh: @ghallak Hmm. that's a problem indeed. Okay, let's have dates earlier. Maybe we will figure out something later.

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -343,8 +343,6 @@ spec_Parser = do
       parseDateTime "1979-05-27T07:32:00Z"             (makeZoned (makeLocal (makeDay 1979 5 27) (makeHours 7 32 0)) (makeOffset 0 0))
       parseDateTime "1979-05-27T00:32:00+07:10"        (makeZoned (makeLocal (makeDay 1979 5 27) (makeHours 0 32 0)) (makeOffset 7 10))
       parseDateTime "1979-05-27T00:32:00.999999-07:25" (makeZoned (makeLocal (makeDay 1979 5 27) (makeHours 0 32 0.999999)) (makeOffset (-7) 25))
-    it "can parse a date-time with an offset when the T delimiter is replaced with a space" $ do
-      parseDateTime "1979-05-27 07:32:00Z" (makeZoned (makeLocal (makeDay 1979 5 27) (makeHours 7 32 0)) (makeOffset 0 0))
```
ghallak: This test is deleted because the current implementation doesn't support replacing the 'T' delimiter with a space.

stack.yaml
```diff
@@ -1 +1,4 @@
 resolver: lts-12.0
+extra-deps:
+- time-1.9.1
+allow-newer: true
```
ghallak: Not sure that this is the right thing to do, but this line is needed to run the tests.

Without it, I get `Plan construction failed` when I try to run `stack test`.

PR #75, opened 2018-07-15
Add `Alternative` instance for `Bijection`

ChShersh:
This is a first step towards supporting sum types. Not sure that I'm moving in the right direction... What I want is to have `Alternative` instance for `Bijection`.

### DIFF
--- a/src/Toml/Bi/Code.hs
+++ b/src/Toml/Bi/Code.hs
@@ -19,6 +19,7 @@ import Control.Monad.State (State, execState)
 import Control.Monad.Trans.Maybe (MaybeT (..))
 import Data.Bifunctor (first)
 import Data.Foldable (toList)
+import Data.Semigroup (Semigroup (..))
 import Data.Semigroup ((<>))
 import Data.Text (Text)
 
@@ -41,7 +42,6 @@ data DecodeException
 
 instance Semigroup DecodeException where
     TrivialError <> e = e
-    e <> TrivialError = e
     e <> _ = e
 
 instance Monoid DecodeException where
@@ -66,7 +66,7 @@ prettyException = \case
 type Env = ExceptT DecodeException (Reader TOML)
 
 {- | Mutable context for 'Toml' conversion.
--- This is @w@ type variable in 'Bijection' data type.
+This is @w@ type variable in 'Bijection' data type.
 
 @
 MaybeT (State TOML) a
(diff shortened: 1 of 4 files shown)


src/Toml/Bi/Code.hs
```diff
@@ -67,6 +86,7 @@ decode biToml text = do
     toml <- first ParseError (parse text)
     runReader (runExceptT $ biRead biToml) toml
 
--- | Convert object to textual representation.
+-- | Convert object to textual representation. Outputs empty string if for some
+-- reason not able to convert to 'Text'.
```
ChShersh: Oops, this comment is wrong. It doesn't output empty string.

src/Toml/Bi/Code.hs
```diff
@@ -31,15 +32,26 @@ import qualified Data.Text as Text
...
+    = TrivialError
+    | KeyNotFound Key  -- ^ No such key
     | TableNotFound Key  -- ^ No such table
     | TypeMismatch Key Text TValue  -- ^ Expected type vs actual type
     | ParseError ParseException  -- ^ Exception during parsing
     deriving (Eq, Show)  -- TODO: manual pretty show instances
 
+instance Semigroup DecodeException where
+    TrivialError <> e = e
+    e <> TrivialError = e
```
vrom911: I guess this case is not needed

src/Toml/Bi/Code.hs
```diff
@@ -53,9 +65,16 @@ prettyException = \case
 -- This is @r@ type variable in 'Bijection' data type.
 type Env = ExceptT DecodeException (Reader TOML)
 
--- | Mutable context for 'Toml' conversion.
+{- | Mutable context for 'Toml' conversion.
 -- This is @w@ type variable in 'Bijection' data type.
```
vrom911: You can remove `--` from this line

src/Toml/Bi/Code.hs
```diff
@@ -53,9 +65,16 @@ prettyException = \case
...
-type St = State TOML
+
+@
+MaybeT (State TOML) a
+    = State TOML (Maybe a)
+    = TOML -> (Maybe a, TOML)
```
vrom911: Cool note!

src/Toml/Bi/Code.hs
ChShersh: @vrom911 You're right. And removing this case is actually a good improvement, because it will make implementation lazier.

PR #77, opened 2018-07-29
Add bi combinator for newtypes

ChShersh:


src/Toml/Bi/Combinators.hs
```diff
@@ -201,3 +203,7 @@ table bi key = Bijection input output
     handleTableName (TableNotFound name)      = throwError $ TableNotFound (key <> name)
     handleTableName (TypeMismatch name t1 t2) = throwError $ TypeMismatch (key <> name) t1 t2
     handleTableName e                         = throwError e
+
+-- | Used for @newtype@ wrappers.
+wrapper :: forall b a . Coercible a b => (Key -> BiToml a) -> Key -> BiToml b
+wrapper bi key = dimap coerce coerce (bi key)
```
vrom911: This is so clever and useful! I've noticed that this is used so often.

PR #79, opened 2018-08-11
Minor refactoring in preparation for sum types

ChShersh:
This PR just introduces some functions. It's very small. I still don't have full vision of how sum types should be implemented. But I'm slightly moving in this direction.

### DIFF
--- a/src/Toml/Bi/Monad.hs
+++ b/src/Toml/Bi/Monad.hs
@@ -6,6 +6,7 @@ module Toml.Bi.Monad
        ( Bijection (..)
        , Bi
        , dimap
+       , (<!>)
        , (.=)
        ) where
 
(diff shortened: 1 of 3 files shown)


src/Toml/Bi/Monad.hs
```diff
@@ -89,6 +90,12 @@ instance (MonadPlus r, MonadPlus w) => MonadPlus (Bijection r w c) where
     mzero = empty
     mplus = (<|>)
 
+infixl 3 <!>
+-- | Alternative instance for function arrow but without 'empty'.
+(<!>) :: Alternative f => (a -> f x) -> (a -> f x) -> (a -> f x)
+f <!> g = \a -> f a <|> g a
+
+
```
ghallak: This looks like an extra empty line.

PR #80, opened 2018-08-18
[#54] Support sum types

### LINKED ISSUE #54: Support sum types in bidirectional conversion

ChShersh:
Resolves #54 

This PR adds support for sum types. Please, share your feedback regarding new design (see example). And let me know what do you think.


src/Toml/Bi/Combinators.hs
```diff
@@ -56,25 +54,24 @@ import qualified Toml.PrefixTree as Prefix
...
--- | General function to create bidirectional converters for values.
-bijectionMaker :: forall a t . Typeable a
-               => (forall f . Value f -> Maybe a)  -- ^ How to convert from 'AnyValue' to @a@
-               -> (a -> Value t)                   -- ^ Convert @a@ to 'Anyvale'
-               -> Key                              -- ^ Key of the value
-               -> BiToml a
-bijectionMaker fromVal toVal key = Bijection input output
+{- | General function to create bidirectional converters for values.
+-}
+match :: forall a . Typeable a => BiMap AnyValue a -> Key -> BiToml a
```
vrom911: Can you also add this renaming to the CHANGELOG?

examples/Playground.hs
```diff
@@ -42,6 +45,19 @@ testT = Test
...
+    -- different keys for sum type
+    eitherT1 :: BiToml (Either Integer String)
+    eitherT1 = Toml.match (Toml._Integer >>> Toml._Left)  "either.Left"
+           <|> Toml.match (Toml._String  >>> Toml._Right) "either.Right"
+
+    -- same key for sum type
+    eitherT2 :: BiToml (Either String Double)
+    eitherT2 = ( Toml.match (Toml._String >>> Toml._Left)
+             <!> Toml.match (Toml._Double >>> Toml._Right)
+               ) "either"
```
vrom911: Probably should write about the drawbacks of this case (like when the types of `Left` and `Right` are the same it would take the first one)?

ChShersh: Yeah, having same types in constructors of sum types is something that can be improved (but I don't have a good solution at the moment). Current implementation basically tries each constructor one by one. It would be good to have something like _chose constructor based on value_ but I don't know how to support this for now...

PR #83, opened 2018-08-30
Fix Haddock for Toml.Edsl

neongreen:
`"Oh"` is parsed as a module name.


neongreen:
I think that somewhere you also have unescaped `<$>`, `<*>` and `"Kowainik"`, but I have to go back to work now.

ChShersh:
@neongreen Thanks for fixes! I'm aware of this problem, I just haven't enough time to fix Haddock. It probably will change a lot since library in experimental state.

PR #84, opened 2018-08-31
[#85] Add date test generator

willbasky:
Resolves [#85](https://github.com/kowainik/tomland/issues/85)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The change log is available [on GitHub][2].
 =====
 
 * [#72](https://github.com/kowainik/tomland/issues/72):
-  Improve 'AnyValue' generator for property-based tests.
+  Add `Data` generator for property-based tests.
 
 0.4.0
 =====
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,8 @@ The change log is available [on GitHub][2].
 0.4.1
 =====
 
-* [#72](https://github.com/kowainik/tomland/issues/72):
-  Add `Data` generator for property-based tests.
+* [#85](https://github.com/kowainik/tomland/issues/85):
+  Add `Date` generator for property-based tests.
 
 0.4.0
 =====
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Change log
 tomland uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
+0.4.1
+=====
+
+* [#72](https://github.com/kowainik/tomland/issues/72):
+  Improve 'AnyValue' generator for property-based tests.
+
 0.4.0
 =====
 
--- a/src/Toml/Printer.hs
+++ b/src/Toml/Printer.hs
@@ -80,7 +80,7 @@ prettyKeyValue i = Text.concat . map kvText . HashMap.toList
     valText (Array a)   = "[" <> Text.intercalate ", " (map valText a) <> "]"
 
     timeText :: DateTime -> Text
-    timeText (Zoned z) = showZoneTime z
+    timeText (Zoned z) = showZonedTime z
     timeText (Local l) = showText l
     timeText (Day d)   = showText d
     timeText (Hours h) = showText h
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ Change log
 tomland uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
+0.4.1
```
ChShersh: You also need to bump up package version in `tomland.cabal` file

src/Toml/Printer.hs
```diff
@@ -78,14 +80,21 @@ prettyKeyValue i = Text.concat . map kvText . HashMap.toList
...
-    timeText (Zoned z) = showText z
+    timeText (Zoned z) = showZoneTime z
     timeText (Local l) = showText l
     timeText (Day d)   = showText d
     timeText (Hours h) = showText h
 
     showText :: Show a => a -> Text
     showText = Text.pack . show
 
+    showZoneTime :: ZonedTime -> Text
```
ChShersh: I would name this function as `showZonedTime`

src/Toml/Printer.hs
```diff
@@ -78,14 +80,21 @@ prettyKeyValue i = Text.concat . map kvText . HashMap.toList
...
+    showZoneTime 
+        = Text.pack 
+        . (\(x,y) -> x ++ ":" ++ y) 
+        . (\z -> splitAt (length z - 2) z)
+        . formatTime defaultTimeLocale "%FT%T%Q%z"
```
ChShersh: The implementation looks a bit hacky. I understand that there's a reason for that. But I would split the implementation into two functions:

```haskell
showZonedTime t = showZonedDateTime t <> showZonedZone t
```

And then you can perform dirty hacks inside `showZonedZone` function only that contains `formatTime "%z"`.

test/Test/Toml/Gen.hs
```diff
@@ -56,11 +58,19 @@ genAnyValue = do
...
+  let randL = genLocal 
   Gen.choice
       [ AnyValue . Bool    <$> randB
       , AnyValue . Integer <$> randI
       , AnyValue . Double  <$> randD
       , AnyValue . Text    <$> randT
+      , AnyValue . Date    <$> randZ 
+      , AnyValue . Date    <$> randL
+      , AnyValue . Date    <$> randDay
+      , AnyValue . Date    <$> randH
```
ChShersh: Let's group these 4 generators into single function `genDate`.

test/Test/Toml/Gen.hs
```diff
@@ -56,11 +58,19 @@ genAnyValue = do
   let randI = toInteger <$> Gen.int (Range.constantBounded @Int)
   let randD = Gen.double $ Range.constant @Double (-1000000.0) 1000000.0
   let randT = Gen.text (Range.constant 0 256) Gen.alphaNum
```
ChShersh: Also, let's perform some renaming. It's more conventional to name generators as `genSomething`.

test/Test/Toml/Gen.hs
```diff
@@ -119,3 +129,33 @@ genToml = do
     kv     <- HashMap.fromList <$> genKeyAnyValueList
     tables <- Gen.list (Range.linear 0 10) genTableHeader
     pure $ TOML kv (fromList tables)
+
+genDay :: MonadGen m => m DateTime
```
ChShersh: All these generators return `DateTime`. It's better if instead of they return types of `DateTime` field constructors and then you can write function:

```
genDate = Gen.choice [Day <$> getDay, Hours <$> genHours, ...]
```

test/Test/Toml/Gen.hs
```diff
@@ -119,3 +129,33 @@ genToml = do
...
+genLocal = do
+    Day day <- genDay
+    Hours time <- genHours
+    let local = LocalTime day time
+    pure $ Local local
+
+genZoned :: MonadGen m => m DateTime
+genZoned = do
+    Local local <- genLocal
+    zMin <- Gen.int (Range.constant 1 720)
```
ChShersh: Looks like this generator can take negative minutes. It's totally valid to have negative timezone like `-0800`.

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ The change log is available [on GitHub][2].
 =====
 
 * [#72](https://github.com/kowainik/tomland/issues/72):
```
ChShersh: You should change issue number here as well.

CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ The change log is available [on GitHub][2].
...
-  Improve 'AnyValue' generator for property-based tests.
+  Add `Data` generator for property-based tests.
```
ChShersh: It's `Date`, not `Data`.

PR #89, opened 2018-09-03
[#88] Add Array test generator

willbasky:
Resolves [#88](https://github.com/kowainik/tomland/issues/88)

### DIFF
--- a/test/Test/Toml/Gen.hs
+++ b/test/Test/Toml/Gen.hs
@@ -58,17 +58,10 @@ genVal = Gen.int (Range.constant 0 256)
 -- TODO: Arrays and Date.
 -- | Generates random value of 'AnyValue' type.
 genAnyValue :: MonadGen m => m AnyValue
-genAnyValue = Gen.choice
-    [ AnyValue . Bool    <$> genBool
-    , AnyValue . Integer <$> genInteger
-    , AnyValue . Double  <$> genDouble
-    , AnyValue . Text    <$> genText
-    , AnyValue . Date    <$> genDate
-    , AnyValue <$> genArray
-    , AnyValue <$> genArrayList
-    ]
+genAnyValue = Gen.choice $
+    (AnyValue <$> genArray) : noneArrayList
 
--- TODO: unicode support
+    -- TODO: unicode support
 genPiece :: MonadGen m => m Piece
 genPiece = Piece <$> Gen.text (Range.constant 1 50) Gen.alphaNum
 
@@ -171,6 +164,7 @@ genDouble = Gen.double $ Range.constant @Double (-1000000.0) 1000000.0
 genText :: MonadGen m => m Text
 genText = Gen.text (Range.constant 0 256) Gen.alphaNum
 
+-- | List of AnyValue generators.
 noneArrayList :: MonadGen m => [m AnyValue]
 noneArrayList =
     [ AnyValue . Bool    <$> genBool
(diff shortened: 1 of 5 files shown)


test/Test/Toml/Gen.hs
```diff
@@ -150,9 +151,55 @@ genZoned = do
...
+
+genListTextR :: MonadGen m =>  m [Value 'TText]
+genListTextR = Gen.recursive Gen.choice [genListText] [genListTextR]
+  where 
+    genListText = Gen.list (Range.constant 1 10) $ Text <$> genText
+
+genListDateR :: MonadGen m => m [Value 'TDate]
+genListDateR = Gen.recursive Gen.choice [genListDate] [genListDateR]
+  where
+    genListDate = Gen.list (Range.constant 1 10) $ Date <$> genDate
```
ChShersh: Current implementation is too verbose and contains a lot of boilerplate. I propose the following solution:

1. Have `genNonArray` function which is like `genAnyValue` but without `genArray`.
2. Implement `genArray` function (using `genNonArray` generator) that generates list of `AnyValue` and then unsafely typechecks this list and creates `Array`. You should use `recursive` combinator for this.
3. Implemeng `genAnyValue` which is a combination of `genNonArray` and `genArray`.

ChShersh: So, basically you now have more code than needed. And you don't use `recursive` in `genArray` (thus you can't generate nested arrays yet).

I imagine the following implementation:

```haskell
genArray = Gen.recursive Gen.choice [genArrayFromAnyValueList] [genArray]
```

ChShersh:
The CI error looks strange to me...

@willbasky Did you manage to figure out how to run generators in GHCi to see the result?

test/Test/Toml/Gen.hs
```diff
@@ -53,18 +58,15 @@ genVal = Gen.int (Range.constant 0 256)
...
-      , AnyValue . Date    <$> genDate
-      ]
+genAnyValue = Gen.choice
+    [ AnyValue . Bool    <$> genBool
+    , AnyValue . Integer <$> genInteger
+    , AnyValue . Double  <$> genDouble
+    , AnyValue . Text    <$> genText
+    , AnyValue . Date    <$> genDate
+    , AnyValue <$> genArray
+    , AnyValue <$> genArrayList
```
ChShersh: `genArrayList` shouldn't be needed here actually. `genArray` generator should be able to generate either arrays of non-arrays (like array of integers) or arrays of arrays. `genArray` generator should be implemented in a such way that it produces any possible valid `Array`.

test/Test/Toml/Gen.hs
```diff
@@ -150,9 +152,41 @@ genZoned = do
...
+    , AnyValue . Text    <$> genText
+    , AnyValue . Date    <$> genDate
+    ]
+
+genArrayFrom :: MonadGen m => m AnyValue -> m (Value 'TArray)
+genArrayFrom noneArray = fromMaybe (error "Error in genArrayFrom") . toMArray <$> Gen.list (Range.constant 3 6) noneArray
+
+genArray :: MonadGen m => m (Value 'TArray)
+genArray = Gen.recursive Gen.choice
+    [Gen.choice $ map genArrayFrom noneArrayList]
```
ChShersh: I think that this argument of `recursive` function should be `map genArrayFrom noneArrayList` instead of `[Gen.choice $ map genArrayFrom noneArrayList]`. This argument contains list of non-recursive generators. In case of array generation, it should be either array of integers or array of doubles or array of text, etc. Don't need to apply `choice` additionally.

test/Test/Toml/Gen.hs
```diff
@@ -53,18 +58,15 @@ genVal = Gen.int (Range.constant 0 256)
...
-      , AnyValue . Double  <$> genDouble
-      , AnyValue . Text    <$> genText
-      , AnyValue . Date    <$> genDate
-      ]
+genAnyValue = Gen.choice
+    [ AnyValue . Bool    <$> genBool
+    , AnyValue . Integer <$> genInteger
+    , AnyValue . Double  <$> genDouble
+    , AnyValue . Text    <$> genText
+    , AnyValue . Date    <$> genDate
```
ChShersh: You can reuse `nonArrayList` value here to avoid code duplication.

test/Test/Toml/Gen.hs
```diff
@@ -150,9 +152,41 @@ genZoned = do
...
+noneArrayList =
+    [ AnyValue . Bool    <$> genBool
+    , AnyValue . Integer <$> genInteger
+    , AnyValue . Double  <$> genDouble
+    , AnyValue . Text    <$> genText
+    , AnyValue . Date    <$> genDate
+    ]
+
+genArrayFrom :: MonadGen m => m AnyValue -> m (Value 'TArray)
+genArrayFrom noneArray = fromMaybe (error "Error in genArrayFrom") . toMArray <$> Gen.list (Range.constant 3 6) noneArray
```
ChShersh: Constant here should be from 0 to 10.

test/Test/Toml/Gen.hs
```diff
@@ -150,9 +152,41 @@ genZoned = do
...
+
+genArray :: MonadGen m => m (Value 'TArray)
+genArray = Gen.recursive Gen.choice
+    [Gen.choice $ map genArrayFrom noneArrayList]
+    [genArray]
```
ChShersh: It would be great to show results of `genArray` generator. It's not obvious from code that it can generate arrays of arrays. But it should. I think it's not enough to have only `[genArray]` here. I guess the last argument of `recursive` should be something smarter.

test/Test/Toml/Gen.hs
willbasky: It does`t work without `Gen.choice` that giving `[m a]`, otherwise incorrect `[[ma]]`.
`recursive` allow to pick up all values from `noneArrayList`. 
`Print` shows (new line = new start)
```
Array [Bool True,Bool True,Bool True]
Array [Date (Zoned 1974-04-03 22:04:00.000000000051 -0751),Date (Day 2018-04-06),Date (Zoned 2005-04-22 21:14:00.000000000024 -0628)]
and so on
```
So, `genArray` generates any value arrays.

test/Test/Toml/Gen.hs
willbasky: Where to show results, here or in code comments?

test/Test/Toml/Gen.hs
willbasky: 10 is too many. Tests go long.

ChShersh: Okay, if it's too long, then let's keep 6.

test/Test/Toml/Gen.hs
ChShersh: Just here. No need to put under comments. Just want to make sure that everything works as expected.

ChShersh:
@willbasky You can leave it. There's no harm in the comment if you already wrote it.

PR #92, opened 2018-09-16
[#87] Improve double generator

willbasky:
Resolves [#87](https://github.com/kowainik/tomland/issues/87)

### DIFF
--- a/src/Toml/Printer.hs
+++ b/src/Toml/Printer.hs
@@ -74,7 +74,7 @@ prettyKeyValue i = Text.concat . map kvText . HashMap.toList
     valText :: Value t -> Text
     valText (Bool b)    = Text.toLower $ showText b
     valText (Integer n) = showText n
-    valText (Double d)  = showText d
+    valText (Double d)  = showDouble d
     valText (Text s)    = showText s
     valText (Date d)    = timeText d
     valText (Array a)   = "[" <> Text.intercalate ", " (map valText a) <> "]"
@@ -88,12 +88,21 @@ prettyKeyValue i = Text.concat . map kvText . HashMap.toList
     showText :: Show a => a -> Text
     showText = Text.pack . show
 
+    showDouble :: Double -> Text
+    showDouble d | isInfinite d && isNegate d = "-inf"
+                 | isInfinite d = "inf"
+                 | isNaN d = "nan"
+                 | otherwise = showText d
+      where
+        isNegate :: Double -> Bool
+        isNegate x = take 1 (show x) == "-"
+
     showZonedTime :: ZonedTime -> Text
     showZonedTime t = Text.pack $ showZonedDateTime t <> showZonedZone t
-      where 
+      where
         showZonedDateTime = formatTime defaultTimeLocale "%FT%T%Q"
-        showZonedZone 
-            = (\(x,y) -> x ++ ":" ++ y) 
+        showZonedZone
+            = (\(x,y) -> x ++ ":" ++ y)
             . (\z -> splitAt (length z - 2) z)
             . formatTime defaultTimeLocale "%z"
 
--- a/src/Toml/Printer.hs
+++ b/src/Toml/Printer.hs
@@ -89,13 +89,10 @@ prettyKeyValue i = Text.concat . map kvText . HashMap.toList
     showText = Text.pack . show
 
     showDouble :: Double -> Text
-    showDouble d | isInfinite d && isNegate d = "-inf"
+    showDouble d | isInfinite d && d < 0 = "-inf"
                  | isInfinite d = "inf"
                  | isNaN d = "nan"
                  | otherwise = showText d
-      where
-        isNegate :: Double -> Bool
-        isNegate x = take 1 (show x) == "-"
 
     showZonedTime :: ZonedTime -> Text
     showZonedTime t = Text.pack $ showZonedDateTime t <> showZonedZone t
--- a/src/Toml/Type/AnyValue.hs
+++ b/src/Toml/Type/AnyValue.hs
@@ -21,7 +21,7 @@ import Data.Text (Text)
 import Data.Type.Equality ((:~:) (..))
 import Data.Either (isRight)
 
-import Toml.Type.Value (DateTime, TValue, TypeMismatchError, Value (..), eqValueList, sameValue)
+import Toml.Type.Value (DateTime, TValue, TypeMismatchError, Value (..), sameValue)
 
 -- | Existential wrapper for 'Value'.
 data AnyValue = forall (t :: TValue) . AnyValue (Value t)
@@ -30,13 +30,7 @@ instance Show AnyValue where
     show (AnyValue v) = show v
 
 instance Eq AnyValue where
-    (AnyValue (Bool b1))    == (AnyValue (Bool b2))    = b1 == b2
-    (AnyValue (Integer i1)) == (AnyValue (Integer i2)) = i1 == i2
-    (AnyValue (Double f1))  == (AnyValue (Double f2))  = isRight $ sameValue (Double f1) (Double f2)
-    (AnyValue (Text s1))    == (AnyValue (Text s2))    = s1 == s2
-    (AnyValue (Date d1))    == (AnyValue (Date d2))    = d1 == d2
-    (AnyValue (Array a1))   == (AnyValue (Array a2))   = eqValueList a1 a2
-    _                       == _                       = False
+    (AnyValue val1)  == (AnyValue val2)  = isRight $ sameValue val1 val2
 
 ----------------------------------------------------------------------------
 -- Matching functions for values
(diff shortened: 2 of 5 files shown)


src/Toml/Printer.hs
```diff
@@ -88,12 +88,21 @@ prettyKeyValue i = Text.concat . map kvText . HashMap.toList
     showText :: Show a => a -> Text
     showText = Text.pack . show
 
+    showDouble :: Double -> Text
+    showDouble d | isInfinite d && isNegate d = "-inf"
```
ChShersh: Don't need to have `isNegate` function. You can just check `d < 0`. This should be enough.

test/Test/Toml/Gen.hs
```diff
@@ -207,6 +214,7 @@ noneArrayList =
     [ AnyValue . Bool    <$> genBool
     , AnyValue . Integer <$> genInteger
     , AnyValue . Double  <$> genDouble
+    , AnyValue . Double  <$> genDoubleText
```
ChShersh: This shouldn't be separate generator in this list. You should combine it with `genDouble`. And you should use `frequency` generator to adjust frequency of value appearance. Let's have:

5 for Inf
5 for -Inf
5 for NaN
50 for Ordinary Double

src/Toml/Type/Value.hs
```diff
@@ -116,7 +116,9 @@ instance (t ~ 'TText) => IsString (Value t) where
 instance Eq (Value t) where
     (Bool b1)    == (Bool b2)    = b1 == b2
     (Integer i1) == (Integer i2) = i1 == i2
-    (Double f1)  == (Double f2)  = f1 == f2
+    (Double f1)  == (Double f2)
+        | isNaN f1 && isNaN f2 = True
```
ChShersh: Okay, I see the problem why it doesn't work. The problem is in the fact that in `TOML` data type we actually store `AnyValue`. And `Eq` instance for `AnyValue` is written like this:

https://github.com/kowainik/tomland/blob/543ae70c960134da49852c757b4b96e810490cf5/src/Toml/Type/AnyValue.hs#L34

So actual check is not triggered. The problem is in the fact that `Eq` for `AnyValue` doesn't reuse `Eq` for `Value`. This should be fixed. Unfortunately, implementation like:

```haskell
AnyValue val1 == AnyValue val2 = val1 == val2
```

doesn't work because `val1` and `val2` may have different types.

You need to reuse `sameValue` function in the `Eq` instance for `AnyValue`:

https://github.com/kowainik/tomland/blob/543ae70c960134da49852c757b4b96e810490cf5/src/Toml/Type/Value.hs#L200

willbasky: Now tests has passed! Thank you!

src/Toml/Type/AnyValue.hs
```diff
@@ -31,7 +32,7 @@ instance Show AnyValue where
 instance Eq AnyValue where
     (AnyValue (Bool b1))    == (AnyValue (Bool b2))    = b1 == b2
     (AnyValue (Integer i1)) == (AnyValue (Integer i2)) = i1 == i2
-    (AnyValue (Double f1))  == (AnyValue (Double f2))  = f1 == f2
+    (AnyValue (Double f1))  == (AnyValue (Double f2))  = isRight $ sameValue (Double f1) (Double f2)
```
ChShersh: The solution here is not completely correct... What I would like to see is that if you have `Anyvalue val1` and `AnyValue val2` you need to call `sameValue` on `val1` and `val2`, pattern match on the result and use `Eq` instance for `Value`. `sameValue` function only checks that types of values are the same, not the values themselves.

src/Toml/Type/AnyValue.hs
```diff
@@ -30,13 +30,7 @@ instance Show AnyValue where
...
 
 instance Eq AnyValue where
-    (AnyValue (Bool b1))    == (AnyValue (Bool b2))    = b1 == b2
-    (AnyValue (Integer i1)) == (AnyValue (Integer i2)) = i1 == i2
-    (AnyValue (Double f1))  == (AnyValue (Double f2))  = isRight $ sameValue (Double f1) (Double f2)
-    (AnyValue (Text s1))    == (AnyValue (Text s2))    = s1 == s2
-    (AnyValue (Date d1))    == (AnyValue (Date d2))    = d1 == d2
-    (AnyValue (Array a1))   == (AnyValue (Array a2))   = eqValueList a1 a2
-    _                       == _                       = False
+    (AnyValue val1)  == (AnyValue val2)  = isRight $ sameValue val1 val2
```
ChShersh: It's not enough to call `isRight` on `sameValue`. You also need to do `val1 == val2` here.

src/Toml/Type/AnyValue.hs
```diff
@@ -30,7 +30,15 @@ instance Show AnyValue where
     show (AnyValue v) = show v
 
 instance Eq AnyValue where
-    (AnyValue val1)  == (AnyValue val2)  = isRight $ sameValue val1 val2
+    (AnyValue (Bool b1))    == (AnyValue (Bool b2))    = b1 == b2
+    (AnyValue (Integer i1)) == (AnyValue (Integer i2)) = i1 == i2
+    (AnyValue (Double f1))  == (AnyValue (Double f2))
+        | isNaN f1 && isNaN f2 && isRight (sameValue (Double f1) (Double f2)) = True
+        | otherwise = f1 == f2
```
ChShersh: Okay, this piece of code requires some advanced knowledge of Haskell. I can't find any blog post about this so I will try to describe here. In `tomland` we have the following data type:

```haskell
data AnyValue = forall (t :: TValue) . AnyValue (Value t)
```

This is so-called an existential wrapper around `Value`. But basically this data type stores single field of type `Value t`. What I would like to have is `Eq` instance for `AnyValue` in the following form:

```haskell
instance Eq AnyValue where
    AnyValue val1 == AnyValue val2 = val1 == val2
```

`AnyValue` just stores `Value` and we already have `Eq` instance for `Value`, so why not reuse it, right? Unfortunately, this exact code doesn't compile. You probably tried this solution and it didn't work. The reason for this behavior is that we have `Eq` instance for `Value t` and `==` operator has the following type:

```haskell
(==) :: Value t -> Value t -> Bool
```

Which means that you can't compare `Value` parametrized by different types. And since `Value` type is GADT it actually can be parametrized by different types.

```haskell
λ: let val1 = Bool True
λ: let val2 = Double 3.5
λ: :t val1
val1 :: Value 'TBool
λ: :t val2
val2 :: Value 'TDouble
λ: val2 == val2
True
λ: val1 == val2

<interactive>:5:9: error:
    • Couldn't match type ‘'TDouble’ with ‘'TBool’
      Expected type: Value 'TBool
        Actual type: Value 'TDouble
    • In the second argument of ‘(==)’, namely ‘val2’
      In the expression: val1 == val2
      In an equation for ‘it’: it = val1 == val2
```

But, there's magic `Refl` data type (also known as _propositional equality_) in Haskell with the following definition:

```haskell
data a :~: b where
  Refl :: a :~: a
```

It's also a GADT. This data type parametrized by two type variables. But the only way to create values of this type is only when you know that types are equal. And this is where `sameValue` function comes in. `sameValue` function checks whether two `Value` of possibly different types are the same:

```haskell
λ: sameValue val2 val2
Right Refl
λ: sameValue val1 val2
Left Expected type 'Bool' but actual type: 'Double'
```

This function doesn't check that the content of `Value` is the same. We have `Eq` instance to check content of `Value`.

So, returning back to `AnyValue`: here some GHC magic involves. You can't write `val1 == val2` because the types can be different. But if you pattern-match on `Refl`, compiler will know that types of `val1` and `val2` are actually the same! And we can call `==` on this.


So this doesn't work:

```haskell
instance Eq AnyValue where
    AnyValue val1 == AnyValue val2 = val1 == val2
```

But this works:

```haskell
instance Eq AnyValue where
    AnyValue val1 == AnyValue val2 = case sameValue val1 val2 of
        Right Refl -> val1 == val2
        Left ... you can figure out that to do here by your own ;) ...
```

willbasky: Thank you for explain!
Actually I tried this one
```
(AnyValue val1)  == (AnyValue val2)  = isRight (sameValue val1 val2) && (val1 == val2)
```
And got
```
Couldn't match type ‘t1’ with ‘t’
  ‘t1’ is a rigid type variable bound by
    a pattern with constructor:
      AnyValue :: forall (t :: TValue). Value t -> AnyValue,
    in an equation for ‘==’
```
Here `Right Refl -> val1 == val2` both values are checked and `(==`) works, in my attempt  `(==)`  was parallel to checking, and therefore the error returned. That is the explain, I guess.

ChShersh: Yeah, you need to perform pattern-matching manually to bring type equality into scope

PR #93, opened 2018-09-22
Support GHC 8.6.1

ChShersh:


willbasky:
Is it ok not to support older GHC versions?

vrom911:
@willbasky, yes, we follow the '3 latest major versions' policy in our projects.

PR #96, opened 2018-10-01
[#90] Migrate to megaparsec 7

### LINKED ISSUE #76: Use `binary` from megaparsec when tomland start to use megaparsec-7.0.0
This will make the integer parser code shorter.

The current integer parser code: https://github.com/kowainik/tomland/blob/00a2cac30f5af281bcc82f2c4d64579ad9e4d69a/src/Toml/Parser.hs#L159-L168

`binary` from megaparsec: https://github.com/mrkkrp/megaparsec/blob/f3e26a42231dfac748045325a7f45ba07a04f069/Text/Megaparsec/Char/Lexer.hs#L442-L452

### LINKED ISSUE #90: Migrate to megaparsec-7
Performance of `megaparsec` was greatly improved in version 7. So it's a good idea to migrate to newer version.

cocreature:
fixes #90, fixes #76

### DIFF
--- a/tomland.cabal
+++ b/tomland.cabal
@@ -43,7 +43,7 @@ library
 
   build-depends:       base >= 4.10 && < 5
                      , hashable ^>= 1.2
-                     , megaparsec ^>= 6.5.0
+                     , megaparsec ^>= 7.0.1
                      , mtl ^>= 2.2
                      , parser-combinators
                      , text ^>= 1.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The change log is available [on GitHub][2].
 0.5.0
 =====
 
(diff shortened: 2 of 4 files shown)


tomland.cabal
```diff
@@ -43,7 +43,7 @@ library
 
   build-depends:       base >= 4.10 && < 5
                      , hashable ^>= 1.2
-                     , megaparsec ^>= 6.5.0
+                     , megaparsec ^>= 7.0
```
ChShersh: Could you specify `^>= 7.0.1` version here? I see from the `megaparsec` changelog that `7.0.1` version contains bugfix, so I would like to constraint version even more to be sure that we have the bugfix.

cocreature:
I’ve added it to the changelog and tightened the constraint.

PR #97, opened 2018-10-01
[#17] Allow underscores in integers

### LINKED ISSUE #17: Add parsing of _ in numbers
TOML allows you to specify numbers like `1_000_000`. This is not supported currently. But probably should  be.

tomphp:
This PR aims to allow underscores in number as mentioned in #17. 

### Some facts about this implementation
- I've used existing commented out tests and assumed they contain the desired behaviour (the TOML spec doesn't cover some of the edge cases in the commented tests).
- This only works for integers.

### Some concerns about the implementation (any suggestions to improve work but appreciated):
- The removing of the underscores is done in a separate action from the parsing. I can use explicit parsers rather than the `match` function, but it's less concise.
- I'm not happy with the use of `undefined` in `fromRight` but the parser should ensure this never happens.

Resolves #17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ The change log is available [on GitHub][2].
 0.5.0
 =====
 
+* [#97](https://github.com/kowainik/tomland/pull/97)
+  Allow underscores in integers*.
 * [#96](https://github.com/kowainik/tomland/issues/96):
   Migrate to megaparsec 7.0
 * [#81](https://github.com/kowainik/tomland/issues/81):
--- a/src/Toml/Parser.hs
+++ b/src/Toml/Parser.hs
@@ -20,13 +20,14 @@ import Control.Applicative (Alternative (..))
 import Control.Applicative.Combinators (between, count, manyTill, optional, sepEndBy, skipMany)
 import Control.Monad (void)
 import Data.Char (chr, isControl)
+import Data.Either (fromRight)
 import Data.Fixed (Pico)
 import Data.Semigroup ((<>))
 import Data.Text (Text)
 import Data.Time (LocalTime (..), ZonedTime (..), fromGregorianValid, makeTimeOfDayValid,
                   minutesToTimeZone)
 import Data.Void (Void)
-import Text.Megaparsec (Parsec, anySingle, errorBundlePretty, satisfy, try)
+import Text.Megaparsec (Parsec, anySingle, errorBundlePretty, match, satisfy, try)
 import Text.Megaparsec.Char (alphaNumChar, char, digitChar, eol, hexDigitChar, space, space1,
                              string, tab)
 
@@ -36,6 +37,7 @@ import Toml.Type (AnyValue, DateTime (..), TOML (..), UValue (..), typeCheck)
 import qualified Control.Applicative.Combinators.NonEmpty as NC
 import qualified Data.HashMap.Lazy as HashMap
 import qualified Data.Text as Text
+import qualified Data.Text.Read as TR
 import qualified Text.Megaparsec as Mega (parse)
 import qualified Text.Megaparsec.Char.Lexer as L
 
@@ -159,10 +161,19 @@ tableNameP = lexeme $ between (char '[') (char ']') keyP
 
 -- Values
 
+decimalP :: Parser Integer
+decimalP = mkInteger <$> decimalStringP
+  where
+    decimalStringP   = fst <$> match (some digitChar >> many _digitsP)
+    _digitsP         = try (char '_') >> some digitChar
+    mkInteger        = textToInt . stripUnderscores
+    textToInt        = fst . fromRight (error "Underscore parser has a bug") . TR.decimal
+    stripUnderscores = Text.filter (/= '_')
+
 integerP :: Parser Integer
 integerP = lexeme $ binary <|> octal <|> hexadecimal <|> decimal
   where
-    decimal      = L.signed sc L.decimal
+    decimal      = L.signed sc decimalP
     binary       = try (char '0' >> char 'b') >> L.binary
     octal        = try (char '0' >> char 'o') >> L.octal
     hexadecimal  = try (char '0' >> char 'x') >> L.hexadecimal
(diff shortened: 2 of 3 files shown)


src/Toml/Parser.hs
```diff
@@ -159,10 +161,19 @@ tableNameP = lexeme $ between (char '[') (char ']') keyP
 
 -- Values
 
+decimalP :: Parser Integer
+decimalP = mkInteger <$> decimalStringP
+  where
+    decimalStringP   = fst <$> match (some digitChar >> optional (try $ many _digitsP))
```
ChShersh: The usage of `match` looks interesting 🤔It looks like and efficient enough way to get the part of the input. Well done!

src/Toml/Parser.hs
```diff
@@ -159,10 +161,19 @@ tableNameP = lexeme $ between (char '[') (char ']') keyP
...
+    _digitsP         = char '_' >> some digitChar
```
ChShersh: I'm not really sure that `try $ many smth` is an efficient parser. Looks like you can move `try` to `_digitsP`, like:

```haskell
try (char '_') *> some digitChar
```

Also, if you replace `some digitChar` with `many digitChar` you can remove `optional` from `decimalStringP`. I hope it works! And if it works, it will make the code cleaner.

src/Toml/Parser.hs
```diff
@@ -159,10 +161,19 @@ tableNameP = lexeme $ between (char '[') (char ']') keyP
...
+    mkInteger        = textToInt . stripUnderscores
+    textToInt        = fst . fromRight undefined . TR.decimal
+    stripUnderscores = Text.replace "_" ""
```
ChShersh: I think `Text.filter` is more efficient here than `replace`:

* https://hackage.haskell.org/package/text-1.2.3.1/docs/Data-Text.html#v:filter

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -161,15 +162,15 @@ spec_Parser = do
...
-          --  parseInt "13_" 13
-          --  intFailOn "_123_"
-          --  intFailOn "_13"
-          --  intFailOn "_"
+            it "can parse numbers with underscores between digits" $ do
+               parseInteger "1_000" 1000
+               parseInteger "5_349_221" 5349221
+               parseInteger "1_2_3_4_5" 12345
+               parseInteger "1_2_3_" 1
+               parseInteger "13_" 13
```
ChShersh: That's great that you found these tests and uncommented them! These tests should be enough.

src/Toml/Parser.hs
```diff
@@ -159,10 +161,19 @@ tableNameP = lexeme $ between (char '[') (char ']') keyP
 
 -- Values
 
+decimalP :: Parser Integer
```
ChShersh: I think it's a good idea for you to open an issue in the `megaparsec` library and ask to add some default parser for integers with underscores. You can link `megaparsec` maintainer to your implementation under this PR as an example of such parser 🙂 I think it's a good idea to have such parser in standard parser-combinators library. Especially in the context that there's support for `_` in numeric literals in Haskell since GHC 8.6.1 (enabled by `NumericUnderscores` language extension). And if such parser appears in `megaparsec`, we can replace our parser in `tomland` with more efficient and standard one. I think this will happen only about one month in `megaparec` library and we still can have this feature in `tomland`. 

I'm okay if the parser in `megaparsec` doesn't parse things like `13_`, it's still better to use standard function, especially then the specification of configuration format is not precise enough.

src/Toml/Parser.hs
tomphp: It needs to be this way to pass the test `1_2_3_ == 1`. If the `optional` is removed and the `try` is moved into `_digitsP` then you get `1_2_3_ == 123`. I don't really know what the correct behaviour should be here so happy to change it if the second one is valid?

ChShersh: Oops, that was unfortunate. Yeah, `1_2_3_ == 1` test is invalid, it should be `123` so feel free to change this particular test 👌

src/Toml/Parser.hs
tomphp: Great idea. I'll open a PR for megaparsec and come back and update tomland if it's accepted 👍

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ Change log
 tomland uses [PVP Versioning][1].
 The change log is available [on GitHub][2].
 
+0.5.1
+=====
+
+* [#97](https://github.com/kowainik/tomland/pull/97)
```
ChShersh: This should go in version 0.5.0 entry because it's not released on Hackage yet.

src/Toml/Parser.hs
```diff
@@ -159,10 +161,19 @@ tableNameP = lexeme $ between (char '[') (char ']') keyP
...
+decimalP = mkInteger <$> decimalStringP
+  where
+    decimalStringP   = fst <$> match (some digitChar >> many _digitsP)
+    _digitsP         = try (char '_' >> some digitChar)
+    mkInteger        = textToInt . stripUnderscores
+    textToInt        = fst . fromRight undefined . TR.decimal
```
ChShersh: `fromRight undefined` is not a pleasant thing... But usually instead of `undefined` it's always better to write `error` with some description, like:

```haskell
fromRight (error "Underscore parser has a bug")
```

src/Toml/Parser.hs
```diff
@@ -159,10 +161,19 @@ tableNameP = lexeme $ between (char '[') (char ']') keyP
 
 -- Values
 
+decimalP :: Parser Integer
+decimalP = mkInteger <$> decimalStringP
+  where
+    decimalStringP   = fst <$> match (some digitChar >> many _digitsP)
+    _digitsP         = try (char '_' >> some digitChar)
```
ChShersh: Shouldn't it be okay to wrap only `char '_'` in `try`? My personal feeling is that `try` can decrease performance a lot, so it's better to reduce scope of `try` parser combinator as much as possible.

tomphp: I tried with it just on the `char '_'` but then I get a parse error on strings that end with an `_` :(

ChShersh: @tomphp From the TOML specification:

> For large numbers, you may use underscores between digits to enhance readability

So it's okay if parser can't parsee number like `13_`. Underscores are allowed only between characters.

tomphp: Done

PR #100, opened 2018-10-02
Add PrintOptions for pretty printer.

### LINKED ISSUE #11: Add 'PrintOptions' for 'tomlToText' function

astynax:
Closes #11

### DIFF
--- a/src/Toml/Printer.hs
+++ b/src/Toml/Printer.hs
@@ -3,15 +3,17 @@
 {- | Contains functions for pretty printing @toml@ types. -}
 
 module Toml.Printer
-       ( prettyToml
+       ( PrintOptions(..)
+       , pretty
+       , prettyOptions
        , prettyTomlInd
        ) where
 
 import Data.HashMap.Strict (HashMap)
-import Data.Monoid ((<>))
+import Data.Monoid ((<>), mconcat)
 import Data.Text (Text)
 import Data.Time (formatTime, defaultTimeLocale, ZonedTime)
-import Data.List (splitAt)
+import Data.List (splitAt, sortOn)
 
 import Toml.PrefixTree (Key (..), Piece (..), PrefixMap, PrefixTree (..))
 import Toml.Type (AnyValue (..), DateTime (..), TOML (..), Value (..))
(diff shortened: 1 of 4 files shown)


src/Toml/Printer.hs
```diff
@@ -104,21 +124,25 @@ prettyKeyValue i = Text.concat . map kvText . HashMap.toList
             . formatTime defaultTimeLocale "%z"
 
 -- | Returns pretty formatted tables section of the 'TOML'.
-prettyTables :: Int -> Text -> PrefixMap TOML -> Text
-prettyTables i pref = Text.concat . map prettyTable . HashMap.elems
+prettyTables :: PrintOptions -> Int -> Text -> PrefixMap TOML -> Text
+prettyTables options i pref = Text.concat . map prettyTable . HashMap.elems
```
ChShersh: I think that it's a good idea to apply `order` here as well. I think if you specify `shouldSort`, you should sort table names as well.

src/Toml/Printer.hs
```diff
@@ -20,11 +22,21 @@ import qualified Data.HashMap.Strict as HashMap
...
+    , indent     :: Int   -- ^ indentation size
+    } deriving (Show)
+
+{- | Default printing options. -}
+defaultOptions :: PrintOptions
+defaultOptions = PrintOptions True 2
+
 -- Tab is equal to 2 spaces now.
 tab :: Int -> Text
 tab n = Text.cons '\n' (Text.replicate (2*n) " ")
```
ChShersh: I think you should also patch this function and replace `2 * n` with `n`. As I understand correctly, the `tab` function will be given numbers `0, 2, 4, 6, 8, ...` or `0, 4, 8, 12, ...` (depending on the indentation size) and previously it took numbers `0, 1, 2, 3, ...`. So instead of multiplying sequent natural numbers by indentation we can just add indent to the `nextI` (as you already did) but also remove multiplier from the `tab` function

ChShersh:
The CI error is really weird and I don't quite understand, why it happens... Requires further investigation.

Also, could you please update the CHANGELOG.md?

vrom911:
CI is fixed now! See the comment in the other `tomland` PR: https://github.com/kowainik/tomland/pull/101#issuecomment-426503781

src/Toml/Printer.hs
```diff
@@ -20,11 +22,28 @@ import qualified Data.HashMap.Strict as HashMap
...
 
-{- | Converts 'TOML' type into 'Text'.
+{- | Default printing options. -}
+defaultOptions :: PrintOptions
+defaultOptions = PrintOptions True 2
+
+-- Returns an indentation prefix
+tabWith :: PrintOptions -> Int -> Text
+tabWith options n =
+    Text.cons '\n' (Text.replicate (n * indent options) " ")
```
ChShersh: `n` argument is redundant here and there's no need to multiply by `n`. In the code you assign `nextI` variable (which is responsible for indentation) to the value `i + indent options` and you pass `options` and `i` to the `tab` function. So, for example, if default indentation is `4` then layers will have the following indents:

* Nested 0: `indent = 4`, `n = 0` => tab is 0 spaces
* Nested 1: `indent = 4`, `n = 4` => tab is 16 spaces
* Nested 2: `indent = 4`, `n = 8` => tab is 32 spaces
* and so on

astynax: Oh, I thought about it, but forgot to drop `+ indent options`...

astynax: Now `n` means only a level of indentation (as it was before)

Issue #82, opened 2018-08-30
Remove `maybeT` in favour of `dioptional`


ChShersh:
See here:

https://github.com/kowainik/summoner/blob/master/src/Summoner/Config.hs#L165-L169

willbasky:
What is `dioptional`?
Link points to 
```
lastT :: (Key -> BiToml a) -> Key -> BiToml (Last a)
lastT = Toml.wrapper . Toml.maybeT
```
the function uses `maybeT` self.

ChShersh:
@willbasky Sorry, link is wrong. It should be with commit. Look at `dimaybeT`. This function should convert `BiCodec a` to `BiCodec (Maybe a)`

willbasky:
> This function should convert BiCodec a to BiCodec (Maybe a)

Do you mean `TomlCodec a -> TomlCodec (Maybe a)`?
as it is in `maybeT :: forall a . (Key -> TomlCodec a) -> Key -> TomlCodec (Maybe a)`

willbasky:
And **why** change `maybeT` with `dimaybeT`?

ChShersh:
> Do you mean TomlCodec a -> TomlCodec (Maybe a)?

No, this function should be general and work for `BiCodec`

https://github.com/kowainik/tomland/blob/d1720a08534482527b63dff6474d4ac84bc73dcb/src/Toml/Bi/Monad.hs#L47

> And why change maybeT with dimaybeT?

`maybeT` is wrong an it often happens that `maybeT` doesn't work as expected. So `maybeT` should be removed. So I want to have something like `optional` for `Alternative` for our our `Codec`. Unfortunately, `optional` for `Codec` works in the following way:

```haskell
optional :: Codec r w a b -> Codec r w a (Maybe b)`
```

while I want:

```haskell
dioptional :: Codec r w a b -> Codec r w (Maybe a) (Maybe b)
```

If you can write `dioptional` for `Codec` instead of `BiCodec`, it would be even better :+1:

ChShersh:
@willbasky Yeah, types were renamed in some recent PR. And the new version is not on Hackage yet.

willbasky:
Does `dimaybeT` something like workaround?

ChShersh:
Yes, it is. After `dioptional` is implemented in `tomland`, it will replace `dimaybeT` in `summoner`

PR #101, opened 2018-10-02
[#82] Remove maybeT, add diOptional

willbasky:
Resolves: [#82](https://github.com/kowainik/tomland/issues/82)


src/Toml/Bi/Combinators.hs
```diff
@@ -169,18 +169,13 @@ arrayOf bimap key = Codec input output
...
-maybeT :: forall a . (Key -> TomlCodec a) -> Key -> TomlCodec (Maybe a)
-maybeT converter key = let codec = converter key in Codec
-    { codecRead  = (Just <$> codecRead codec) `catchError` handleNotFound
+diOptional :: (Functor r, Functor w, Applicative w) => Codec r w c a -> Codec r w (Maybe c) (Maybe a)
+diOptional Codec {..} = Codec
+    { codecRead = Just <$> codecRead
     , codecWrite = \case
         Nothing -> pure Nothing
-        Just v  -> codecWrite codec v >> pure (Just v)
+        Just c -> Just <$> codecWrite c
```
ChShersh: Nice! You've just implemented `Traversable` instance for `Maybe`. If you look at this instance, you probably might notice how to implement `codecWrite` in more concise way:

* https://hackage.haskell.org/package/base-4.12.0.0/docs/src/Data.Traversable.html#line-232

src/Toml/Bi/Combinators.hs
```diff
@@ -169,18 +169,13 @@ arrayOf bimap key = Codec input output
         a <$ modify (\(TOML vals tables) -> TOML (HashMap.insert key anyVal vals) tables)
 
 -- | Bidirectional converter for @Maybe smth@ values.
-maybeT :: forall a . (Key -> TomlCodec a) -> Key -> TomlCodec (Maybe a)
-maybeT converter key = let codec = converter key in Codec
-    { codecRead  = (Just <$> codecRead codec) `catchError` handleNotFound
+diOptional :: (Functor r, Functor w, Applicative w) => Codec r w c a -> Codec r w (Maybe c) (Maybe a)
+diOptional Codec {..} = Codec
```
ChShersh: There should be no space between `Codec` and `{..}`

vrom911:
Travis fails with an error for `tasty-quickcheck` for `Cabal`. Seems like this package has breaking change in the newest version and one of our dependencies (which depends on the `tasty-quickcheck` because we don't 🙂 ) didn't have the upper bounds for that. I will investigate this..

vrom911:
It seems it fixed itself. Version `2.12.6.1` of the `QuickCheck` just a few hours after the `2.12.6`, so guess the issue was in there, cause it works now 🙂

src/Toml/Bi/Combinators.hs
willbasky: If so, maybe someday I will implement somewhat function for `relude` :)

willbasky:
While I slept the problem was solved :)

PR #102, opened 2018-10-04
Updated integerP to fail on leading zeros

### LINKED ISSUE #64: Do not allow leading zeros in the decimal representation of integer values
As per [TOML specifications](https://github.com/toml-lang/toml/blob/master/README.md), leading zeros are not allowed in the decimal representation of integer values.

Might be easier to implement after #17 is implemented.

MitchStevens:
Modified Parser.hs to fail on leading zeroes. Also added a small test to verify that the parser is working as expected. Please advise if I have done something wrong here, this is my first pull request.

Resolves #64

### DIFF
--- a/src/Toml/Parser.hs
+++ b/src/Toml/Parser.hs
@@ -17,16 +17,18 @@ module Toml.Parser
 
 -- I hate default Prelude... Do I really need to import all this stuff manually?..
 import Control.Applicative (Alternative (..))
-import Control.Applicative.Combinators (between, count, manyTill, optional, sepEndBy, skipMany)
+import Control.Applicative.Combinators (between, count, manyTill, optional, sepEndBy, skipMany, some, sepBy1)
 import Control.Monad (void)
+--import Control.Monad.Fail (fail)
 import Data.Char (chr, isControl)
+import Data.Either (fromRight)
 import Data.Fixed (Pico)
 import Data.Semigroup ((<>))
 import Data.Text (Text)
 import Data.Time (LocalTime (..), ZonedTime (..), fromGregorianValid, makeTimeOfDayValid,
                   minutesToTimeZone)
 import Data.Void (Void)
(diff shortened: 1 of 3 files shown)


src/Toml/Parser.hs
```diff
@@ -162,7 +162,11 @@ tableNameP = lexeme $ between (char '[') (char ']') keyP
 integerP :: Parser Integer
 integerP = lexeme $ binary <|> octal <|> hexadecimal <|> decimal
   where
-    decimal      = L.signed sc L.decimal
+    decimal      = L.signed sc (noLeadingZero <|> zero)
```
ChShersh: Looks like you didn't update on latest master, there's different `decimalP` parser is used:

* https://github.com/kowainik/tomland/blob/3187fed8dc353b68553f229b5dfb76fdc4e5e9f0/src/Toml/Parser.hs#L176

I think you should update `decimalP` parser directly to forbid leading zeroes. Otherwise the implementation looks good 👍

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -156,8 +157,12 @@ spec_Parser = do
             it "can parse sign-prefixed zero as an unprefixed zero" $ do
                 parseInteger "+0" 0
                 parseInteger "-0" 0
-            it
-                    "can parse both the minimum and maximum numbers in the 64 bit range"
+            it "will fail for numbers with a leading zero" $ do
+                integerFailOn "00"
+                integerFailOn "01234"
+                integerFailOn "+0333"
+                integerFailOn "-02"
```
ChShersh: The tests looks good!

ChShersh:
@MitchStevens Thank you very much for the fix! I really appreciate the fix and the documentation. Also, I see that your `master` branch in your fork is more updated and fresher, so I think in that case it's okay to close this PR and open new one directly from your `master` branch.

PR #103, opened 2018-10-04
Add golden tests for pretty printer

astynax:
- Add golden tests for pretty printer
- Actualize haddock for `pretty`
- Export forgotten (?) `(<|)` from `Toml.PrefixTree`


test/Test/Toml/Printer/Golden.hs
```diff
@@ -0,0 +1,49 @@
...
+          , ("doo", mempty)
+          , ("baz", mempty)
+          ]
+    }
+
+test_prettyGolden :: TestTree
+test_prettyGolden =
+    testGroup "Toml.Printer"
+        [ test pretty "test/golden/pretty_default.golden"
+        , test (prettyOptions $ PrintOptions True 0)
```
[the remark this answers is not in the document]
astynax: Good idea!

src/Toml/PrefixTree.hs
```diff
@@ -2,6 +2,7 @@
 
 module Toml.PrefixTree
        ( PrefixTree (..)
+       , (<|)
```
ChShersh: Thank you for reexporting this function! It was forgotten to reexport indeed.

test/Test/Toml/Printer/Golden.hs
```diff
@@ -0,0 +1,49 @@
...
+import Toml (AnyValue (..), TOML (..), Value (..))
+import Toml.PrefixTree ((<|))
+import Toml.Printer (PrintOptions (..), pretty, prettyOptions)
+
+import qualified Data.HashMap.Strict as HashMap
+import qualified Toml.PrefixTree as PrefixTree
+
+example :: TOML
+example = mempty
+    { tomlPairs  = HashMap.fromList
```
ChShersh: Actually, we have eDSL implemented by @vrom911 for creating TOML data types. It was designed specially for such cases. See the following module:

* https://github.com/kowainik/tomland/blob/master/src/Toml/Edsl.hs

And here is one usage example:

* https://github.com/kowainik/tomland/blob/3187fed8dc353b68553f229b5dfb76fdc4e5e9f0/examples/Playground.hs#L81-L95

So you can write this test in more convenient way. And we also will test our eDSL as well 🙂

astynax: :+1:

Issue #70, opened 2018-06-28
Add easier way to write custom `BiMap` for values


ChShersh:
If I have custom data type with couple functions like this one:

```haskell
data GhcVer = ...

showGhcVer  :: GhcVer -> Text
parseGhcVer :: Text -> Maybe GhcVer
```

Then writing custom `BiMap` for this data type is a little bit tedious:

```haskell
    _GhcVer :: BiMap AnyValue GhcVer
    _GhcVer = BiMap
        { forward = \(AnyValue t) -> Toml.matchText t >>= parseGhcVer
        , backward = Just . AnyValue . Toml.Text . showGhcVer
        }
```

I want easier way to write such prisms.

jiegillet:
If you mean something like
```
toBiMap :: (a -> Text) -> (Text -> Maybe a) -> BiMap AnyValue a
toBiMap showA parseA = BiMap
        { forward = \(AnyValue t) -> Toml.matchText t >>= parseA
        , backward = Just . AnyValue . Toml.Text . showA
        }
```
I'd be happy to add it!

ChShersh:
@jiegillet This is good start! But I would like to have more generic solution, that can work not only with `Text` constructor of `Value` data type. Having function with type `(a -> Text) -> (Text -> Maybe a) -> BiMap AnyValue a` is already useful, but I guess it should have more specific name rather than `toBiMap`. Probably `textValBiMap` or something like that.

ChShersh:
Done in #104

PR #104, opened 2018-10-05
[Issue #70] Added function to create BiMap from concrete types

jiegillet:
This is to address #70.
Maybe the class definition should move to `AnyValue.hs`.

This is my first time sending a PR to a project like this one, I hope I can contribute.

### DIFF
--- a/src/Toml/BiMap.hs
+++ b/src/Toml/BiMap.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE DataKinds      #-}
 {-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE Rank2Types     #-}
-{-# LANGUAGE TypeFamilies     #-}
 
 {- | Implementation of partial bidirectional mapping as a data type.
 -}
@@ -16,7 +15,7 @@ module Toml.BiMap
          -- * Helpers for BiMap and AnyValue
        , matchValueForward
        , mkAnyValueBiMap
-       -- , mkAnyTypeBiMap
+       , mkAnyTypeBiMap
 
          -- * Some predefined bi mappings
        , _Array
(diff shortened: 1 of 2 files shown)


ChShersh:
@jiegillet Thanks for your contribution! You solution looks interesting. I would like to avoid typeclasses, so need to think how to do this properly :thinking:

jiegillet:
Why avoid typeclasses?
Plan B is to define textValBiMap, integerValBimap, ...

ChShersh:
> Why avoid typeclasses?

`tomland` intentionally took decision to build library around functions and treat everything as values because there're might be multiple ways to represent things. For example, if you have `Bool` in Haskell, this can be `Bool` from `Value` or this might be `"yes"\"no"` from `Text`. So there're could be a lot of possible ways to get `Bool` from TOML but by having typeclasses you allow to use only one because there could be only single instance of typeclass for the data type. You want typeclasses usually when in reality there could be only single meaningful instance or if just select the one that covers 95% use cases. But this is not the case in `tomland`.

> Plan B is to define textValBiMap, integerValBimap, ...

I think, currently I want the following two combinators that cover most use cases:

```haskell
_TextBy :: (Text -> Maybe a) -> (a -> Text) -> BiMap AnyValue a
_Show   :: (Show a, Read a) => BiMap AnyValue a
```

I would like to avoid create tens of functions unless we sure that there're no better way to do this :slightly_smiling_face:

src/Toml/BiMap.hs
```diff
@@ -126,6 +134,12 @@ _TextToString = iso T.unpack T.pack
 _String :: BiMap AnyValue String
 _String = _Text >>> _TextToString
 
+_StringToShow :: (Show a, Read a) => BiMap String a
+_StringToShow = iso read show
```
ChShersh: This is not actually an `iso` because you might not be able parse value of type `a` from `String`. So here you should use `prism` and `readMaybe`

jiegillet:
Alright, thank you for explaining ^^
I've been using Haskell a lot on my own but never for a package or for other people, so I appreciate the conversation.

Issue #95, opened 2018-10-01
Swap order of type arguments for `BiMap` in `match` and `arrayOf` functions


ChShersh:
Currently we have the following combinator:

```haskell
data BiMap a b = BiMap
    { forward  :: a -> Maybe b
    , backward :: b -> Maybe a
    }

prism :: (object -> Maybe field) -> (field -> object) -> BiMap object field
prism preview review = BiMap preview (Just . review)
```

But usually it's used with `invert` function to get required type:

```haskell
_Url :: BiMap Text Source
_Url = Toml.invert $ Toml.prism matchUrl Url
```

We need to add combinator like `invert . prism` but I don't know proper name for it 😞 The name should be short because it's going to be used often.

ChShersh:
@jiegillet The reason for this is because prisms in `tomland` has two uses cases:

1. For array values:
https://github.com/kowainik/tomland/blob/044de639b8c34285f24e244378c9116b0a86b7e6/src/Toml/Bi/Combinators.hs#L150
2. For parsing sum types with the help of `match` function
https://github.com/kowainik/tomland/blob/044de639b8c34285f24e244378c9116b0a86b7e6/src/Toml/Bi/Combinators.hs#L58

Though, as I can see now, those functions can take `BiMap a AnyValue` instead of `BiMap AnyValue a` and there won't be any need in having inverted prism. Even if this is a quite serious breaking change, I think this is the way to go. I'd rather prefer consistency with the rest of the Haskell ecosystem and not have extra function.

jiegillet:
I'll try to do the change. It might take a while, may the compiler be with me :)

ChShersh:
Closed in #106

PR #106, opened 2018-10-07
[Issue #95 ]Reversed BiMap Anything a into BiMap a Anything

jiegillet:
A couple of thoughts:

- I'm not super happy about my modification of `mkAnyValueBiMap`, I couldn't use `prism` since the direction didn't work out, so I ended up using `BiMap`. It works, but it's a bit blunt.
- This PR renders #104 obsolete, probably
- Maybe we need to add comments on `_Show`. Same as the others?
- `stack test` runs fine

Happy to discuss more.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The change log is available [on GitHub][2].
 
 0.5.0
 =====
+
+* [#106](https://github.com/kowainik/tomland/pull/106)
+  Swap fields in BiMaps for consistency with `lens` package. 
 * [#104](https://github.com/kowainik/tomland/pull/104)
   Add `_TextBy` and `_Show` combinators
 * [#100](https://github.com/kowainik/tomland/pull/100)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The change log is available [on GitHub][2].
 0.5.0
 =====
 
-* [#95](https://github.com/kowainik/tomland/issues/95
+* [#95](https://github.com/kowainik/tomland/issues/95)
   Swap fields in BiMaps for consistency with `lens` package.
 * [#70](https://github.com/kowainik/tomland/issues/70)
   Add `_TextBy` and `_Show` combinators
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,15 +7,15 @@ The change log is available [on GitHub][2].
 0.5.0
 =====
 
-* [#106](https://github.com/kowainik/tomland/pull/106)
-  Swap fields in BiMaps for consistency with `lens` package. 
-* [#104](https://github.com/kowainik/tomland/pull/104)
+* [#95](https://github.com/kowainik/tomland/issues/95
+  Swap fields in BiMaps for consistency with `lens` package.
+* [#70](https://github.com/kowainik/tomland/issues/70)
   Add `_TextBy` and `_Show` combinators
-* [#100](https://github.com/kowainik/tomland/pull/100)
+* [#11](https://github.com/kowainik/tomland/issues/11)
   Add `PrintOptions` (sorting, indentation) for pretty printer.
-* [#97](https://github.com/kowainik/tomland/pull/97)
+* [#17](https://github.com/kowainik/tomland/issues/17)
   Allow underscores in integers*.
-* [#96](https://github.com/kowainik/tomland/issues/96):
+* [#90](https://github.com/kowainik/tomland/issues/90):
   Migrate to megaparsec 7.0
 * [#81](https://github.com/kowainik/tomland/issues/81):
   **Important:** Rename data types.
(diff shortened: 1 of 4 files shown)


ChShersh:
@jiegillet Wow, you continue to bring very good improvements to the package! Let me take a look at the original issue and the original example one more time very carefully to make sure that we really need to swap order of arguments.

jiegillet:
Please take your time.
The best motivation I could see for switching the arguments was to do this `Toml._Left >>> Toml._Integer`. Without using inverse it would not have been possible.

src/Toml/BiMap.hs
```diff
@@ -83,11 +81,11 @@ prism preview review = BiMap preview (Just . review)
 -- General purpose bimaps
 ----------------------------------------------------------------------------
 
-_Left :: BiMap l (Either l r)
-_Left = invert $ prism (either Just (const Nothing)) Left
+_Left :: BiMap (Either l r) l
```
ChShersh: While you're on this, let's also write `_Just :: BiMap (Maybe a) a`

src/Toml/BiMap.hs
```diff
@@ -96,41 +94,52 @@ _Right = invert $ prism (either (const Nothing) Just) Right
...
 mkAnyValueBiMap :: (forall t . Value t -> Maybe a)
                 -> (a -> Value tag)
-                -> BiMap AnyValue a
+                -> BiMap a AnyValue
 mkAnyValueBiMap matchValue toValue =
-    prism (\(AnyValue value) -> matchValue value) (AnyValue . toValue)
+    BiMap (Just . AnyValue . toValue) (\(AnyValue value) -> matchValue value)
+
+-- | Creates prism for 'Text' to 'AnyValue' with custom functions
+_TextBy :: (Text -> Maybe a) -> (a -> Text) -> BiMap a AnyValue
```
ChShersh: Wasn't that function implemented in different PR? I don't see in the deleted diff...

src/Toml/BiMap.hs
```diff
@@ -96,41 +94,52 @@ _Right = invert $ prism (either (const Nothing) Just) Right
...
 
 -- | 'Text' bimap for 'AnyValue'. Usually used with 'arrayOf' combinator.
-_Text :: BiMap AnyValue Text
+_Text :: BiMap Text AnyValue
 _Text = mkAnyValueBiMap matchText Text
 
-_TextToString :: BiMap Text String
-_TextToString = iso T.unpack T.pack
+_StringToText :: BiMap String Text
+_StringToText = iso T.pack T.unpack
```
ChShersh: I have better idea regarding naming convention: let's remove `To` part from such prism. So it will be just: `_StringText`. It doesn't make much sense to me now to have `To` part because conversion is bidirectional.

src/Toml/BiMap.hs
jiegillet: Yes, it's a mistake on my part. It was added in #104 which was merged after I submitted this PR.  But I messed up my branches and ended up including it here. I'll make sure it merges properly.

CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ The change log is available [on GitHub][2].
 
 0.5.0
 =====
+
+* [#106](https://github.com/kowainik/tomland/pull/106)
```
ChShersh: Could, you, please, change the link to the pull request here to the link for corresponding issue? The issue number is more important. I somehow missed it once, and now everyone specifies link to the PR...

src/Toml/BiMap.hs
```diff
@@ -100,52 +104,52 @@ _Right = invert $ prism (either (const Nothing) Just) Right
...
+    BiMap (Just . AnyValue . toValue) (\(AnyValue value) -> matchValue value)
+
+-- | Allows to match against given 'Value' using provided prism for 'AnyValue'.
+matchValueBackward :: BiMap a AnyValue -> Value t -> Maybe a
+matchValueBackward = liftMatch . backward
 
--- | Creates prism for 'Text' to 'AnyValue' bimap with custom functions
-_TextBy :: (Text -> Maybe a) -> (a -> Text) -> BiMap AnyValue a
+-- | Creates prism for 'Text' to 'AnyValue' with custom functions
+_TextBy :: (Text -> Maybe a) -> (a -> Text) -> BiMap a AnyValue
```
ChShersh: Let's also swap order of arguments for `_TextBy` to be more consistent with the `prism` combinator

CHANGELOG.md
```diff
@@ -7,15 +7,15 @@ The change log is available [on GitHub][2].
 0.5.0
 =====
 
-* [#106](https://github.com/kowainik/tomland/pull/106)
-  Swap fields in BiMaps for consistency with `lens` package. 
-* [#104](https://github.com/kowainik/tomland/pull/104)
+* [#95](https://github.com/kowainik/tomland/issues/95
```
ChShersh: Probably a typo here: missing `)`

Issue #59, opened 2018-06-13
Add helpful bi combinators


ChShersh:
Add utility converters for the following types:

- ~[ ] `Set`~
- ~[ ] `HashSet`~
- [x] `String` (we have only for `Text` but sometimes you want `String`)
- [x] `ByteString`
- [x] `Natural`
- [x] `Float`
- [x] `Word`
- [x] `Read/Show` via `show/readMaybe`

jiegillet:
I'm happy to tackle this too after #106 is settled maybe. 
I'm assuming:
- Set, HashSet <=> TArray
- ByteString <=> TText
- Natural, Word <=> TInteger
- Float <=> TDouble

ChShersh:
@jiegillet You're very close to the truth!

> Natural, Word <=> TInteger
> Float <=> TDouble

The above combinators are fine :+1: 

> ByteString <=> TText

The types are true. But this should be done carefully since not every `ByteString` can be converted to `Text`.

> Set, HashSet <=> TArray

Probably, should wait for this one. Currently we don't even have combinator for the list. After we have list, we can implement `Set` and `HashSet` trivially.

jiegillet:
Got it. I'll be careful with the ByteString and I'll think about lists for a bit.

jiegillet:
I've thought for a bit, and (sorry) my typeclass idea showed up again.
Here is what I have in mind 

```haskell
class Valuable a where
  toValue :: a -> Maybe AnyValue
  matchValue :: AnyValue -> Maybe a

  _Valuable :: BiMap AnyValue a
  _Valuable = BiMap matchValue toValue

instance Valuable Bool where
  toValue = Just . AnyValue . Bool
  matchValue (AnyValue t) = matchBool t

instance (Valuable a) => Valuable [a] where
  toValue = mapM toValue >=> fmap AnyValue . toMArray
  matchValue (AnyValue v) = matchArray matchValue v

_Bool :: BiMap AnyValue Bool
_Bool = _Valuable

_List :: Valuable a => BiMap AnyValue [a]
_List = _Valuable
```
It compiles and I think it's very easy to read and write. What do you think?

BTW, I just want to say that `toMArray` has an awesome implementation, I was struggling to implement something like that before I realized it was already there.

ChShersh:
@jiegillet In that case `Valuable` doesn't really decreases amount of work. You still write implementation for every conversion function, just put it into typeclass.

[...]

Issue #99, opened 2018-10-02
Split `Toml.Parser` file into smaller files


ChShersh:
This file is pretty big and takes a lot of time to compile, unfortunately. So this is the moment where it should be split into smaller files. I imagine the following hierarchy:

1. `Toml.Parser.Core`: basic general purpose combinators under `Library for parsing ` section
2. `Toml.Parser.Value`: `valueP` parser and [all parsers required for `valueP`](https://github.com/kowainik/tomland/blob/9d114b5d91bfd6f8650a9296c22970525cdf30a5/src/Toml/Parser.hs#L269-L275)
3. `Toml.Parser.TOML`: all parsers for TOML data type
4. `Toml.Parser`: `ParseException` data type and `parse` function.

MorrisLaw:
Can I take this?

vrom911:
@MorrisLaw go ahead!

ChShersh:
@gahag Sure, you can work on this!

MorrisLaw:
Yes, I haven't had time to really dive in. Go for it @gahag

Issue #13, opened 2018-04-04
Support array of tables


ChShersh:
This is important issue since array of tables is probably the only reasonable way to specify data structures like `List` and `Map` in TOML. Though, it's hard to support them...

GooseDB:
If you need https://github.com/toml-lang/toml#user-content-array and https://github.com/toml-lang/toml#user-content-table I can take this issue.

GooseDB:
If you need [arrays](https://github.com/toml-lang/toml#user-content-array) and [tables](https://github.com/toml-lang/toml#user-content-table) I can take this issue.

ChShersh:
@GooseDB Arrays and tables are already supported by the `tomland` library :slightly_smiling_face: This issue is about [Array of tables](https://github.com/toml-lang/toml#user-content-array-of-tables).

GooseDB:
Oh, I see. I'll try.

ChShersh:
@GooseDB Note, that you should uncomment and use this field of `TOML` data type:

* https://github.com/kowainik/tomland/blob/044de639b8c34285f24e244378c9116b0a86b7e6/src/Toml/Type/TOML.hs#L24

Also, be aware that the refactoring of the `Toml.Parser` module by @MorrisLaw is coming and it's probably a good idea to wait some time to avoid huge conflicts :slightly_smiling_face:

GooseDB:
@ChShersh So, when will the refactoring be completed?

ChShersh:
@GooseDB I think, only @MorrisLaw can answer...

ChShersh:
@GooseDB On the other hand, I think, you can start to work in this. I will help with resolving conflicts if anything happens 👍

gahag:
I've sent a pull request with the refactoring.

jiegillet:
I might start working on this one if it's OK.

jiegillet:
I have a version that seems to be working (I need to run more tests though), but it doesn't include the new parsers in #121 so I'll wait for that one to be merged to proceed.

PR #107, opened 2018-10-08
[Issue #59] Adding a bunch of BiMaps

jiegillet:
Resovles #59 

Pretty straightforward.
- I'm not 100% sure about was where to add the `bytestring` dependency in the `.cabal` files. Please double check.
- I'm not including the List, Set and HashSet discussed in #59 yet.
- This will have conflicts with #106, but I'll fix whatever needed when/if needed
- I haven't commented the functions. Should I go with the standard `'XXX' bimap for 'AnyValue'. Usually used with 'arrayOf' combinator.` everywhere?


jiegillet:
I've modified the PR to adjust to #95.
Should I add `_List` `_Set` and `_HashSet` as discussed in #59? They are probably what makes most sense in the current context: give me the conversion bimap for en element and I'll give you a list.

src/Toml/BiMap.hs
```diff
@@ -145,6 +157,31 @@ _ShowString = BiMap (Just . show) readMaybe
...
+          | otherwise = Just (toInteger n)
+
+_Natural :: BiMap Natural AnyValue
+_Natural = _UnsignedInteger >>> _Integer
+
+_Word :: BiMap Word AnyValue
+_Word = _UnsignedInteger >>> _Integer
+
+_RealReal ::  (Real a, Real b, Fractional a, Fractional b) => BiMap a b
+_RealReal = iso realToFrac realToFrac
```
ChShersh: I feel like `_RealReal` is not going to be used outside `_Float`...

ChShersh:
> Should I add _List _Set and _HashSet as discussed in #59?

This is a hard question. As it was discussed, it doesn't really makes sense to use `Read/Show` for custom data types like this one:

```haskell
data User = User
    { userFirstName :: Text
    , userLastName :: Text
    , userFollowers :: Natural
    }
```

But we still can have sets via `Array` for primitive types, but I propose to somehow reflect this via name. Like `_ArraySet` and `_ArrayHashSet`.

jiegillet:
These type of custom values are much more like a `TOML` table than an `Anything`.

ChShersh:
> These type of custom values are much more like a TOML table than an Anything.

Yes, and currently `tomland` doesn't allow to parse values of type `[User]`. But TOML format supports this with the Array of Tables (especially inline tables).

But Array of Tables is not yet implemented in the `tomland`:

* https://github.com/kowainik/tomland/issues/13

jiegillet:
what do you think of this function:
```haskell
_Pair :: BiMap a AnyValue -> BiMap b AnyValue -> BiMap (a, b) AnyValue
_Pair (BiMap av va) (BiMap bv vb) = BiMap for back
  where
  for (a, b) = case (av a, bv b) of
    (Just (AnyValue v1), Just (AnyValue v2)) ->
      Just $ AnyValue $ Array [Array [v1], Array [v2]]
    _ -> Nothing
  back (AnyValue v) = case v of
    Array [Array [v1], Array [v2]] ->
      (,) <$> va (AnyValue v1) <*> vb (AnyValue v2)
    _ -> Nothing
```
It would be a step forward maybe. If we could generalize it to more than 2 values it could be neat. 
Note: the pattern matching for `Array [...]` in `back` requires the extension `GADTs`.

src/Toml/BiMap.hs
```diff
@@ -145,6 +157,31 @@ _ShowString = BiMap (Just . show) readMaybe
...
+
+_Float :: BiMap Float AnyValue
+_Float = _RealReal >>> _Double
+
+_ByteStringText :: BiMap ByteString Text
+_ByteStringText = prism encodeUtf8 maybeText
+  where maybeText = either (const Nothing) Just . decodeUtf8'
+
+_ByteString:: BiMap ByteString AnyValue
+_ByteString = _ByteStringText >>> _Text
```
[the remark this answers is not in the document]
jiegillet: Done!

src/Toml/BiMap.hs
jiegillet: That's very likely. I removed it from the list and only use it in `_Float`.

src/Toml/BiMap.hs
```diff
@@ -157,31 +161,42 @@ _ShowString = BiMap (Just . show) readMaybe
 _Show :: (Show a, Read a) => BiMap a AnyValue
 _Show = _ShowString >>> _String
 
-_UnsignedInteger :: (Integral n) => BiMap n Integer
-_UnsignedInteger = prism fromIntegral maybeInteger
+_NaturalInteger :: BiMap Natural Integer
+_NaturalInteger = BiMap (Just . toInteger) maybeInteger
   where maybeInteger n
```
ChShersh: I would appreciate having explicit type signature for `maybeInteger` function to make it clear in what direction numbers are moving. It's quite easy to swap arguments of `_NaturalInteger` bimap and not notice it... So something like this really improves code readability:

```haskell
  where
    maybeInteger :: Integer -> Maybe Natural
    maybeInteger n
         | n < 0 ....
```

ChShersh:
@jiegillet I was thinking on using array of arrays to encode lists of custom data types. But this becomes too complicated. And it loses important property of TOML: human-readability.

It's actually quite easy to implement list combinator but it requires support for array of tables (#13). Currently you're thinking on the `BiMap` level and we already have `_Array`:

```haskell
_Array :: BiMap a AnyValue -> BiMap [a] AnyValue
```

`BiMap` works on the level of `AnyValue` and this level is for primitive types like `Integer`, `Double`, `Text`, `DateTime`, `Bool`. To parse custom user data types, we need to use `TomlCodec` from `Toml.Bi.Combinators` library. So if we have user data type like this one:

```haskell
data User = User
    { userFirstName :: Text
    , userLastName :: Text
    , userFollowers :: Natural
    }
```

we currently can write bidirectional TOML converter for this data type in the following way:

```haskell
userCodec :: TomlCodec User
userCodec = User
    <$> Toml.text "firstName" .= userFirstName
    <*> Toml.text "lastName" .= userLastName
    <*> Toml.natural "followers" .= userFollowers
```

And then this data type can be parsed from the following TOML:

```toml
firstName = "Dmitrii"
lastName  = "Kovanikov"
followers = 500
```

But this TOML file represents only single `User` data type. Though, it's possible to represent list of users in the following way in TOML using _Array of tables_:

```toml
[[user]]
firstName = "Dmitrii"
lastName  = "Kovanikov"
followers = 500

[[user]]
firstName = "Foo"
lastName  = "Bar"
followers = 256
```

This is a list in TOML that contains two users. It's can be represented even better with _inline tables_:

```toml
users =
    [ { firstName = "Dmitrii", lastName = "Kovanikov", followers = 500 }
    , { firstName = "Foo", lastName = "Bar", followers = 256 }
    ]
```

But `tomland` doesn't support currently neither array of tables nor inline tables. Though, there're open issues :slightly_smiling_face:

jiegillet:
You're right that working at the `BiMap` levels gets too complicated.

I still haven't added `_Set` and `_HashSet` but they're ready to go if you want them.

I was thinking of working on parsing inlines tables and/or array of tables. I know someone already volunteered but they haven't been active. I'll see.

ChShersh:
> I still haven't added _Set and _HashSet but they're ready to go if you want them.

I will create separate issue for that 👍
I think this can be done separately.

> I was thinking of working on parsing inlines tables and/or array of tables. I know someone already volunteered but they haven't been active. I'll see.

The issue was blocked because of #99. But now #116 is merged, so you can try to ask about array of tables again 🙂

PR #108, opened 2018-10-10
[#64] Re-implemented `decimalP`

MitchStevens:
Re-implemented `decimalP`, disallowing leading zeros and double underscores to be matched when parsing integers. This is a copy of my previous pull request but from the master branch.

### DIFF
--- a/src/Toml/Parser.hs
+++ b/src/Toml/Parser.hs
@@ -28,7 +28,7 @@ import Data.Text (Text)
 import Data.Time (LocalTime (..), ZonedTime (..), fromGregorianValid, makeTimeOfDayValid,
                   minutesToTimeZone)
 import Data.Void (Void)
-import Text.Megaparsec (Parsec, anySingle, errorBundlePretty, match, satisfy, try, notFollowedBy)
+import Text.Megaparsec (Parsec, anySingle, errorBundlePretty, satisfy, try)
 import Text.Megaparsec.Char (alphaNumChar, char, digitChar, eol, hexDigitChar, space, space1,
                              string, tab)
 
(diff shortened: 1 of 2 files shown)


src/Toml/Parser.hs
```diff
@@ -160,15 +161,30 @@ tableNameP :: Parser Key
 tableNameP = lexeme $ between (char '[') (char ']') keyP
 
 -- Values
-
 decimalP :: Parser Integer
-decimalP = mkInteger <$> decimalStringP
+decimalP = matchInteger >>= validateInteger
   where
-    decimalStringP   = fst <$> match (some digitChar >> many _digitsP)
```
ChShersh: I want to say that I really like the previous implementation with `match` and `try`. I think it was quite efficient. Your implementation uses list of `Char`, but `match` relies on the efficient implementation of the underlying stream and uses of offsets. So I would like to have the implementation that behaves like the old one. But in this case it's kinda tricky to test, that the number is not started with zero but also not zero itself. For example, something like the following wouldn't work (but very close to the solution):

```haskell
mkInteger <$> notFollowedBy (some $ char '0') decimalStringP
```

ChShersh:
@MitchStevens Do you need any help with this? Also, note that the `Toml.Parser` module was refactored recently and you need to add your changes now to another module.

MitchStevens:
Sorry I haven't been looking at it lately, got a little distracted. I will have a look at how to write more efficient parsers and the new directory structure over the next couple of days.

ChShersh:
@MitchStevens Don't worry! Feel free to ask any questions is something is not clear :slightly_smiling_face:

Issue #109, opened 2018-10-16
Add function with type like `TomlCodec a -> TOML -> Either DecodeException a`


willbasky:
SubIssue for #98. 

tomland has two step convertion: 
```
from
Text 
to
1. Toml (with parse)
to 
2. Either DecodeException a (with ??) 
```
It has function function [parse](https://hackage.haskell.org/package/tomland-0.4.0/docs/Toml-Parser.html#v:parse) for first step and [decode](https://hackage.haskell.org/package/tomland-0.4.0/docs/Toml-Bi-Code.html#v:decode) for both steps, but hasn`t function for second step.

About [steps](https://www.youtube.com/watch?v=zP-DgjIcCho).

PR #112, opened 2018-10-17
[#109] Add `decodeToml` function

willbasky:
Resolves [#109](https://github.com/kowainik/tomland/issues/109)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ The change log is available [on GitHub][2].
 * [#82](https://github.com/kowainik/tomland/issues/82):
   Remove `maybeT`. Add `diOptional`.
 * [#109](https://github.com/kowainik/tomland/issues/109):
-  Add function `decodeToml`
+  Add function `decodeToml`.
 
 0.4.0
 =====
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ The change log is available [on GitHub][2].
   Drop support for GHC 8.0.2.
 * [#82](https://github.com/kowainik/tomland/issues/82):
   Remove `maybeT`. Add `diOptional`.
+* [#109](https://github.com/kowainik/tomland/issues/109):
+  Add function `decodeToml`
 
 0.4.0
 =====
--- a/src/Toml/Bi/Code.hs
+++ b/src/Toml/Bi/Code.hs
@@ -12,7 +12,7 @@ module Toml.Bi.Code
          -- * Encode/Decode
        , decode
        , decodeFile
-       , decodeToml
+       , runTomlCodec
        , encode
        ) where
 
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -33,6 +33,8 @@ The change log is available [on GitHub][2].
   Drop support for GHC 8.0.2.
 * [#82](https://github.com/kowainik/tomland/issues/82):
   Remove `maybeT`. Add `diOptional`.
+* [#109](https://github.com/kowainik/tomland/issues/109):
+  Add function `decodeToml`
```
ChShersh:
```suggestion
  Add function `decodeToml`.
```

src/Toml/Bi/Code.hs
```diff
@@ -88,7 +89,12 @@ type TomlCodec a = BiCodec Env St a
 decode :: TomlCodec a -> Text -> Either DecodeException a
 decode codec text = do
     toml <- first ParseError (parse text)
-    runReader (runExceptT $ codecRead codec) toml
+    decodeToml codec toml
+
+-- | Convert toml into user data type.
+decodeToml :: TomlCodec a -> TOML -> Either DecodeException a
+decodeToml codec =
+    runReader (runExceptT $ codecRead codec)
```
ChShersh: I think that this function can be implemented in single line, no need to have extra line break

src/Toml/Bi/Code.hs
```diff
@@ -88,7 +89,12 @@ type TomlCodec a = BiCodec Env St a
 decode :: TomlCodec a -> Text -> Either DecodeException a
 decode codec text = do
     toml <- first ParseError (parse text)
-    runReader (runExceptT $ codecRead codec) toml
+    decodeToml codec toml
+
+-- | Convert toml into user data type.
+decodeToml :: TomlCodec a -> TOML -> Either DecodeException a
+decodeToml codec =
```
ChShersh: `decodeToml` name looks like the obvious choice but I don't think that it's the right. If the module is imported qualified the this will be `Toml.decodeToml` which is kinda awkward... Also, `Toml` is kinda generic suffix, the existing `decode` function can be named `decodeToml` as well. The point here is that `Toml` suffix doesn't introduce hints on how to use the function.

One example of better name: `decodeTOML` -- same as `TOML` data type. But this is still bad.

I think perfect name is `runTomlCodec` because it kinda fits the naming scheme for monadic functions.

If you will implement the dual of `decodeToml`, like `pretty` but function that returns `TOML` instead of `Text`, you could name this function as `execTomlCodec`: like `execState` that returns end state.

Btw, adding `execTomlCodec` is also a good idea :+1:

willbasky: Thank you for explain!

src/Toml/Bi/Code.hs
```diff
@@ -12,6 +12,7 @@ module Toml.Bi.Code
          -- * Encode/Decode
        , decode
        , decodeFile
+       , runTomlCodec
        , encode
```
ChShersh: I think you need to export `execTomlCodec` as well

willbasky:
> I think, that when you apply suggested change, you need to do locally git pull -r origin to bring commits from GitHub to your local machine. 

I saw my commit restart CI and thought it is not approved yet. Ok.

PR #113, opened 2018-10-19
[#22] Report proper type checking error during parsing

### LINKED ISSUE #22: Report proper type checking error during parsing
For example, if I have TOML like this:

```toml
list = ["one", "two"]

[table.name.2]
  2Inner = 42
```

I see no problem. But with the following:

```toml
list = ["one", 123]

[table.name.2]
  2Inner = 42
```

I see not really helpful message.
 
```
3:1:
  |
3 | [table.name.2]
  | ^
Can't type check value!
```

When what I want to see is something like this

```
1:15-17:
  |
1 | list = ["one", 123]
  |                ^^^
Can't type check value! Expected type String, actual type Int
```

Depends on #21.

gahag:
Resolves #22 

The problem is a consequence of typechecking after parsing, using the parser monad. When
the error was produced, as the parser had already consumed the input of the whole array,
the error position was set to the beggining of the next line. A trivial improvement would
be to typecheck before the lexeme parser, which consumes the newline. This way, the error
would be pointed to the end of the array, instead of the next line.

I've implemented a better improvement: to parse the array's elements using the parser that
succeeded to parse the first element. This way, typechecking is done as the array is being
parsed, and type errors are pointed to the exact mismatching element.

A better solution would be to split the parsing and typechecking in two phases. The parser
would preserve the token position in a custom datatype, and the typechecker would use that
data to inform the position of the type mismatch. This would require a major rewrite of the parser and type modules.

### DIFF
--- a/src/Toml/Parser.hs
+++ b/src/Toml/Parser.hs
@@ -17,7 +17,8 @@ module Toml.Parser
 
 -- I hate default Prelude... Do I really need to import all this stuff manually?..
 import Control.Applicative (Alternative (..))
-import Control.Applicative.Combinators (between, count, manyTill, optional, sepEndBy, skipMany)
+import Control.Applicative.Combinators (between, count, manyTill, option, optional, sepEndBy,
+                                        skipMany)
 import Control.Monad (void)
 import Data.Char (chr, isControl)
 import Data.Either (fromRight)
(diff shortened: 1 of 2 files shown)


src/Toml/Parser.hs
```diff
@@ -258,10 +259,25 @@ dateTimeP = lexeme $ try hoursP <|> dayLocalZoned
...
   where
     elements :: Parser [UValue]
-    elements = valueP `sepEndBy` spComma <* skipMany (text ",")
+    elements = do v <- valueP -- Parse the first value to determine the type
+                  sep <- optional spComma
+                  vs <- case sep of
+                          Nothing -> return []
+                          Just _  -> (element v `sepEndBy` spComma) <* skipMany spComma
+                  return (v:vs)
+               <|> return []
```
ChShersh: I think it's better to refactor this parser. Like, have parser `orEmpty` with the type:

```haskell
orEmpy :: Alternative f => f [a] -> f [a]
```

Maybe `parser-combinators` library have it. So you can write `orEmpty $ do ...` instead of `do { ... } <|> return []`.

src/Toml/Parser.hs
```diff
@@ -258,10 +259,25 @@ dateTimeP = lexeme $ try hoursP <|> dayLocalZoned
...
+arrayP = lexeme (between (char '[' *> sc) (char ']') elements) <?> "array"
   where
     elements :: Parser [UValue]
-    elements = valueP `sepEndBy` spComma <* skipMany (text ",")
+    elements = do v <- valueP -- Parse the first value to determine the type
+                  sep <- optional spComma
+                  vs <- case sep of
+                          Nothing -> return []
+                          Just _  -> (element v `sepEndBy` spComma) <* skipMany spComma
+                  return (v:vs)
```
ChShersh: Personal preference, but to me `pure` looks better than `return `and it's also shorter :slightly_smiling_face:

src/Toml/Parser.hs
```diff
@@ -258,10 +259,25 @@ dateTimeP = lexeme $ try hoursP <|> dayLocalZoned
...
+               <|> return []
+
+    element :: UValue -> Parser UValue
+    element = \case
+                 UBool    _ -> UBool    <$> boolP
```
ChShersh: I think that indentation here is too big... 4 spaces should be enough. But I like that you've used `LambdaCase` extension here and aligned `->` under each other! :+1:

src/Toml/Parser.hs
gahag: I found that the option combinator does exactly that. Thanks for the hint.

Issue #118, opened 2018-10-25
Add more bimaps for containers


ChShersh:
We already have `_Array`:

```haskell
_Array :: BiMap a AnyValue -> BiMap [a] AnyValue
```

I propose to add the following combinators as well:

```haskell
_NonEmpty :: BiMap a AnyValue -> BiMap (NonEmpty a) AnyValue
_Set      :: BiMap a AnyValue -> BiMap (Set      a) AnyValue
_HashSet  :: BiMap a AnyValue -> BiMap (HashSet  a) AnyValue
_IntSet   ::                     BiMap IntSet       AnyValue
```

PR #120, opened 2018-10-25
[#118] more container bimaps

jiegillet:
I went ahead with #118 since I had most of them ready to go...

- I added the `containers` dependency 
- I made `_Int` because `_IntSet` required it
- I abused the `iso` function for the sets, I know they are not isomorphisms, but it made the definition much sorter
- I haven't changed CHANGELOG yet

### DIFF
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -1,7 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE GADTs               #-}
-{-# LANGUAGE ScopedTypeVariables #-}
 
 -- | Contains TOML-specific combinators for converting between TOML and user data types.
 
--- a/src/Toml/BiMap.hs
+++ b/src/Toml/BiMap.hs
@@ -35,11 +35,17 @@ module Toml.BiMap
        , _Natural
        , _WordInteger
        , _Word
+       , _IntInteger
+       , _Int
        , _Float
        , _ByteStringText
        , _ByteString
        , _LByteStringText
        , _LByteString
+       , _Set
+       , _IntSet
+       , _HashSet
+       , _NonEmpty
 
        , _Left
        , _Right
@@ -56,6 +62,7 @@ import Data.ByteString (ByteString)
 import Data.Text (Text)
 import Data.Word (Word)
 import Numeric.Natural (Natural)
+import Data.Hashable (Hashable)
 import Text.Read (readMaybe)
 import Data.Time (Day, LocalTime, TimeOfDay, ZonedTime)
 
@@ -69,6 +76,10 @@ import qualified Data.Text.Encoding as T
 import qualified Data.ByteString.Lazy as BL
 import qualified Data.Text.Lazy as TL
 import qualified Data.Text.Lazy.Encoding as TL
+import qualified Data.Set as S
+import qualified Data.HashSet as H
+import qualified Data.IntSet as IS
+import qualified Data.List.NonEmpty as NE
 
 ----------------------------------------------------------------------------
 -- BiMap concepts and ideas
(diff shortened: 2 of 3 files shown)


yanikkoval:
What a quick squirrel 🐿️

src/Toml/BiMap.hs
```diff
@@ -69,6 +76,10 @@ import qualified Data.Text.Encoding as T
 import qualified Data.ByteString.Lazy as BL
 import qualified Data.Text.Lazy as TL
 import qualified Data.Text.Lazy.Encoding as TL
+import qualified Data.Set as S
+import qualified Data.HashSet as H
```
vrom911: It's not a very clear name, I guess, `HS` or `HashSet` would be better here 🙂

src/Toml/BiMap.hs
```diff
@@ -198,16 +209,27 @@ _Natural :: BiMap Natural AnyValue
...
           | n > toInteger (maxBound :: Word) = Nothing
           | otherwise                        = Just (fromIntegral n)
 
 _Word :: BiMap Word AnyValue
 _Word = _WordInteger >>> _Integer
 
+_IntInteger :: BiMap Int Integer
+_IntInteger = BiMap (Just . toInteger) maybeInt
+  where maybeInt :: Integer -> Maybe Int
+        maybeInt n
```
vrom911: Probably it makes sense to create a polymorphic function for this with the necessary constraints for it as it's used in several places?
Something like:

```haskell
foo :: (Ord a, Integral a, Integral b, Bounded b) => a -> Maybe b
```

src/Toml/BiMap.hs
```diff
@@ -198,16 +209,27 @@ _Natural :: BiMap Natural AnyValue
...
+          | n < toInteger (minBound :: Int) = Nothing
```
vrom911: With `TypeApplications` it looks neater: `minBound @Int` (but it's up to you 🙂)

src/Toml/BiMap.hs
```diff
@@ -232,6 +254,20 @@ _Array elementBimap = BiMap
     , backward = \(AnyValue val) -> matchArray (backward elementBimap) val
     }
 
+_NonEmpty :: BiMap a AnyValue -> BiMap (NE.NonEmpty a) AnyValue
+_NonEmpty bimap = BiMap (Just . NE.toList) toMaybeNonEmpty >>> _Array bimap
+  where toMaybeNonEmpty [] = Nothing
+        toMaybeNonEmpty x = Just (NE.fromList x)
```
vrom911: This is just [`nonEmpty`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-List-NonEmpty.html#v:nonEmpty) function.

src/Toml/BiMap.hs
```diff
@@ -232,6 +254,20 @@ _Array elementBimap = BiMap
...
+
+_Set :: (Ord a) => BiMap a AnyValue -> BiMap (S.Set a) AnyValue
+_Set bimap = iso S.toList S.fromList >>> _Array bimap
```
ChShersh: Using `iso` here is fine actually :+1:

tomland.cabal
```diff
@@ -53,6 +53,7 @@ library
                      , text ^>= 1.2
                      , time >= 1.8 && < 1.10
                      , transformers ^>= 0.5
+                     , containers ^>= 0.5
```
ChShersh: I think here versions should be `containers >= 0.5.7 && < 0.7`. Also, could you please put this dependency according to alphabetic order?

src/Toml/BiMap.hs
jiegillet: Thank you, I missed it in the docs. I've never used `NonEmpty` before

src/Toml/BiMap.hs
jiegillet: I tried to do that before I submitted and gave up. 
```haskell
integerToBounded :: (Integral a, Bounded a) => Integer -> Maybe a
integerToBounded n
  | n < toInteger (minBound :: a) = Nothing
  | n > toInteger (maxBound :: a) = Nothing
  | otherwise  = Just (fromIntegral n)
```
This won't compile because of the `(minBound :: a)`. It's telling me `Could not deduce (Bounded a1) arising from a use of ‘minBound’ from the context: (Integral a, Bounded a)`. Any suggestion?

src/Toml/BiMap.hs
jiegillet: I didn't actually know that notation, thank you. I left it as is for the moment in case I manage to define the polymorphic function that alludes me at the moment.

src/Toml/BiMap.hs
vrom911: To be able to use `a` type in the function body you need `ScopedTypeVariables` extension. Actually, you can add it to default extensions in the `.cabal` file, it very useful one, we usually have it in our projects as default.

src/Toml/Bi/Combinators.hs
```diff
@@ -1,7 +1,6 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE GADTs               #-}
-{-# LANGUAGE ScopedTypeVariables #-}
```
vrom911: Thanks for the clean-up!

src/Toml/BiMap.hs
```diff
@@ -189,38 +203,47 @@ _Show = _ShowString >>> _String
...
 _Natural = _NaturalInteger >>> _Integer
 
-_WordInteger :: BiMap Word Integer
-_WordInteger = BiMap (Just . toInteger) maybeInteger
-  where maybeInteger :: Integer -> Maybe Word
-        maybeInteger n
-          | n < toInteger (minBound :: Word) = Nothing
-          | n > toInteger (maxBound :: Word) = Nothing
-          | otherwise                        = Just (fromIntegral n)
+_BoundedInteger :: (Integral a, Bounded a) => BiMap a Integer
```
vrom911: That's wonderful! So it can be used to create a similar function for `Int32` or any other type that you could need in your code! I like this solution 👍

jiegillet: I like it too, much cleaner than a copy/paste function for each possible type ^^

PR #121, opened 2018-10-26
[#14] Parsing inline tables

### LINKED ISSUE #14: Add parsing of inline tables
This might be convenient with #13.

jiegillet:
Resolves #14 

I am fairly jet-lagged at the moment so this should probably be considered work in progress, especially since I have not tested the parser very extensively yet.

### DIFF
--- a/src/Toml/Parser/TOML.hs
+++ b/src/Toml/Parser/TOML.hs
@@ -1,47 +1,46 @@
 module Toml.Parser.TOML
-       ( keyValP
+       ( hasKeyP
        , tableHeaderP
        , inlineTableP
        , tomlP
        ) where
 
-import Control.Applicative (Alternative (many, (<|>)))
-import Control.Applicative.Combinators (sepEndBy, between)
-import Data.Either (lefts, rights)
+import Control.Applicative (Alternative (many))
+import Control.Applicative.Combinators (sepEndBy, between, eitherP)
 
-import Toml.Parser.Core (Parser, sc, text, char, try)
+import Toml.Parser.Core (Parser, sc, text)
 import Toml.Parser.Value (keyP, tableNameP, anyValueP)
 import Toml.PrefixTree (Key (..), fromList)
 import Toml.Type (AnyValue, TOML (..))
 
 import qualified Data.HashMap.Lazy as HashMap
 
-hasKeyP :: Parser (Either (Key, AnyValue) (Key, TOML))
-hasKeyP = try (Left <$> keyValP) <|> (Right <$> inlineTableP)
+hasKeyP :: Parser (Key, Either AnyValue TOML)
+hasKeyP = (,) <$> keyP <* text "=" <*> eitherP anyValueP inlineTableP
 
-keyValP :: Parser (Key, AnyValue)
-keyValP = (,) <$> keyP <* text "=" <*> anyValueP
-
-inlineTableP :: Parser (Key, TOML)
-inlineTableP =
-  (,) <$> keyP <* text "="
-      <*> between (chSc '{') (chSc '}')
-                  (makeToml <$> hasKeyP `sepEndBy` chSc ',')
-  where
-    chSc :: Char -> Parser ()
-    chSc c = char c *> sc
+inlineTableP :: Parser TOML
+inlineTableP = between (text "{") (text "}")
+                       (makeToml <$> hasKeyP `sepEndBy` text ",")
 
 tableHeaderP :: Parser (Key, TOML)
 tableHeaderP = (,) <$> tableNameP <*> (makeToml <$> many hasKeyP)
 
-makeToml :: [Either (Key, AnyValue) (Key, TOML)] -> TOML
-makeToml kvs = TOML (HashMap.fromList $ lefts kvs) (fromList $ rights kvs)
+distributeEithers :: [(c, Either a b)] -> ([(c, a)], [(c, b)])
+distributeEithers = foldr distribute ([], [])
+  where
+    distribute (k, Left a) (ls, rs) = ((k, a) : ls, rs)
+    distribute (k, Right b) (ls, rs) = (ls, (k, b) : rs)
+
+makeToml :: [(Key, Either AnyValue TOML)] -> TOML
+makeToml kvs = TOML (HashMap.fromList lefts) (fromList rights)
+  where
+    (lefts, rights) = distributeEithers kvs
 
 tomlP :: Parser TOML
 tomlP = do
     sc
-    kvs    <- many hasKeyP
+    (val, inline) <- distributeEithers <$> many hasKeyP
     tables <- many tableHeaderP
-    return TOML { tomlPairs  = HashMap.fromList $ lefts kvs
-                , tomlTables = fromList $ tables ++ rights kvs
+    return TOML { tomlPairs  = HashMap.fromList val
+                , tomlTables = fromList $ tables ++ inline
                 }
--- a/src/Toml/Parser/TOML.hs
+++ b/src/Toml/Parser/TOML.hs
@@ -28,6 +28,7 @@ tableHeaderP = (,) <$> tableNameP <*> (makeToml <$> many hasKeyP)
 distributeEithers :: [(c, Either a b)] -> ([(c, a)], [(c, b)])
 distributeEithers = foldr distribute ([], [])
   where
+    distribute :: (c, Either a b) -> ([(c, a)], [(c, b)]) -> ([(c, a)], [(c, b)])
     distribute (k, Left a) (ls, rs) = ((k, a) : ls, rs)
     distribute (k, Right b) (ls, rs) = (ls, (k, b) : rs)
 
--- a/src/Toml/Parser/Value.hs
+++ b/src/Toml/Parser/Value.hs
@@ -288,7 +288,6 @@ valueP = UBool    <$> boolP
 
 
 anyValueP :: Parser AnyValue
-anyValueP = valueP >>= \uval ->
-  case typeCheck uval of
+anyValueP = typeCheck <$> valueP >>= \case
     Left err -> fail $ show err
     Right v  -> return v
--- a/src/Toml/Parser/Value.hs
+++ b/src/Toml/Parser/Value.hs
@@ -142,7 +142,7 @@ keyP = Key <$> NC.sepBy1 keyComponentP (char '.')
 
 
 tableNameP :: Parser Key
-tableNameP = lexeme $ between (char '[') (char ']') keyP
+tableNameP = between (text "[") (text "]") keyP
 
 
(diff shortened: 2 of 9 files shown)


src/Toml/Parser/TOML.hs
```diff
@@ -1,44 +1,47 @@
...
+import Toml.Parser.Core (Parser, sc, text, char, try)
+import Toml.Parser.Value (keyP, tableNameP, anyValueP)
 import Toml.PrefixTree (Key (..), fromList)
-import Toml.Type (AnyValue, TOML (..), typeCheck)
+import Toml.Type (AnyValue, TOML (..))
 
 import qualified Data.HashMap.Lazy as HashMap
 
+hasKeyP :: Parser (Either (Key, AnyValue) (Key, TOML))
+hasKeyP = try (Left <$> keyValP) <|> (Right <$> inlineTableP)
```
ChShersh: I think that this parser should be rewrittern in the following way:

1. Change type to `Parser (Key, Either TOML AnyValue)`. You will always parse key, so it makes sense to move it out.
2. If you parse `inlineTable` first than you don't need `try` and I think that this is better.
3. You can use `eitherP` from `parser-combinators` library:
    * https://hackage.haskell.org/package/parser-combinators-1.0.0/docs/Control-Applicative-Combinators.html#v:eitherP

src/Toml/Parser/TOML.hs
```diff
@@ -1,44 +1,47 @@
...
 tomlP = do
     sc
-    kvs    <- many keyValP
+    kvs    <- many hasKeyP
     tables <- many tableHeaderP
-    pure TOML { tomlPairs  = HashMap.fromList kvs
-              , tomlTables = fromList tables
-              }
+    return TOML { tomlPairs  = HashMap.fromList $ lefts kvs
+                , tomlTables = fromList $ tables ++ rights kvs
```
ChShersh: You use `lefts` and `rights` function and because of that you traverse list twice and this is not efficient. It's better if you write polymorphic function like:

```haskell
distributeEithers :: [(c, Either a b)] -> ([(c, a)], [(c, b)])
```

src/Toml/Parser/TOML.hs
```diff
@@ -1,44 +1,47 @@
...
 tomlP :: Parser TOML
 tomlP = do
     sc
-    kvs    <- many keyValP
+    kvs    <- many hasKeyP
     tables <- many tableHeaderP
-    pure TOML { tomlPairs  = HashMap.fromList kvs
-              , tomlTables = fromList tables
-              }
+    return TOML { tomlPairs  = HashMap.fromList $ lefts kvs
```
ChShersh: Is there any reason why `return` was changed to `pure`?

src/Toml/Parser/Value.hs
```diff
@@ -284,3 +285,10 @@ valueP = UBool    <$> boolP
      <|> UInteger <$> integerP
      <|> UText    <$> textP
      <|> UArray   <$> arrayP
+
+
+anyValueP :: Parser AnyValue
+anyValueP = valueP >>= \uval ->
+  case typeCheck uval of
+    Left err -> fail $ show err
```
ChShersh: I feel like this function can be written in the following way:

```haskell
anyValueP = typeCheck <$> valueP >>= \case
    Left err -> ...
    Right -> ...
```

Personally, to me the priority of `<$>` and `>>=` is always clear and this code is anambiguous. But when you write something like `foo >>= x =<< bar` you should reconsider your coding style... Combination of `<$>` and `>>=` is totally valid.

src/Toml/Parser/TOML.hs
jiegillet: I just prefer using monadic functions while using the do notation. I can switch it back of you want.

src/Toml/Parser/Value.hs
jiegillet: That's a good suggestion, but did I use `foo >>= x =<< bar` somewhere?

jiegillet:
Aw, the commit history is a mess, sorry, it's probably because I switched computers. I think this is a GitHub bug that's I've seen before, it says that 8 files have been changed, but when I do `git diff upstream/master` I only see differences in 2 files. `git rebase` also doesn't help. However, the PR can still be merged and there will only be changes in the 2 relevant files.

src/Toml/Parser/Value.hs
ChShersh: > but did I use foo >>= x =<< bar somewhere?

No, you haven't 🙂 I saw it once previously. The idea here is that operators can be a mess but combination of `<$>` and `>>=` is actually quite idiomatic and fine 👍

src/Toml/Parser/TOML.hs
ChShersh: I usually prefer `pure` because it's shorter 🙂 In some sense `pure` name is more idiomatic because you _lift pure value into monadic context_.

ChShersh:
@jiegillet I see that you have merge commits in your existing branch. This usually happens when you do `git pull` because by default `git pull` uses `merge` and creates extra `Merge branch ...` commit (which I personally find quite annoying and I'm not using `git pull`). The workaround here is to use `git pull -r` (with `-r` flag): with this flag it uses `rebase` strategy, so instead of merging branch into your branch and sorting commits in chronological way, it puts your local commits on top of the branch you're synching with and make the history look linear and sequential.

My strategy regarding synching with `master` branch:

```haskell
$ git fetch origin master # in your case instead of `origin` here should be something like `owner` which points to this repo, not your fork
$ git rebase origin/name-of-the-branch
```

The same can be done to sync with your branch.

Regarding commits in this PR: it's okay to have mess in the PR because we're using `Squash and merge` strategy so every mess in any PR will be collapsed in single useful commit. But you can use interactive rebase to prettify commit history in your branch, like squash all commits into single one. So Github will show nicer diff.

src/Toml/Parser/TOML.hs
```diff
@@ -1,44 +1,46 @@
...
-    toml <- makeToml <$> many keyValP
-    pure (k, toml)
+tableHeaderP = (,) <$> tableNameP <*> (makeToml <$> many hasKeyP)
+
+distributeEithers :: [(c, Either a b)] -> ([(c, a)], [(c, b)])
+distributeEithers = foldr distribute ([], [])
   where
-    makeToml :: [(Key, AnyValue)] -> TOML
-    makeToml kv = TOML (HashMap.fromList kv) mempty
+    distribute (k, Left a) (ls, rs) = ((k, a) : ls, rs)
```
ChShersh: Could you please add explicit type signature to the `distribute` function?

src/Toml/Parser/TOML.hs
```diff
@@ -1,44 +1,46 @@
...
 
 tableHeaderP :: Parser (Key, TOML)
-tableHeaderP = do
-    k    <- tableNameP
-    toml <- makeToml <$> many keyValP
-    pure (k, toml)
+tableHeaderP = (,) <$> tableNameP <*> (makeToml <$> many hasKeyP)
+
+distributeEithers :: [(c, Either a b)] -> ([(c, a)], [(c, b)])
+distributeEithers = foldr distribute ([], [])
```
ChShersh: I wonder whether usage of `foldr` is okay here and whether it leaks memory or not 🤔

src/Toml/Parser/TOML.hs
```diff
@@ -1,44 +1,46 @@
...
-keyValP = do
-    k    <- keyP
-    _    <- text "="
-    uval <- valueP
-    case typeCheck uval of
-        Left err -> fail $ show err
-        Right v  -> pure (k, v)
-
+inlineTableP :: Parser TOML
+inlineTableP = between (text "{") (text "}")
```
ChShersh: I feel like `char '{'` can be more efficient here but it probably will work worse with consuming spaces after `{` so having `text` is okay for now 👍

src/Toml/Parser/TOML.hs
jiegillet: My first instinct was to use `Data.List.partition :: (a -> Bool) -> [a] -> ([a],[a])` but I wanted to use pattern matching. Also, looking at it now, it's implemented with `foldr`. 
I don't know anything about memory leaks, but I'm pretty sure `foldr` is used a lot. Also, I doubt there would be toml files with that many keys to cause trouble.

src/Toml/Parser/TOML.hs
jiegillet: Speaking of which: the TOML specifications say "it is strongly discouraged to break an inline table onto multiples lines" and somewhere else I found the same info with even stronger language. Right now `inlineTableP` can parse inline tables with line breaks, and it would probably messy to change that. What do you think? 
I think it's probably fine to have the capability of parsing, as long as we are never pretty-printing an inline table on multiple lines.

ChShersh:
@jiegillet You can enable `-r` by default for `git pull` command. See example of how to do this in my `.gitconfig` settings:

* https://github.com/ChShersh/dotfiles/blob/6c19ebd2d9b7de8285ac84d4bc766c9285e2cc1a/.gitconfig#L9-L10

But I still enable paranoia mode every time and specify `-r` manually because I always can perform this from some other machine with different settings...

src/Toml/Parser/TOML.hs
ChShersh: @jiegillet Sure, feel free to replace! It would be a good addition 👍

ChShersh: > Right now inlineTableP can parse inline tables with line breaks, and it would probably messy to change that. What do you think?

I'm okay if `tomland` parses inline tables on multiple lines. And I understand that it will be hard to fix with existing code because spaces and line breaks are all the same for megaparsec and introducing distinction will be hard...

src/Toml/Parser/TOML.hs
ChShersh: @jiegillet Yeah, `foldr` looks okay in this case, it's just not obvious whether it's totally okay. `foldr` can introduce space leaks, so it should be used with care. I also don't expect the tables to be big, so it's okay to leave it as it is now.

ChShersh:
@jiegillet Currently I see the error in tests:

```
      can parse an empty TOML table:                                                             FAIL
        expected: (Key {unKey = Piece {unPiece = "table"} :| []},Right (TOML {tomlPairs = fromList [], tomlTables = fromList []}))
        but parsing failed with error:
          1:1:
            |
          1 | [table]
            | ^
          unexpected '['
          expecting '"', ''', '-', '_', or alphanumeric character
```

Looks like empty inline tables like `{}` can't be parsed...

jiegillet:
Whoops. Actually the parser it fine, the test had a problem :)
(and yeah, I know how suspicious that sounds)

Issue #122, opened 2018-10-26
Boil `htoml` for benchmark


willbasky:
subissue of #111
As I see [htoml](https://hackage.haskell.org/package/htoml) has `parseTomlDoc` for parse raw toml and  `toHaskell :: TOML -> HaskellType` via JSON.
To implement: 
- [ ]  `decode :: Text -> HaskellType`

Issue #98, opened 2018-10-02
Write `tomland` benchmarks


ChShersh:
We're interested only in bechmarks for parsing (mostly because other Haskell TOML libraries don't have pretty-printing abilities) and for converting to Haskell data types (aka 2 phases). See this talk by George Wilson about CSV encoding/decoding library for multiple phases explanation:

* https://www.youtube.com/watch?v=zP-DgjIcCho

I would like to have a comparison with the following libraries:

* [`htoml`](http://hackage.haskell.org/package/htoml)
* [`htoml-megaparsec`](https://hackage.haskell.org/package/htoml-megaparsec)
* [`toml-parser`](https://hackage.haskell.org/package/toml-parser)
* [`toml`](http://hackage.haskell.org/package/toml)

willbasky:
What functions/actions/operations to benchmark?

willbasky:
It seems `htoml` has just parser step. 
`parseTomlDoc :: String -> Text -> Either ParseError Table` where `Table` is HashMap with Text keys and Node as value, the last is like `AnyValue` in tomland. Am I right?

willbasky:
`toml-parser` has just parser, without convertor. Is it correct?

ChShersh:
@willbasky 

> What functions/actions/operations to benchmark?

I suppose we should have some TOML example file with multiple fields:

* Bool
* Integer
* Text
* DateTime
* Array
* Table1
* Table2

I would like to have different types of TOML values, so we can put inside tables 1 and 2 something different. For example, if you put ZonedTIme on top-level `DateTime` then you can put `LocalDay` in table1 and just time in Table2. Same for array: you can have array of strings, array of different doubles (with nan and inf) and array of array of integers. String can be either raw or quoted or multiline string. Just some random different values to test everything.

> Do all other libraries have two step converting?

Not sure. But for those, who don't have, we need to write conversion manually by our own just for this library.

> Does tomland has separate function for parsing and for converting?
> `decode :: TomlCodec a -> Text -> Either DecodeException a` function includes both steps.

`tomland` has separate function only for parsing:
* http://hackage.haskell.org/package/tomland-0.4.0/docs/Toml-Parser.html#v:parse

But it doesn't have separate function only for conversion. Probably it's a good idea to have one.

So I I see the following steps to work on this issue:

1. Add function with type like `TomlCodec a -> TOML -> Either DecodeException a`
2. Implement example of TOML for the benchmarks.
3. Write data type, that describes this example.
[...]

PR #126, opened 2018-10-30
[#122] Boil htoml for benchmark

willbasky:
Resolve [#122](https://github.com/kowainik/tomland/issues/122)


willbasky:
My plan was to write decode functions for libraries in `boil` issues
and then write benchmark at once for all of them within #98.

benchmark/Benchmark/Type.hs
```diff
@@ -77,13 +63,14 @@ instance FromJSON SizeInside where
...
-    <$> Toml.text "title" .= testTitle
-    <*> Toml.double "atom" .= testAtom
-    <*> Toml.bool "cash" .= testCash
-    <*> Toml.arrayOf Toml._Text "words" .= testWords
-    <*> Toml.arrayOf Toml._Bool "bool" .= testBool
-    <*> Toml.zonedTime "today" .= testToday
-    <*> Toml.table insideF "fruit" .= testFruit
-    <*> Toml.table insideS "size" .= testSize
+-- | Codec to use in tomland decode and convert functions.
+codec :: TomlCodec HaskellType
```
ChShersh: Let's move `tomland` codecs into module called `Benchmark.Tomland`

tomland.cabal
```diff
@@ -138,6 +139,8 @@ benchmark tomland-benchmark
                       -Wmissing-export-lists
                       -Wpartial-fields
   build-depends:      base
+                    , aeson
+                    , htoml
```
ChShersh: Let's specify version of `htoml` explicitly

benchmark/Benchmark/Type.hs
```diff
@@ -1,10 +1,12 @@
 module Benchmark.Type
-       ( test
+       ( Test(..)
```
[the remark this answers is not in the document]
ChShersh: Yeah, agree, but I couldn't come up with better name. `HaskellType` is slightly better but I still feel that it can be improved 🙂

ChShersh:
Regarding testing: I guess we don't need to test that TOML parses well with libraries. We can decode and just do `error` on `Nothing` or on `Left _`.

benchmark/Benchmark/Compare.hs
```diff
@@ -0,0 +1,45 @@
+module Benchmark.Compare
```
ChShersh: I think that this module can be safely renamed to `Main`. We already have `Main.hs` in this package, so need to create `Benchmark.Compare` as a separate module.

benchmark/Benchmark/Compare.hs
```diff
@@ -0,0 +1,45 @@
...
+       (main
+       ) where
+
+import Data.Text (Text)
+import Gauge.Main (bench, bgroup, defaultMain, whnf)
```
ChShersh: You should use `nf` everywhere here instead of `whnf`. Because data types are very nested, it's not enough to evaluate only to weak-head-normal form, you need to evaluate to normal form.

benchmark/Benchmark/Compare.hs
```diff
@@ -0,0 +1,45 @@
...
+
+import qualified Data.Text.IO as TIO
+
+
+-- | Benchmark.
+main :: IO ()
+main = do
+    (txt, htoml, tomland) <- preload
+    defaultMain
+        [ bgroup "tomland"
```
ChShersh: I feel like instead of grouping by library it makes more sense to group by stage. So you can visually compare, for example, now parsing works for different libraries, because in the terminal they all will be near each other.

benchmark/Benchmark/Compare.hs
```diff
@@ -0,0 +1,45 @@
...
+            [ bench "parse"   $ whnf parse txt
+            , bench "convert" $ whnf (runTomlCodec codec) tomland
+            , bench "decode"  $ whnf (decode codec) txt
```
ChShersh: I think every module in `Benchmark.*` should export three function: `parse`, `convert` and `decode` so you can just write:

```haskell
bgroup "parse"
    [ bench "tomland" $ nf Tomland.parse txt
    , bench "htoml"   $ nf   Htoml.parse txt
    ...
```

benchmark/Benchmark/Compare.hs
```diff
@@ -0,0 +1,45 @@
...
+-- | Preload toml file, and parsed version for libraries.
+preload :: IO (Text, Table, TOML)
+preload = do
+    txt <- TIO.readFile "./benchmark/benchmark.toml"
+    case parseTomlDoc "" txt of
+        Left _      -> error "Htoml parsing failed"
+        Right htoml ->
+            case parse txt of
+                Left _        -> error "Tomland parsing failed"
+                Right tomland -> pure (txt, htoml, tomland)
```
ChShersh: I feel like this pattern-matching of `Left` and `Right` can become messy with addition of more libraries... What if you do the following in `main` function?

1. Read content of the file into variable `txt`
2. Have variables: `tomlandVal`, `htomlVal` for intermediate representations of parsed types to values.

So the code will look like:

```haskell
txt <- readTomlExample
tomlandVal <- unsafeFromRight $ Tomland.parse txt
htomlVal   <- unsafeFromRight $ Htoml.parse txt
```

Probably something better instead of `Val` should be used as suffix...

benchmark/Benchmark/Type.hs
```diff
@@ -1,73 +1,53 @@
...
-    <*> Toml.timeOfDay "due" .= fiDue
+instance FromJSON FruitInside where
+    parseJSON = withObject "FruitInside" $ \o -> FruitInside
+        <$> o .: "name"
+        <*> o .: "description"
 
 newtype SizeInside = SizeInside
     { unSize :: [[Double]]
     }
+    deriving Show
```
ChShersh: According to our style-guide `deriving` should go on the same line as `}` after `}`

src/Toml/Type/AnyValue.hs
```diff
@@ -31,7 +33,12 @@ instance Show AnyValue where
 instance Eq AnyValue where
     (AnyValue val1) == (AnyValue val2) = case sameValue val1 val2 of
         Right Refl -> val1 == val2
-        Left _ -> False
+        Left _     -> False
+
+instance NFData AnyValue where
+    rnf = rwhnf
+
+instance Generic AnyValue
```
ChShersh: Heh.. Deriving of `Generic` is not possible for existential types like `AnyValue`. So let's remove this instance.

src/Toml/Type/AnyValue.hs
```diff
@@ -31,7 +33,12 @@ instance Show AnyValue where
 instance Eq AnyValue where
     (AnyValue val1) == (AnyValue val2) = case sameValue val1 val2 of
         Right Refl -> val1 == val2
-        Left _ -> False
+        Left _     -> False
+
+instance NFData AnyValue where
+    rnf = rwhnf
```
ChShersh: I think that here it should be:

```haskell
rnf (AnyValue val) = rnf val
```

tomland.cabal
```diff
@@ -47,6 +47,7 @@ library
   build-depends:       base >= 4.10 && < 5
                      , bytestring ^>= 0.10
                      , containers >= 0.5.7 && < 0.7
+                     , deepseq
```
ChShersh: Let's specify bounds for `deepseq` as well

tomland.cabal
```diff
@@ -122,7 +123,7 @@ benchmark tomland-bench
   hs-source-dirs:     benchmark
   main-is:            Main.hs
 
-  other-modules:      Benchmark.Compare
+  other-modules:      Benchmark.Main
```
ChShersh: It shouldn't be `Benchmark.Main`. Couple lines above you have: `main-is: Main.hs` so you can put everything from `Benchmark.Main` module to `Main.hs` module.

benchmark/Benchmark/Htoml.hs
```diff
@@ -0,0 +1,30 @@
...
+decode txt = case parseTomlDoc "" txt of
+    Left _ -> error "Parsing failed"
+    Right toml -> convert toml
+
+-- | Convert from already parsed toml to Haskell type.
+convert :: Table -> Either String HaskellType
+convert = parseEither parseJSON . TTT.toJSON
+
+instance NFData Node where
+    rnf = rwhnf
```
ChShersh: Such instances are not reliable. It's better to use `StandaloneDeriving` extension to derive instances.

benchmark/Benchmark/Htoml.hs
```diff
@@ -0,0 +1,30 @@
+module Benchmark.Htoml
```
ChShersh: To get rid of orphans warning, add the following to the top of this module:

```haskell
{-# OPTIONS_GHC -fno-warn-orphans #-}
```

benchmark/Benchmark/Main.hs
```diff
@@ -0,0 +1,35 @@
...
+import qualified Benchmark.Htoml as Htoml
+import qualified Toml.Bi.Code as Tomland
+
+
+-- | Benchmark.
+main :: IO ()
+main = do
+    txt <- TIO.readFile "./benchmark/benchmark.toml"
+    Right htomlVal <- pure $ parseTomlDoc "" txt
+    Right tomlandVal <- pure $ parse txt
```
ChShersh: I think that `parse` functions should be reexported from corresponding modules as well. So here you will have just:

```haskell
Right htomlVal <- pure $ Htoml.parse txt
Right tomlandVal <- pure $ Tomland.parse txt
```

benchmark/Benchmark/Main.hs
```diff
@@ -0,0 +1,35 @@
...
+    defaultMain
+        [ bgroup "Parse"
+            [ bench "tomland" $ nf parse txt
+            , bench "htoml"   $ nf (parseTomlDoc "") txt
+            ]
+        , bgroup "Convert"
+            [ bench "tomland" $ nf (Tomland.runTomlCodec codec) tomlandVal
```
ChShersh: Let's reexport this as a `convert` function from `Benchmark.Tomland` module.

benchmark/Benchmark/Main.hs
```diff
@@ -0,0 +1,35 @@
...
+            , bench "htoml"   $ nf Htoml.convert htomlVal
+            ]
+        , bgroup "Decode"
+            [ bench "tomland" $ nf (Tomland.decode codec) txt
```
ChShersh: You can have `decode` module function in the `Benchmark.Tomland` module which is `decode` from the `tomland` library partially applied to `codec`.

src/Toml/Bi/Code.hs
```diff
@@ -44,7 +48,7 @@ data DecodeException
     | TableNotFound Key  -- ^ No such table
     | TypeMismatch Key Text TValue  -- ^ Expected type vs actual type
     | ParseError ParseException  -- ^ Exception during parsing
-    deriving (Eq, Show)  -- TODO: manual pretty show instances
+    deriving (Eq, Generic, NFData, Show)  -- TODO: manual pretty show instances
```
ChShersh: Just to have consistent deriving list
```suggestion
    deriving (Eq, Show, Generic, NFData)  -- TODO: manual pretty show instances
```

benchmark/Benchmark/Htoml.hs
```diff
@@ -0,0 +1,44 @@
+{-# LANGUAGE DeriveAnyClass     #-}
+{-# LANGUAGE DeriveGeneric      #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
ChShersh: According to our style guide, `OPTIONS_GHC` should go before `{-# LANGUAGE #-}` in separate section. See example here:

* https://github.com/kowainik/summoner/blob/f5b12ed01e3f5c607398f443153f4f5746557c50/src/Summoner/Process.hs#L1-L4

benchmark/Benchmark/Htoml.hs
```diff
@@ -0,0 +1,44 @@
...
+-- | Convert from already parsed toml to Haskell type.
+convert :: Table -> Either String HaskellType
+convert = parseEither parseJSON . TTT.toJSON
+
+-- | Wrapper on htoml's parseTomlDoc
+parse :: Text -> Either ParseError Table
+parse = parseTomlDoc ""
+
+deriving instance NFData Node
+
```
ChShersh:
```suggestion
```

benchmark/Benchmark/Htoml.hs
```diff
@@ -0,0 +1,44 @@
...
+deriving instance Generic Node
+
+instance NFData ParseError where
+    rnf = rwhnf
```
ChShersh: Argh, `ParseError` from `parsec` library doesn't export it's constructor so it's not possible to use `StandaloneDeriving` with it 😠 But the instance for `ParseError` is not that import, we just need to have one for compilation, instances for data types are more important.

benchmark/Benchmark/Htoml.hs
```diff
@@ -0,0 +1,44 @@
...
+
+import Text.Toml (parseTomlDoc)
+import Text.Toml.Types (Node, Table)
+
+import qualified Text.Toml.Types as TTT
+
+
+-- | Decode toml file to Haskell type.
+decode :: Text -> Either String HaskellType
+decode txt = case parseTomlDoc "" txt of
```
ChShersh: Well, since you pattern-match on `Right` constructor in the `main` module already, we don't need to have explicit error here. You can just write:

```haskell
decode = parse >=> convert
```

ChShersh: Oh, wait. No, you can't, because `Left` constructor has different types for `parse` and `convert` functions. Okay, nevermind.

benchmark/Benchmark/Tomland.hs
```diff
@@ -0,0 +1,39 @@
+module Benchmark.Tomland
+       ( convert
+       , decode
```
ChShersh: Let's also reexport and use `parse` from here for consistency across benchmark suits.

benchmark/Main.hs
```diff
@@ -1,4 +1,32 @@
-module Main (main) where
+module Main
+       (main
```
ChShersh:
```suggestion
       ( main
```

src/Toml/PrefixTree.hs
```diff
@@ -42,7 +44,13 @@ import qualified Data.Text as Text
 
 -- | Represents the key piece of some layer.
 newtype Piece = Piece { unPiece :: Text }
-    deriving (Show, Eq, Ord, Hashable, IsString)
+    deriving (Show, Eq, Ord, Hashable, Generic)
```
ChShersh: I think here you can use `DerivingStrategies` extension to not write instances manually. This is a typical problem with `DeriveAnyClass` and `GeneralizedNewtypeDeriving` because GHC doesn't know what strategy are you using. But since GHC 8.2.2 you can specify strategy explicitly.

```suggestion
    deriving stock (Generic)
    deriving newtype (Show, Eq, Ord, Hashable, IsString, NFData)
```

src/Toml/PrefixTree.hs
```diff
@@ -59,10 +67,16 @@ Key (Piece "site" :| [Piece "\\"google.com\\""])
 
 -}
 newtype Key = Key { unKey :: NonEmpty Piece }
-    deriving (Show, Eq, Ord, Semigroup, Generic)
+    deriving (Show, Eq, Ord, Generic)
```
ChShersh: Same here as for `Piece` data type

src/Toml/Type/AnyValue.hs
```diff
@@ -31,7 +32,10 @@ instance Show AnyValue where
 instance Eq AnyValue where
     (AnyValue val1) == (AnyValue val2) = case sameValue val1 val2 of
         Right Refl -> val1 == val2
-        Left _ -> False
+        Left _     -> False
+
+instance NFData AnyValue where
+    rnf (AnyValue val) = rwhnf val
```
ChShersh: I wonder, whether's it's possible to use `rnf` here instead of `rwhnf`? 🤔 Because `rwhnf` is not accurate enough... :(

tomland.cabal
```diff
@@ -137,7 +139,13 @@ benchmark tomland-benchmark
                       -fhide-source-paths
                       -Wmissing-export-lists
                       -Wpartial-fields
+
   build-depends:      base
+                    , aeson
+                    , deepseq <= 1.4.3.0
```
ChShersh:
```suggestion
                    , deepseq
```

tomland.cabal
ChShersh:
```suggestion
                     , deepseq ^>= 1.4
```

tomland.cabal
```diff
@@ -137,7 +139,13 @@ benchmark tomland-benchmark
...
+                    , gauge
+                    , htoml <= 1.0.0.3
```
ChShersh: It's usually better to specify with upper bound as well. Try to replace `<=` with `^>=`

src/Toml/Type/AnyValue.hs
willbasky: If `rnf` then 
-- deriving instance NFData (Value t)
-- deriving instance Generic (Value t)
and that entails
```
• Can't make a derived instance of ‘Generic (Value t)’:
    Bool must be a vanilla data constructor, and
    Integer must be a vanilla data constructor, and
    Double must be a vanilla data constructor, and
    Text must be a vanilla data constructor, and
    Date must be a vanilla data constructor, and
    Array must be a vanilla data constructor
```

willbasky: need some :ice_cream:

ChShersh: @willbasky Yes, unfortunately, you can't derive `Generic` and `NFData` instances for GADTs. It's a known problem. But you can implement `NFData` instance manually. Like:

```
rnf (Integer n) = rnf n
rnf (Text t) = rnf t
...
```

PR #128, opened 2018-11-06
[#64] No leading zeros

jiegillet:
Addressing #64 

I re-implemented `decimalP`, it's much simpler now. 
I've uncommented the test on leading zeroes, and modified the `does not parse numbers` tests to actually test failure and not success. To make those pass I had to use the `lookAhead` trick.

While testing these, I realized that the string `"a = 123.3b=23"` is accepted and parsed like `"a = 123.3\nb=23"`. Same for `"a = trueb=23"` and others. It's kind of a weird case, I don't know if it's worth spending time on. Note: `"a = 123b=23"` does not parse because I've used the `lookAhead` trick for integers, so it's definitely fixable.

### DIFF
--- a/src/Toml/Parser/Value.hs
+++ b/src/Toml/Parser/Value.hs
@@ -25,7 +25,7 @@ import Data.Time (LocalTime (..), ZonedTime (..), fromGregorianValid, makeTimeOf
 
 import Toml.Parser.Core (Parser, alphaNumChar, anySingle, binary, char, digitChar, eol, float,
                          hexDigitChar, hexadecimal, lexeme, octal, satisfy, sc, signed,
-                         space, string, tab, text, try, (<?>), space1, eof)
+                         space, string, tab, text, try, (<?>), space1, eof, oneOf, lookAhead)
 import Toml.PrefixTree (Key (..), Piece (..))
 import Toml.Type (DateTime (..), UValue (..), AnyValue, typeCheck)
 
@@ -148,16 +148,17 @@ tableNameP = between (text "[") (text "]") keyP
 decimalP :: Parser Integer
 decimalP = zero <|> more
   where
-    zero = 0 <$ char '0' <* (space1 <|> eof)
+    zero = 0 <$ char '0'
     more = fst . head . readDec . concat <$> sepBy1 (some digitChar) (char '_')
 
 integerP :: Parser Integer
-integerP = lexeme (bin <|> oct <|> hex <|> dec) <?> "integer"
+integerP = lexeme ((bin <|> oct <|> hex <|> dec) <* lookAhead sep) <?> "integer"
   where
     dec = signed sc decimalP
     bin = try (char '0' >> char 'b') >> binary
     oct = try (char '0' >> char 'o') >> octal
     hex = try (char '0' >> char 'x') >> hexadecimal
+    sep = space1 <|> eof <|> (() <$ oneOf (",=]}" :: String))
 
 
 doubleP :: Parser Double
--- a/src/Toml/Parser/Value.hs
+++ b/src/Toml/Parser/Value.hs
@@ -25,7 +25,7 @@ import Data.Time (LocalTime (..), ZonedTime (..), fromGregorianValid, makeTimeOf
 
 import Toml.Parser.Core (Parser, alphaNumChar, anySingle, binary, char, digitChar, eol, float,
                          hexDigitChar, hexadecimal, lexeme, octal, satisfy, sc, signed,
-                         space, string, tab, text, try, (<?>))
+                         space, string, tab, text, try, (<?>), space1, eof)
 import Toml.PrefixTree (Key (..), Piece (..))
 import Toml.Type (DateTime (..), UValue (..), AnyValue, typeCheck)
 
@@ -146,7 +146,10 @@ tableNameP = between (text "[") (text "]") keyP
 -- Values
 
 decimalP :: Parser Integer
-decimalP = fst . head . readDec . concat <$> sepBy1 (some digitChar) (char '_')
+decimalP = zero <|> more
+  where
+    zero = 0 <$ char '0' <* (space1 <|> eof)
+    more = fst . head . readDec . concat <$> sepBy1 (some digitChar) (char '_')
 
 integerP :: Parser Integer
 integerP = lexeme (bin <|> oct <|> hex <|> dec) <?> "integer"
--- a/src/Toml/Parser/Value.hs
+++ b/src/Toml/Parser/Value.hs
@@ -16,6 +16,7 @@ import Control.Applicative.Combinators (between, count, manyTill, option, option
                                         skipMany, sepBy1)
 
 import Data.Char (chr, isControl)
+import Text.Read (readMaybe)
 import Data.Fixed (Pico)
 import Data.Semigroup ((<>))
 import Data.Text (Text)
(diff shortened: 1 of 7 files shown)


src/Toml/Parser/Value.hs
```diff
@@ -148,22 +146,19 @@ tableNameP = between (text "[") (text "]") keyP
...
-    mkInteger        = textToInt . stripUnderscores
-    textToInt        = fst . fromRight (error "Underscore parser has a bug") . TR.decimal
-    stripUnderscores = Text.filter (/= '_')
-
+    zero = 0 <$ char '0'
+    more = fst . head . readDec . concat <$> sepBy1 (some digitChar) (char '_')
 
 integerP :: Parser Integer
-integerP = lexeme (bin <|> oct <|> hex <|> dec) <?> "integer"
+integerP = lexeme ((bin <|> oct <|> hex <|> dec) <* lookAhead sep) <?> "integer"
```
ChShersh: Hmm, this change requires some thinking 🤔 🤔 🤔 

I also wonder how it affects performance. Let's wait for the benchmarks to be merged to `master` so we can compare performance of this function before and after this change.

ChShersh:
> and modified the does not parse numbers tests to actually test failure and not success

This is kinda weird how those tests passed before 🤔 

> While testing these, I realized that the string "a = 123.3b=23" is accepted and parsed

@jiegillet Is this the old behavior before your change or after?

jiegillet:
> This is kinda weird how those tests passed before 🤔

Well, they were weird tests... Example of old tests:
```haskell
it "does not parse numbers prefixed with `0B`" $ parseInteger "0B1101" 0
```
`parseInteger` would consume the leading `0`, stop there return 0 and pass the test, which makes sense, but if you actually want the parser to fail then the test should really be `integerFailOn "0B1101"`. But then you need to modify the parser so that it only stops consuming digits when it hits a space or something.
 
> > While testing these, I realized that the string "a = 123.3b=23" is accepted and parsed
> 
> @jiegillet Is this the old behavior before your change or after?

That's technically both because `123.3` is a float and I only modified the parser for integers. Before the change "a=1b=2" is a valid toml file but after not.  "a=1 b=2" is valid though, and that maybe should be another discussion whether it should or not.

src/Toml/Parser/Value.hs
jiegillet: OK.

jiegillet:
@willbasky, you wrote the original tests, what's your take on the question?

jiegillet:
I've decided to remove the part about `lookAhead`, I will open a separate issue about it, because it's more general than being about leading zeros. 
I've also reverted the changes to the tests, except the ones relevant to leading zeros.
This is now a very straightforward change to the integer parser.

src/Toml/Parser/Value.hs
```diff
@@ -148,22 +146,20 @@ tableNameP = between (text "[") (text "]") keyP
...
+decimalP = zero <|> more
   where
-    decimalStringP   = fst <$> match (some digitChar >> many _digitsP)
-    _digitsP         = try (char '_') >> some digitChar
-    mkInteger        = textToInt . stripUnderscores
-    textToInt        = fst . fromRight (error "Underscore parser has a bug") . TR.decimal
-    stripUnderscores = Text.filter (/= '_')
+    zero = 0 <$ char '0'
+    more = check =<< readMaybe . concat <$> sepBy1 (some digitChar) (char '_')
+    check = maybe (fail "Not an integer") return
```
ChShersh: @jiegillet Could you benchmark this PR before your PR and after your PR and paste the results here? I wonder what is the difference in performance.

jiegillet: I benchmarked a toml file consisting only of `ints = [0, 1, 2, 120, 9000, -5, +12, -100, 500, +876]
`.
Current implementation:
```
benchmarked Parse/tomland
time                 53.71 μs   (43.82 μs .. 61.27 μs)
                     0.884 R²   (0.835 R² .. 0.949 R²)
mean                 41.11 μs   (39.23 μs .. 43.75 μs)
std dev              7.469 μs   (4.910 μs .. 10.92 μs)
variance introduced by outliers: 84% (severely inflated)
```

After change 

```
benchmarked Parse/tomland
time                 54.02 μs   (50.34 μs .. 59.31 μs)
                     0.969 R²   (0.943 R² .. 0.998 R²)
mean                 53.35 μs   (52.56 μs .. 54.87 μs)
std dev              4.032 μs   (2.875 μs .. 5.893 μs)
variance introduced by outliers: 50% (moderately inflated)
```
Doesn't look very different

jiegillet: Thanks for teaching me how to benchmark @willbasky

jiegillet: Although now I'm thinking that changing a parser potentially changes the behavior of the whole thing, so here is another benchmark of the original toml file + the `ints = [...]` from above.
Before change:
```
benchmarked Parse/tomland
time                 287.0 μs   (274.1 μs .. 299.8 μs)
                     0.986 R²   (0.978 R² .. 0.992 R²)
mean                 299.9 μs   (293.9 μs .. 307.5 μs)
std dev              21.87 μs   (17.97 μs .. 27.69 μs)
variance introduced by outliers: 46% (moderately inflated)
```

After change:
```
benchmarked Parse/tomland
time                 364.9 μs   (338.1 μs .. 390.5 μs)
                     0.973 R²   (0.961 R² .. 0.983 R²)
mean                 319.4 μs   (311.8 μs .. 331.4 μs)
std dev              31.78 μs   (26.91 μs .. 38.29 μs)
variance introduced by outliers: 60% (severely inflated)
```
That's a bigger change...

vrom911: Oh, that's quite a lot I guess. Seems like this feature is costly. How do you guys think?

ChShersh: Yeah, I personally think that it's not worth it to forbid leading zeroes for such performance cost. It's like, not an addition of some new feature, it's more work and complication of parser for not so great benefit. I can understand why TOML specification forbids leading zeroes. But I'm not sure that we should support every bit of TOML spec. Performance is probably not a problem, it's still fast and TOML parsing is supposed to be run only once at the program start. I'm mostly worried about code obfuscation. I already don't understand some places of parser code :disappointed:

jiegillet: (@ChShersh you will cry when you see the parser for arrays of tables then :p )
It's a 6% decrease in speed. I'm quite surprise it's that much. Maybe I can play around more, it's my first time using benchmarks.
I thought my implementation was less obfuscated than the previous one, to be honest (if you forget about `zero`, it's 2 lines instead of 5).

ChShersh: @jiegillet I guess we can live with 6% performance decrease for array of tables :slightly_smiling_face: This is a very important feature. But 20% for not allowing leading zeroes... Way to much! :slightly_smiling_face: And I think that parser for array of tables should be manageable, this part we can handle :+1:

jiegillet: What if, instead of having a 6% increase, I had a 4% _decrease_ with zero more complexity to the code, would you merge the feature then? :)

ChShersh: @jiegillet This looks proimising! :smirk: But be careful in your benchmarks. Sometimes you see that the results _were inflated_. This means that the numbers are not precise and accurate enough.

jiegillet:
I did another proper benchmark after my last commit.
Current master:
```
time                 287.4 μs   (274.3 μs .. 303.4 μs)
                     0.989 R²   (0.984 R² .. 0.996 R²)
mean                 291.0 μs   (286.4 μs .. 295.1 μs)
std dev              14.58 μs   (11.97 μs .. 18.00 μs)
variance introduced by outliers: 29% (moderately inflated)
```

This PR:
```
benchmarked Parse/tomland
time                 282.7 μs   (270.1 μs .. 294.3 μs)
                     0.992 R²   (0.990 R² .. 0.997 R²)
mean                 283.5 μs   (279.0 μs .. 288.8 μs)
std dev              15.61 μs   (11.67 μs .. 21.23 μs)
variance introduced by outliers: 32% (moderately inflated)
```

It's a 3% speedup, with similar moderately inflated variances. 3% speedup doesn't mean much, but at least I'm not slowing things down anymore.

Issue #130, opened 2018-11-06
Add `TomlCodec`s for combinators to `Toml.Bi.Combinators` from `Toml.BiMap` module


ChShersh:
A lot of bidirectional maps were added to the `tomland`. But they miss corresponding codecs.

jiegillet:
I'll take this one.

Issue #119, opened 2018-10-25
Add property-based tests for `BiMap`


ChShersh:
I would like to have `hedgehog` roundript tests for our `BiMap` because now they contain non-trivial logic.

jiegillet:
I've been working on this, and I have an interesting problem. 
My test `tripping x (forward bimap) (join . fmap (backward bimap))` for `_Double` fails, because apparently
```
ghci> 0/0 == (0/0 :: Double)
False
```
What to do? :)

ChShersh:
@jiegillet Trick with `isNan` is used for `Eq` instance for `Value` as well:

* https://github.com/kowainik/tomland/blob/f7ac2946505ad154ac087044ee30051b1399684b/src/Toml/Type/Value.hs#L127-L135

I guess it's okay to violate semantic for `NaN` in our case.

PR #136, opened 2018-11-08
[#130] Added combinators

jiegillet:
#130 

- I replaced the implementation of `arrayOf ` by simply using `match`. It works, but it seems too simple, am I missing anything?
- Because I've removed the big implementation of `arrayOf`, `matchValueBackward` wasn't needed, so I removed it as well, it felt like it only existed for that purpose. I can bring it back in if needed.
- I'm not sure if we need to keep `dimapNum`, the type casting logic feels like it should stay in `BiMap.hs`
- I re-grouped the exports of `BiMap.hs`
- All tests pass, but we should really add more (related to #119) 
- I really think we should rename `_Array` and `arrayOf` as `_List` and `listOf` since the name reflects the Haskell type (just like with `_Word`), not the way data is stored in the `Anything`

### DIFF
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -59,7 +59,7 @@ import Toml.Bi.Code (DecodeException (..), Env, St, TomlCodec)
 import Toml.Bi.Monad (Codec (..), dimap)
 import Toml.BiMap (BiMap (..), _Array, _Bool, _Double,
                    _Integer, _String, _Text, _ZonedTime, _LocalTime, _Day,
-                   _TimeOfDay, _Int, _Word, _Natural, _Float, _Show,
+                   _TimeOfDay, _Int, _Word, _Natural, _Float, _Read,
                    _ByteString, _LByteString, _Set, _IntSet, _HashSet,
                    _NonEmpty)
 import Toml.Parser (ParseException (..))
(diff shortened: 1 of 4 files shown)


src/Toml/Bi/Combinators.hs
```diff
@@ -162,28 +200,22 @@ day = match _Day
...
-                    Nothing   -> throwError $ TypeMismatch key (typeName @a) (valueType x)
-                    Just vals -> pure vals
-            Just _ -> throwError $ TypeMismatch key (typeName @a) TArray
-
-    output :: [a] -> St [a]
-    output a = do
-        anyVal <- MaybeT $ pure $ forward (_Array bimap) a
-        a <$ modify (\(TOML vals tables) -> TOML (HashMap.insert key anyVal vals) tables)
+arrayOf :: Typeable a => BiMap a AnyValue -> Key -> TomlCodec [a]
+arrayOf bimap = match (_Array bimap)
```
willbasky: Is `match` so universal it replaces big piece of code above (`arrayOf`)?

jiegillet: Looks that way! If you looks at the implementations, they are very very similar. Turns out, basically identical :)

willbasky:
I wonder will renew functions work well for `Benchmark.Tomland`?

jiegillet:
Let's test!

src/Toml/Bi/Combinators.hs
```diff
@@ -150,6 +179,15 @@ text = match _Text
 string :: Key -> TomlCodec String
 string = match _String
 
+codecShow :: (Show a, Read a, Typeable a) => Key -> TomlCodec a
+codecShow = match _Show
```
ChShersh: I think that this codec should be named just `show`. This modules is supposed to be exported qualified so people will use `Toml.show` which is better than `Toml.codeShow`

src/Toml/Bi/Combinators.hs
```diff
@@ -150,6 +179,15 @@ text = match _Text
...
+
+bytestring :: Key -> TomlCodec ByteString
+bytestring = match _ByteString
```
ChShersh: Let's rename it to `byteString`

src/Toml/Bi/Combinators.hs
```diff
@@ -150,6 +179,15 @@ text = match _Text
...
+
+lBytestring :: Key -> TomlCodec BL.ByteString
+lBytestring = match _LByteString
```
ChShersh: Same here: `lByteString`

src/Toml/Bi/Combinators.hs
ChShersh: Yes, the ultimate goal is to move all logic regarding conversion between Haskell types and TOML values inside `Toml.BiMap` module and almost all codecs should be just `match _Smth`. In some future we will replace `Maybe` in `BiMap` with `Either` to give better error messages (and that's why `arrayOf` converter will have worse error message after this PR. But that's okay for now, I guess.

src/Toml/Bi/Combinators.hs
```diff
@@ -162,28 +200,22 @@ day = match _Day
...
-
-    output :: [a] -> St [a]
-    output a = do
-        anyVal <- MaybeT $ pure $ forward (_Array bimap) a
-        a <$ modify (\(TOML vals tables) -> TOML (HashMap.insert key anyVal vals) tables)
+arrayOf :: Typeable a => BiMap a AnyValue -> Key -> TomlCodec [a]
+arrayOf bimap = match (_Array bimap)
+
+setOf :: (Typeable a, Ord a) => BiMap a AnyValue -> Key -> TomlCodec (Set a)
+setOf bimap = match (_Set bimap)
```
ChShersh: Let's call all such converters as `arraySetOf` to explicitly tell that they will use use arrays from TOML. Once we have array of tables support, we can then introduce `setOf` and `listOf` and others.

src/Toml/Bi/Combinators.hs
```diff
@@ -162,28 +200,22 @@ day = match _Day
...
+
+intSet :: Key -> TomlCodec IntSet
+intSet = match _IntSet
+
+hashSetOf :: (Typeable a, Hashable a, Eq a) => BiMap a AnyValue -> Key -> TomlCodec (HashSet a)
+hashSetOf bimap = match (_HashSet bimap)
```
ChShersh: Same can be done for other codes
```suggestion
hashSetOf = match . _HashSet
```

src/Toml/BiMap.hs
```diff
@@ -13,9 +13,14 @@ module Toml.BiMap
        , prism
 
          -- * Helpers for BiMap and AnyValue
-       , matchValueBackward
```
ChShersh: Yep, this is a hack needed only for `arrayOf`. I'm okay with removing it.

src/Toml/Bi/Combinators.hs
ChShersh: Also, I'm starting to think that maybe it's better to call it `read` instead of `show`? We use both, but I guess `read` is more explicit. What do you think?

src/Toml/Bi/Combinators.hs
```diff
@@ -150,6 +173,15 @@ text = match _Text
 string :: Key -> TomlCodec String
 string = match _String
 
+read :: (Show a, Read a, Typeable a) => Key -> TomlCodec a
+read = match _Show
```
ChShersh: Let's also rename `_Show` to `_Read` to have consistent naming scheme

src/Toml/Bi/Combinators.hs
```diff
@@ -162,28 +194,22 @@ day = match _Day
...
-            Just _ -> throwError $ TypeMismatch key (typeName @a) TArray
-
-    output :: [a] -> St [a]
-    output a = do
-        anyVal <- MaybeT $ pure $ forward (_Array bimap) a
-        a <$ modify (\(TOML vals tables) -> TOML (HashMap.insert key anyVal vals) tables)
+arrayOf :: Typeable a => BiMap a AnyValue -> Key -> TomlCodec [a]
+arrayOf = match . _Array
+
+arraySetOf :: (Typeable a, Ord a) => BiMap a AnyValue -> Key -> TomlCodec (Set a)
```
ChShersh: Could you also please add Haddock documentation to all function you've added? For `array*Of` you can just copy-paste docs for `arrayOf` and replace `list` with your data structure.

PR #146, opened 2018-11-13
[#129] Allow underscore in floats

### LINKED ISSUE #129: Accept underscores in floats
From the specifications:

> Similar to integers, you may use underscores to enhance readability. Each underscore must be surrounded by at least one digit.
> `flt8 = 224_617.445_991_228`

It probably involves rewriting a variant of `Text.Megaparsec.Char.Lexer.float`

jiegillet:
Resolves #129 

### :white_check_mark: Check list
- [x] CHANGELOG
- [x] New/fixed features work as expected (Added unit tests).
- [x] There are no warnings during compilation.
- [x] `hlint .` output is: _No Hints_ .
- [x] The code is formatted with the `stylish-haskell` tool 
      using `stylish-haskell.yaml` file in the repository.
- [x] The code style of the files you changed is preserved.
- [ ] Commit messages are in the proper format.
      Start the first line of the commit with the issue number in square parentheses.

### DIFF
--- a/src/Toml/Parser/Value.hs
+++ b/src/Toml/Parser/Value.hs
@@ -16,11 +16,13 @@ import Control.Applicative.Combinators (between, count, manyTill, option, option
                                         sepEndBy, skipMany)
 
 import Data.Char (chr, isControl)
+import Data.Either (fromRight)
 import Data.Fixed (Pico)
 import Data.Semigroup ((<>))
 import Data.Text (Text)
 import Data.Time (LocalTime (..), ZonedTime (..), fromGregorianValid, makeTimeOfDayValid,
                   minutesToTimeZone)
+import Text.Read (readMaybe)
 import Text.Read (readMaybe)
 
 import Toml.Parser.Core (Parser, alphaNumChar, anySingle, binary, char, digitChar, eol,
@@ -171,15 +173,13 @@ doubleP = lexeme (signed sc (num <|> inf <|> nan)) <?> "double"
     nan = 0 / 0 <$ string "nan"
 
 floatP :: Parser Double
-floatP = read . concat <$> sequence [ digits, expo <|> dot ]
+floatP = check =<< readMaybe . concat <$> sequence [ digits, expo <|> dot ]
   where
-    digits :: Parser String
+    check = maybe (fail "Not a float") return
+
+    digits, dot, expo :: Parser String
     digits = concat <$> sepBy1 (some digitChar) (char '_')
-
-    dot :: Parser String
     dot = concat <$> sequence [pure <$> char '.', digits, option "" expo]
-
-    expo :: Parser String
     expo = concat <$> sequence
              [ pure <$> (char 'e' <|> char 'E')
              , pure <$> option '+' (char '+' <|> char '-')
--- a/src/Toml/Parser/Value.hs
+++ b/src/Toml/Parser/Value.hs
@@ -23,7 +23,7 @@ import Data.Time (LocalTime (..), ZonedTime (..), fromGregorianValid, makeTimeOf
                   minutesToTimeZone)
 import Text.Read (readMaybe)
 
-import Toml.Parser.Core (Parser, alphaNumChar, anySingle, binary, char, digitChar, eol, float,
+import Toml.Parser.Core (Parser, alphaNumChar, anySingle, binary, char, digitChar, eol,
                          hexDigitChar, hexadecimal, lexeme, octal, satisfy, sc, signed, space,
                          string, tab, text, try, (<?>))
 import Toml.PrefixTree (Key (..), Piece (..))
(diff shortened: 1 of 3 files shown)


jiegillet:
I used `read`in this PR and in #128. I realized I probably shouldn't. I'll change that.

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -113,22 +113,24 @@ spec_Parser = do
...
             $ do
                   parseDouble "5e+22" 5e+22
                   parseDouble "1e6"   1e6
                   parseDouble "-2E-2" (-2E-2)
-        it
-                "can parse a number which consists of an integral, a fractional, and an exponent part"
+        it "can parse a number which consists of an integral, a fractional, and an exponent part"
             $ parseDouble "6.626e-34" 6.626e-34
+        it "can parse a number with underscores"
+            $ do
```
ChShersh: I think it's better to write `$ do` on the same line as `it " "` like:

```haskell
it "..." $ do
```
No need to have extra line

src/Toml/Parser/Value.hs
```diff
@@ -170,10 +171,23 @@ doubleP :: Parser Double
...
     num, inf, nan :: Parser Double
-    num = float
+    num = floatP
     inf = 1 / 0 <$ string "inf"
     nan = 0 / 0 <$ string "nan"
 
+floatP :: Parser Double
+floatP = check =<< readMaybe . concat <$> sequence [ digits, expo <|> dot ]
+  where
+    check = maybe (fail "Not a float") return
```
vrom911: Am I understand this correctly that this `fail` will be handled by `<|>`? I just don't know how it's implemented in `megaparsec`, does it use `fail` as well?

jiegillet: I'm not too sure, because that situation should never happen. It's not a `megaparsec` function.

vrom911: Yes, I mean `fail` is somehow implemented for `Parser` from `megaparsec`. And we are using this `floatP` with `<|>` operator, so, that's why I'm asking, how it's going to be handled

jiegillet: I just tried this:
```
ghci> Toml.Parser.Core.parse (fail "whoops" <|> boolP) "" "true"
Right True
```
so it's all good :)

src/Toml/Parser/Value.hs
```diff
@@ -170,10 +171,23 @@ doubleP :: Parser Double
...
+
+    digits, dot, expo :: Parser String
```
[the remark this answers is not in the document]
jiegillet: We are using `Text.Megaparsec.Char` which uses `Char` everywhere, so we are stuck with that. unless we want to use `Text.Megaparsec.Byte`.

src/Toml/Parser/Value.hs
ChShersh: Yeah, let's not do `error`, `fail` is ok for now 👍

src/Toml/Parser/Value.hs
```diff
@@ -170,10 +171,23 @@ doubleP :: Parser Double
 doubleP = lexeme (signed sc (num <|> inf <|> nan)) <?> "double"
   where
     num, inf, nan :: Parser Double
-    num = float
+    num = floatP
     inf = 1 / 0 <$ string "inf"
     nan = 0 / 0 <$ string "nan"
 
+floatP :: Parser Double
+floatP = check =<< readMaybe . concat <$> sequence [ digits, expo <|> dot ]
```
ChShersh: I think that this parser should go `Toml.Parser.Core`. `Toml.Parser.Value` is a very big module, so let's try to not make it bigger 🙂

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -113,22 +113,23 @@ spec_Parser = do
...
                   parseDouble "1e6"   1e6
                   parseDouble "-2E-2" (-2E-2)
-        it
-                "can parse a number which consists of an integral, a fractional, and an exponent part"
+        it "can parse a number which consists of an integral, a fractional, and an exponent part"
             $ parseDouble "6.626e-34" 6.626e-34
+        it "can parse a number with underscores" $ do
+                  parseDouble "5e+2_2" 5e+22
+                  parseDouble "1.1_1e6"   1.11e6
+                  parseDouble "-2_2.21_9E-0_2" (-22.219E-2)
```
ChShersh: Tests are really nice!

src/Toml/Parser/Value.hs
```diff
@@ -170,10 +171,23 @@ doubleP :: Parser Double
...
+  where
+    check = maybe (fail "Not a float") return
+
+    digits, dot, expo :: Parser String
+    digits = concat <$> sepBy1 (some digitChar) (char '_')
+    dot = concat <$> sequence [pure <$> char '.', digits, option "" expo]
+    expo = concat <$> sequence
```
ChShersh: I feel like you can implement helper function for `concat <$> sequence`. And then maybe we can notice that such function already exist.

src/Toml/Parser/Value.hs
jiegillet: I need to point out that `dateTimeP` has 2 `error` in it. Also `decimalP` has one, but I'm getting rid of it in #128

src/Toml/Parser/Value.hs
jiegillet: I tried, but it was not a good fit. I had to add so many import to `Toml.Parser.Core` and it didn't fit with any of the functions there... Maybe a refactor of larger scale might work, moving all the functions that are not exported by `Toml.Parser.Value` to `Core`.

Issue #149, opened 2018-11-15
Warnings from GHC 8.4.4


jiegillet:
Since switching to GHC 8.4.4, there's been a bunch of new warnings, especially `warning: [-Wpartial-fields] Use of partial record field selector` because of data structures like
```haskell
data KeysDiff
    = Equal
    | FstIsPref { diff :: !Key}
```
since you can apply `diff` to `FstIsPref` but not `Equal`.

According to the organization style guide: "Records for data types with multiple constructors are forbidden.", so we should remove the record syntax.

Also a warning about module `warning [-Wmissing-export-lists] The export item ‘module Main’ is missing an export list` in `Playground.hs`

PR #150, opened 2018-11-15
[#149] Removing record syntax from PrefixTree

jiegillet:
Addressing #149.

### DIFF
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -1,4 +1,6 @@
-module Main where
+{-# OPTIONS -Wno-unused-top-binds #-}
+
+module Main (main) where
 
 import Control.Applicative ((<|>))
 import Control.Arrow ((>>>))
--- a/src/Toml/PrefixTree.hs
+++ b/src/Toml/PrefixTree.hs
@@ -96,26 +96,30 @@ type PrefixMap a = HashMap Piece (PrefixTree a)
 
 -- | Data structure to represent table tree for @toml@.
 data PrefixTree a
-      -- | last piece of a key, and value at that point.
-    = Leaf !Key !a
-      -- | greatest common prefix, optional value at that point and values of suffixes.
-    | Branch !Prefix !(Maybe a) !(PrefixMap a)
+    = Leaf             -- ^ End of a key.
+        !Key           -- ^ End piece of the key.
+        !a             -- ^ Value at the end.
+    | Branch           -- ^ Values along pieces of a key.
+        !Prefix        -- ^ Greatest common key prefix.
+        !(Maybe a)     -- ^ Possible value at that point.
+        !(PrefixMap a) -- ^ Values at suffixes of the prefix.
     deriving (Show, Eq, NFData, Generic)
 
 instance Semigroup (PrefixTree a) where
     a <> b = foldl' (\tree (k, v) -> insertT k v tree) a (toListT b)
 
+-- | Data structure to compare keys.
 data KeysDiff
-      -- | Keys are equal
-    = Equal
-      -- | Keys don't have any common part.
-    | NoPrefix
-      -- | The first key is the prefix for the second one.
-    | FstIsPref !Key
-      -- | The second key is the prefix for the first one.
-    | SndIsPref !Key
-      -- | Key have same prefix. Common prefix, rest of first key, rest of scond key.
-    | Diff !Key  !Key  !Key
+    = Equal      -- ^ Keys are equal
+    | NoPrefix   -- ^ Keys don't have any common part.
+    | FstIsPref  -- ^ The first key is the prefix of the second one.
+        !Key     -- ^ Rest of the second key.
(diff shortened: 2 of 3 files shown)


examples/Playground.hs
```diff
@@ -1,4 +1,6 @@
-module Main where
+{-# OPTIONS -Wno-unused-top-binds #-}
```
ChShersh: Is there's something unused? I guess this module shouldn't contain unused definitions and everything in this module should be used unless it's redundant.

src/Toml/PrefixTree.hs
```diff
@@ -96,10 +96,9 @@ type PrefixMap a = HashMap Piece (PrefixTree a)
 -- | Data structure to represent table tree for @toml@.
 data PrefixTree a
     = Leaf !Key !a
```
ChShersh: Let's also comment fields of  `Leaf` constructor while we're on it. Actually, we can make `tomland` quite an excellent library if we add Haddock comments as much as possible just like you did in `KeysDiff` data type.

src/Toml/PrefixTree.hs
```diff
@@ -111,14 +110,13 @@ data KeysDiff
       -- | Keys don't have any common part.
     | NoPrefix
       -- | The first key is the prefix for the second one.
-    | FstIsPref { diff :: !Key}
+    | FstIsPref !Key
```
ChShersh: Let's add comment to this field.

examples/Playground.hs
jiegillet: Yeah it's pretty weird, I think it's an actual GHC bug. 
It's because of `newtype N = N Text`, the compiler says the `N` _constructor_ is never used, even though the `N` _type_ is used once in `data Test = Test { ..., testN  :: N ... }`. I couldn't find a cleaner way to go around this.

src/Toml/PrefixTree.hs
```diff
@@ -95,11 +96,10 @@ type PrefixMap a = HashMap Piece (PrefixTree a)
...
 -- | Data structure to represent table tree for @toml@.
 data PrefixTree a
+      -- | last piece of a key, and value at that point.
     = Leaf !Key !a
-    | Branch { bCommonPref :: !Prefix         -- ^ greatest common prefix
-             , bVal        :: !(Maybe a)      -- ^ value by key = prefix
-             , bPrefixMap  :: !(PrefixMap a)  -- ^ suffixes of prefix
-             }
+      -- | greatest common prefix, optional value at that point and values of suffixes.
+    | Branch !Prefix !(Maybe a) !(PrefixMap a)
```
vrom911: I guess (I'm not sure though) that it's possible to write comments to fields without records, something like:
```haskell
   | Branch 
             !Prefix         -- ^ greatest common prefix
             !(Maybe a)      -- ^ value by key = prefix
             !(PrefixMap a)  -- ^ suffixes of prefix
```

vrom911: Heh, that's sad, I'll check if they have an issue for this feature request.

examples/Playground.hs
ChShersh: @jiegillet You can submit bug to GHC with minimal reproducible example :slightly_smiling_face:
Feel free to ask any questions about your example or bug-reporting process!

src/Toml/PrefixTree.hs
ChShersh: @vrom911 @jiegillet There's a way to document fields of constructor. But you also need to document constructor itself. See this PR:

* https://github.com/haskell/haddock/pull/709

examples/Playground.hs
jiegillet: I've never done that before, exciting, I'll try.
Seriously, I've learned so much and found out so many things in the last 6 weeks, I cannot thank you enough!

jiegillet: Seems there's already a ticket [here](https://ghc.haskell.org/trac/ghc/ticket/10347). It doesn't look like it's considered high priority though...

ChShersh: @jiegillet You can leave the comment under that issue as well. Usually, the more comments under bug the more chances it will be closed :slightly_smiling_face:

jiegillet: Done! 
In the mean time, do I leave the `{-# OPTIONS -Wno-unused-top-binds #-}`?

ChShersh: @jiegillet Yes, let's keep this option.

src/Toml/PrefixTree.hs
jiegillet: It's only available in Haddock 2.21.0, but stack is using 2.20.0. Is that something we can specify in one of the config files?

ChShersh: @jiegillet It's okay if this is not supported by `stack`. I'm uploading packages to Hackage using `cabal new-*` commands, so we shouldn't expect any problems with this :+1:

PR #151, opened 2018-11-17
 [#134] Pretty printer for table arrays

### LINKED ISSUE #134: Road to Arrays of Tables - Pretty Printing
Implement pretty-printing for array of tables and write some golden tests

jiegillet:
Resolves #134

### DIFF
--- a/src/Toml/Printer.hs
+++ b/src/Toml/Printer.hs
@@ -11,6 +11,7 @@ module Toml.Printer
        ) where
 
 import Data.HashMap.Strict (HashMap)
+import Data.List.NonEmpty (NonEmpty)
 import Data.Monoid ((<>), mconcat)
 import Data.Text (Text)
 import Data.Time (formatTime, defaultTimeLocale, ZonedTime)
@@ -77,7 +78,7 @@ pretty = prettyOptions defaultOptions
 
 {- | Converts 'TOML' type into 'Text' using provided 'PrintOptions' -}
 prettyOptions :: PrintOptions -> TOML -> Text
-prettyOptions options = Text.drop 1 . prettyTomlInd options 0 ""
+prettyOptions options = flip Text.snoc '\n' . Text.drop 1 . prettyTomlInd options 0 ""
 
 -- | Converts 'TOML' into 'Text' with the given indent.
 prettyTomlInd :: PrintOptions -- ^ Printing options
@@ -86,8 +87,12 @@ prettyTomlInd :: PrintOptions -- ^ Printing options
               -> TOML         -- ^ Given 'TOML'
               -> Text         -- ^ Pretty result
 prettyTomlInd options i prefix TOML{..} =
-    prettyKeyValue options i tomlPairs <> "\n"
-    <> prettyTables options i prefix tomlTables
+  Text.intercalate "\n" $ filter (not . Text.null)
+    [ prettyKeyValue options i tomlPairs
+    , prettyTables options i prefix tomlTables
+    , prettyTableArrays options i prefix tomlTableArrays
+    ]
+
 
 -- | Returns pretty formatted  key-value pairs of the 'TOML'.
 prettyKeyValue :: PrintOptions -> Int -> HashMap Key AnyValue -> Text
@@ -136,7 +141,7 @@ prettyKeyValue options i =
 -- | Returns pretty formatted tables section of the 'TOML'.
 prettyTables :: PrintOptions -> Int -> Text -> PrefixMap TOML -> Text
 prettyTables options i pref =
-    Text.concat . map (prettyTable . snd) . orderWith options . HashMap.toList
+    Text.intercalate "\n" . map (prettyTable . snd) . orderWith options . HashMap.toList
   where
     prettyTable :: PrefixTree TOML -> Text
     prettyTable (Leaf k toml) =
@@ -167,3 +172,26 @@ prettyTables options i pref =
 
 prettyKey :: Key -> Text
 prettyKey (Key k) = Text.intercalate "." $ map unPiece (NonEmpty.toList k)
+
+prettyTableArrays :: PrintOptions -> Int -> Text -> HashMap Key (NonEmpty TOML) -> Text
+prettyTableArrays options i pref =
+    Text.intercalate "\n" . map arrText . orderWith options . HashMap.toList
+  where
+    arrText :: (Key, NonEmpty TOML) -> Text
+    arrText (k, ne) =
+      let name = getPref k
+          render toml = mconcat
+            [ tabWith options i
+            , prettyTableArrayName name
+            , prettyTomlInd options (succ i) name toml
+            ]
+      in Text.intercalate "\n" $ map render $ NonEmpty.toList ne
+
+    -- Adds next part of the table name to the accumulator.
+    getPref :: Key -> Text
+    getPref k = case pref of
+        "" -> prettyKey k
+        _  -> pref <> "." <> prettyKey k
+
+    prettyTableArrayName :: Text -> Text
+    prettyTableArrayName n = "[[" <> n <> "]]"
--- a/src/Toml/Printer.hs
+++ b/src/Toml/Printer.hs
@@ -7,15 +7,15 @@ module Toml.Printer
        , defaultOptions
        , pretty
        , prettyOptions
-       , prettyTomlInd
+       , prettyKey
        ) where
 
 import Data.HashMap.Strict (HashMap)
+import Data.List (sortOn, splitAt)
 import Data.List.NonEmpty (NonEmpty)
-import Data.Monoid ((<>), mconcat)
+import Data.Monoid ((<>))
 import Data.Text (Text)
-import Data.Time (formatTime, defaultTimeLocale, ZonedTime)
-import Data.List (splitAt, sortOn)
+import Data.Time (ZonedTime, defaultTimeLocale, formatTime)
 
 import Toml.PrefixTree (Key (..), Piece (..), PrefixMap, PrefixTree (..))
 import Toml.Type (AnyValue (..), DateTime (..), TOML (..), Value (..))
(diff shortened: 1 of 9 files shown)


src/Toml/Printer.hs
```diff
@@ -77,7 +78,7 @@ pretty = prettyOptions defaultOptions
 
 {- | Converts 'TOML' type into 'Text' using provided 'PrintOptions' -}
 prettyOptions :: PrintOptions -> TOML -> Text
-prettyOptions options = Text.drop 1 . prettyTomlInd options 0 ""
+prettyOptions options = flip Text.snoc '\n' . Text.drop 1 . prettyTomlInd options 0 ""
```
ChShersh: The addition of `\n` looks strange. Maybe you can just use `unlines` instead of `Text.intercalate "\n" ` in `prettyTomlInd`?

src/Toml/Printer.hs
```diff
@@ -86,8 +87,12 @@ prettyTomlInd :: PrintOptions -- ^ Printing options
               -> TOML         -- ^ Given 'TOML'
               -> Text         -- ^ Pretty result
 prettyTomlInd options i prefix TOML{..} =
-    prettyKeyValue options i tomlPairs <> "\n"
-    <> prettyTables options i prefix tomlTables
+  Text.intercalate "\n" $ filter (not . Text.null)
```
ChShersh: I think you can join this line with previous one

src/Toml/Printer.hs
```diff
@@ -86,8 +87,12 @@ prettyTomlInd :: PrintOptions -- ^ Printing options
...
+    [ prettyKeyValue options i tomlPairs
+    , prettyTables options i prefix tomlTables
+    , prettyTableArrays options i prefix tomlTableArrays
```
ChShersh: Just a minor cosmetic suggestion but I guess that aligning `options`, `i` and other variables under each other will makes this section cleaner 👍

src/Toml/Printer.hs
```diff
@@ -136,7 +141,7 @@ prettyKeyValue options i =
 -- | Returns pretty formatted tables section of the 'TOML'.
 prettyTables :: PrintOptions -> Int -> Text -> PrefixMap TOML -> Text
 prettyTables options i pref =
-    Text.concat . map (prettyTable . snd) . orderWith options . HashMap.toList
+    Text.intercalate "\n" . map (prettyTable . snd) . orderWith options . HashMap.toList
```
ChShersh: This also looks strange. Do you have any idea why `Text.concat` previously was enough and now this is `intercalate`?

src/Toml/Printer.hs
```diff
@@ -167,3 +172,26 @@ prettyTables options i pref =
 
 prettyKey :: Key -> Text
 prettyKey (Key k) = Text.intercalate "." $ map unPiece (NonEmpty.toList k)
+
+prettyTableArrays :: PrintOptions -> Int -> Text -> HashMap Key (NonEmpty TOML) -> Text
+prettyTableArrays options i pref =
+    Text.intercalate "\n" . map arrText . orderWith options . HashMap.toList
```
ChShersh: The following piece of code appears 3 times already:

```haskell
map somethingSomething . orderWith options . HashMap.toList
```

Looks like it can be moved into separate function.

src/Toml/Printer.hs
```diff
@@ -167,3 +172,26 @@ prettyTables options i pref =
...
+            , prettyTableArrayName name
+            , prettyTomlInd options (succ i) name toml
+            ]
+      in Text.intercalate "\n" $ map render $ NonEmpty.toList ne
+
+    -- Adds next part of the table name to the accumulator.
+    getPref :: Key -> Text
+    getPref k = case pref of
+        "" -> prettyKey k
+        _  -> pref <> "." <> prettyKey k
```
ChShersh: The same is in tables. Probably makes sense to move into separate function as well to remove code duplication.

test/golden/pretty_indented_only.golden
```diff
@@ -12,3 +12,15 @@ b = "bb"
 [qux.doo]
     egg = "?"
     spam = "!"
+
+[[deeper]]
+    green = true
```
ChShersh: Tests are really good! Thank you!

src/Toml/Printer.hs
jiegillet: I can't believe I forgot about `unlines`...

jiegillet: Mmh, that might not work so easily... Everything breaks when I use `unlines` because `prettyTomlInd` is called recursively and `unlines` adds a whole bunch of unwanted newlines... `intercalate` has a much nicer behavior, but I'll play around a bit more.

jiegillet:
New version, I've shuffled things around because the `unlines` things was too unsettling. It's still a bit complicated but it's not a trivial function.

I ran into a small annoyance, we have a name clash between `Toml.Edsl.table` and `Toml.Bi.Combinators.table`. Both are useful and both names fit, but the clash is not great for the users. Is it worth opening an issue about it?

ChShersh:
@jiegillet 

> we have a name clash between Toml.Edsl.table and Toml.Bi.Combinators.table. Both are useful and both names fit, but the clash is not great for the users. Is it worth opening an issue about it?

I think this is okay. I guess you only need `Edsl` in some test cases. You always can import modules qualified, like:

```haskell
import qualified Toml.Edsl as Edsl
```

But you also can open issue to discuss better naming :slightly_smiling_face:

src/Toml/Printer.hs
```diff
@@ -134,36 +127,54 @@ prettyKeyValue options i =
...
-    prettyTable :: PrefixTree TOML -> Text
+    prettyTable :: PrefixTree TOML -> [Text]
     prettyTable (Leaf k toml) =
-        let name = getPref k in mconcat
-            [ tabWith options i
-            , prettyTableName name
-            , prettyTomlInd options (succ i) name toml ]
+        let name = addPrefix k pref
+        in "": tabWith options i <> prettyTableName name :
+             dropWhile (== "") (prettyTomlInd options (succ i) name toml)
```
ChShersh: This addition of `""` and `dropWhile (== "")` looks very weird... At this point I think that using `T.intercalate` is better than adding strange logic :thinking:

src/Toml/Printer.hs
```diff
@@ -134,36 +127,54 @@ prettyKeyValue options i =
...
+
     prettyTable (Branch k mToml prefMap) =
-        let name  = getPref k
+        let name  = addPrefix k pref
             nextI = succ i
```
ChShersh: Also, let's replace `succ` with `+ 1` in all places here.

src/Toml/Printer.hs
```diff
@@ -134,36 +127,54 @@ prettyKeyValue options i =
...
-            , toml
-            , prettyTables options nextI name prefMap ]
-
-    -- Adds next part of the table name to the accumulator.
-    getPref :: Key -> Text
-    getPref k = case pref of
-        "" -> prettyKey k
-        _  -> pref <> "." <> prettyKey k
+        in "": tabWith options i <> prettyTableName name :
+             dropWhile (== "") (toml ++ prettyTables options nextI name prefMap)
```
ChShersh: Yeah, I don't think it will be easy to understand what's going on here in future

jiegillet: It's not too hard to explain: a `""` will print as an empty line, we always want an empty line before a table/array name, but never after.

ChShersh: In that case adding some comment to this piece of code might be useful 👍

src/Toml/Printer.hs
```diff
@@ -75,24 +64,28 @@ title = "TOML Example"
...
-    <> prettyTables options i prefix tomlTables
+              -> [Text]         -- ^ Pretty result
+prettyTomlInd options i prefix TOML{..} = concat
+    [ prettyKeyValue    options i tomlPairs
+    , prettyTables      options i prefix tomlTables
+    , prettyTableArrays options i prefix tomlTableArrays
+    ]
+
+prettyKey :: Key -> Text
+prettyKey = Text.intercalate "." . map unPiece . NonEmpty.toList . unKey
```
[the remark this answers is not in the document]
jiegillet: It was just a few lines so I added it to this PR. I can revert that last commit if you want.

src/Toml/Printer.hs
```diff
@@ -134,36 +126,60 @@ prettyKeyValue options i =
...
-    prettyTable :: PrefixTree TOML -> Text
+    prettyTable :: PrefixTree TOML -> [Text]
     prettyTable (Leaf k toml) =
-        let name = getPref k in mconcat
-            [ tabWith options i
-            , prettyTableName name
-            , prettyTomlInd options (succ i) name toml ]
+        let name = addPrefix k pref
+        -- Each "" results in an empty line, inserted above table names
+        in "": tabWith options i <> prettyTableName name :
```
ChShersh: Having comments is a good idea here :+1:

test/Test/Toml/PrefixTree/Property.hs
```diff
@@ -5,7 +5,19 @@ import Hedgehog (forAll, (===))
...
+import qualified Data.Text as Text
 import qualified Toml.PrefixTree as Prefix
+import qualified Toml.Printer as Printer
+
+----------------------------------------------------------------------------
+-- Key printing and parsing
+----------------------------------------------------------------------------
+test_KeyPrinting :: PropertyTest
+test_KeyPrinting = prop "Key printing: fromString . prettyKey == id" $ do
+  key <- forAll genKey
```
ChShersh:
```suggestion
    key <- forAll genKey
```

test/Test/Toml/PrefixTree/Property.hs
```diff
@@ -5,7 +5,19 @@ import Hedgehog (forAll, (===))
...
+  (fromString . Text.unpack) (Printer.prettyKey key) === key
```
ChShersh:
```suggestion
    (fromString . Text.unpack) (Printer.prettyKey key) === key
```

ChShersh: Also, it's better to use `tripping` combinator from `hedgehog` library in such cases because it produces better error messages:

* https://hackage.haskell.org/package/hedgehog-0.6.1/docs/Hedgehog.html#v:tripping

jiegillet:
I didn't know about the force push not sending notifications, I'll leave a comment next time!

ChShersh:
@jiegillet I think that the PR is ready to go. But I see some conflicts...

vrom911:
@jiegillet CI fails for older GHC versions in benchmarks 🤔 Probably some CPP pragma needed (not sure).

ChShersh:
I think it's better to fix CI error under separate PR, because this will unblock a lot of PR's and all existing PR's are already big enough...

jiegillet:
Yeah, I also looked into it and it looks like it has nothing to do with my code...

ChShersh:
@jiegillet Try rebasing on `master`. CI passes there. I guess it will be easier for you to do the following:

1. Squash all commits in this branch into single commit.
2. Rebase on master.
3. Force push
4. Notify about your changes.

I reviewed this PR long time ago, so there's no harm in losing commit history, I will rereview again, but most likely I will just merge, because, as I remember, everything is good there :+1:

jiegillet:
Alright, everything works fine now, all tests pass.
The only change I've done with this commit is removing one unused import in `Toml.Parser.Value` to get rid of a warning.

Issue #137, opened 2018-11-10
Replace `Maybe` with `Either ...smth...` in `BiMap`


ChShersh:
This is needed to give better error messages.

ChShersh:
@willbasky Yes, that's right. I assume that after refactoring this should look something like:

```haskell
data BiMap e a b = BiMap
    { forward :: a -> Either e b
    , backward :: b -> Either e a
    }
```

And then all current `BiMap`s will have the following type:

```haskell
type BiToml = BiMap SomeTomlError
```

PR #153, opened 2018-11-18
[#137] Replace Maybe with Either in BiMap

willbasky:
Resolves [#137](https://github.com/kowainik/tomland/issues/137)

Next task:
- [x] Implement more exceptions for `BiMapError`

### DIFF
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -55,12 +55,12 @@ import Numeric.Natural (Natural)
 
 import Toml.Bi.Code (DecodeException (..), Env, St, TomlCodec)
 import Toml.Bi.Monad (Codec (..))
-import Toml.BiMap (BiMap (..), _Array, _Bool, _ByteString, _Day, _Double, _Float, _HashSet, _Int,
+import Toml.BiMap (BiMap (..), BiMapError (..), _Array, _Bool, _ByteString, _Day, _Double, _Float, _HashSet, _Int,
                    _IntSet, _Integer, _LByteString, _LocalTime, _Natural, _NonEmpty, _Read, _Set,
                    _String, _Text, _TimeOfDay, _Word, _ZonedTime)
 import Toml.Parser (ParseException (..))
 import Toml.PrefixTree (Key)
-import Toml.Type (AnyValue (..), BiMapError (..), TOML (..), insertKeyAnyVal, insertTable,
+import Toml.Type (AnyValue (..), TOML (..), insertKeyAnyVal, insertTable,
                   valueType)
 
 import Prelude hiding (read)
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -55,7 +55,7 @@ import Numeric.Natural (Natural)
 
 import Toml.Bi.Code (DecodeException (..), Env, St, TomlCodec)
 import Toml.Bi.Monad (Codec (..))
-import Toml.BiMap (BiMap (..), BiMapError (..), _Array, _Bool, _ByteString, _Day, _Double, _Float, _HashSet, _Int,
+import Toml.BiMap (BiMap (..), TomlBiMap, _Array, _Bool, _ByteString, _Day, _Double, _Float, _HashSet, _Int,
                    _IntSet, _Integer, _LByteString, _LocalTime, _Natural, _NonEmpty, _Read, _Set,
                    _String, _Text, _TimeOfDay, _Word, _ZonedTime)
 import Toml.Parser (ParseException (..))
@@ -79,7 +79,7 @@ typeName = Text.pack $ show $ typeRep $ Proxy @a
 
 {- | General function to create bidirectional converters for values.
 -}
-match :: forall a . Typeable a => BiMap BiMapError a AnyValue -> Key -> TomlCodec a
+match :: forall a . Typeable a => TomlBiMap a AnyValue -> Key -> TomlCodec a
 match BiMap{..} key = Codec input output
   where
     input :: Env a
@@ -203,12 +203,12 @@ timeOfDay = match _TimeOfDay
 
 -- | Parser for list of values. Takes converter for single value and
 -- returns a list of values.
-arrayOf :: Typeable a => BiMap BiMapError a AnyValue -> Key -> TomlCodec [a]
+arrayOf :: Typeable a => TomlBiMap a AnyValue -> Key -> TomlCodec [a]
 arrayOf = match . _Array
 
 -- | Parser for sets. Takes converter for single value and
 -- returns a set of values.
-arraySetOf :: (Typeable a, Ord a) => BiMap BiMapError a AnyValue -> Key -> TomlCodec (Set a)
+arraySetOf :: (Typeable a, Ord a) => TomlBiMap a AnyValue -> Key -> TomlCodec (Set a)
 arraySetOf = match . _Set
 
 -- | Parser for sets of ints. Takes converter for single value and
@@ -220,14 +220,14 @@ arrayIntSet = match _IntSet
 -- returns a set of hashable values.
 arrayHashSetOf
     :: (Typeable a, Hashable a, Eq a)
-    => BiMap BiMapError a AnyValue
+    => TomlBiMap a AnyValue
     -> Key
     -> TomlCodec (HashSet a)
 arrayHashSetOf = match . _HashSet
 
 -- | Parser for non- empty lists of values. Takes converter for single value and
 -- returns a non-empty list of values.
-arrayNonEmptyOf :: Typeable a => BiMap BiMapError a AnyValue -> Key -> TomlCodec (NonEmpty a)
+arrayNonEmptyOf :: Typeable a => TomlBiMap a AnyValue -> Key -> TomlCodec (NonEmpty a)
 arrayNonEmptyOf = match . _NonEmpty
 
 -- | Parser for tables. Use it when when you have nested objects.
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -60,7 +60,8 @@ import Toml.BiMap (BiMap (..), _Array, _Bool, _ByteString, _Day, _Double, _Float
                    _String, _Text, _TimeOfDay, _Word, _ZonedTime)
 import Toml.Parser (ParseException (..))
 import Toml.PrefixTree (Key)
-import Toml.Type (AnyValue (..), TOML (..), insertKeyAnyVal, insertTable, valueType)
+import Toml.Type (AnyValue (..), BiMapError (..), TOML (..), insertKeyAnyVal, insertTable,
+                  valueType)
 
 import Prelude hiding (read)
 
@@ -78,7 +79,7 @@ typeName = Text.pack $ show $ typeRep $ Proxy @a
 
 {- | General function to create bidirectional converters for values.
 -}
-match :: forall a . Typeable a => BiMap a AnyValue -> Key -> TomlCodec a
+match :: forall a . Typeable a => BiMap BiMapError a AnyValue -> Key -> TomlCodec a
 match BiMap{..} key = Codec input output
   where
     input :: Env a
@@ -87,12 +88,12 @@ match BiMap{..} key = Codec input output
         case mVal of
             Nothing -> throwError $ KeyNotFound key
             Just anyVal@(AnyValue val) -> case backward anyVal of
-                Just v  -> pure v
-                Nothing -> throwError $ TypeMismatch key (typeName @a) (valueType val)
+                Right v -> pure v
+                Left _  -> throwError $ TypeMismatch key (typeName @a) (valueType val)
 
     output :: a -> St a
     output a = do
-        anyVal <- MaybeT $ pure $ forward a
+        anyVal <- MaybeT $ pure $ either (const Nothing) Just $ forward a
         a <$ modify (insertKeyAnyVal key anyVal)
 
 {- | Almost same as 'dimap'. Useful when you want to have fields like this
@@ -202,12 +203,12 @@ timeOfDay = match _TimeOfDay
 
 -- | Parser for list of values. Takes converter for single value and
 -- returns a list of values.
-arrayOf :: Typeable a => BiMap a AnyValue -> Key -> TomlCodec [a]
+arrayOf :: Typeable a => BiMap BiMapError a AnyValue -> Key -> TomlCodec [a]
 arrayOf = match . _Array
 
 -- | Parser for sets. Takes converter for single value and
 -- returns a set of values.
-arraySetOf :: (Typeable a, Ord a) => BiMap a AnyValue -> Key -> TomlCodec (Set a)
+arraySetOf :: (Typeable a, Ord a) => BiMap BiMapError a AnyValue -> Key -> TomlCodec (Set a)
 arraySetOf = match . _Set
 
 -- | Parser for sets of ints. Takes converter for single value and
@@ -217,12 +218,16 @@ arrayIntSet = match _IntSet
 
 -- | Parser for hash sets. Takes converter for single hashable value and
 -- returns a set of hashable values.
-arrayHashSetOf :: (Typeable a, Hashable a, Eq a) => BiMap a AnyValue -> Key -> TomlCodec (HashSet a)
+arrayHashSetOf
+    :: (Typeable a, Hashable a, Eq a)
+    => BiMap BiMapError a AnyValue
+    -> Key
+    -> TomlCodec (HashSet a)
 arrayHashSetOf = match . _HashSet
 
 -- | Parser for non- empty lists of values. Takes converter for single value and
 -- returns a non-empty list of values.
-arrayNonEmptyOf :: Typeable a => BiMap a AnyValue -> Key -> TomlCodec (NonEmpty a)
+arrayNonEmptyOf :: Typeable a => BiMap BiMapError a AnyValue -> Key -> TomlCodec (NonEmpty a)
 arrayNonEmptyOf = match . _NonEmpty
 
 -- | Parser for tables. Use it when when you have nested objects.
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -48,7 +48,6 @@ import Data.Semigroup ((<>))
 import Data.Set (Set)
 import Data.Text (Text)
 import Data.Time (Day, LocalTime, TimeOfDay, ZonedTime)
-import Data.Typeable (Typeable)
 import Data.Word (Word)
 import Numeric.Natural (Natural)
 
@@ -59,8 +58,7 @@ import Toml.BiMap (BiMap (..), TomlBiMap, _Array, _Bool, _ByteString, _Day, _Dou
                    _Read, _Set, _String, _Text, _TimeOfDay, _Word, _ZonedTime)
 import Toml.Parser (ParseException (..))
 import Toml.PrefixTree (Key)
-import Toml.Type (AnyValue (..), TOML (..), Value, insertKeyAnyVal, insertTable, typeName,
-                  valueType)
+import Toml.Type (AnyValue (..), TOML (..), Value, insertKeyAnyVal, insertTable)
 
 import Prelude hiding (read)
 
@@ -74,7 +72,7 @@ import qualified Toml.PrefixTree as Prefix
 
 {- | General function to create bidirectional converters for values.
 -}
-match :: forall a . Typeable a => TomlBiMap a AnyValue -> Key -> TomlCodec a
+match :: forall a . TomlBiMap a AnyValue -> Key -> TomlCodec a
 match BiMap{..} key = Codec input output
   where
     input :: Env a
@@ -82,9 +80,9 @@ match BiMap{..} key = Codec input output
         mVal <- asks $ HashMap.lookup key . tomlPairs
         case mVal of
             Nothing -> throwError $ KeyNotFound key
-            Just anyVal@(AnyValue val) -> case backward anyVal of
-                Right v -> pure v
-                Left _  -> throwError $ TypeMismatch key (typeName @a) (valueType val)
+            Just anyVal@(AnyValue _) -> case backward anyVal of
+                Right v  -> pure v
+                Left err -> throwError $ BiMapError err
 
     output :: a -> St a
     output a = do
@@ -169,7 +167,7 @@ string :: Key -> TomlCodec String
 string = match _String
 
 -- | Parser for values with a `Read` and `Show` instance.
-read :: (Show a, Read a, Typeable a) => Key -> TomlCodec a
+read :: (Show a, Read a) => Key -> TomlCodec a
 read = match _Read
 
 -- | Parser for byte vectors values as strict bytestring.
@@ -198,12 +196,12 @@ timeOfDay = match _TimeOfDay
 
 -- | Parser for list of values. Takes converter for single value and
 -- returns a list of values.
-arrayOf :: Typeable a => TomlBiMap a AnyValue -> Key -> TomlCodec [a]
+arrayOf :: TomlBiMap a AnyValue -> Key -> TomlCodec [a]
 arrayOf = match . _Array
 
 -- | Parser for sets. Takes converter for single value and
 -- returns a set of values.
-arraySetOf :: (Typeable a, Ord a) => TomlBiMap a AnyValue -> Key -> TomlCodec (Set a)
+arraySetOf :: Ord a => TomlBiMap a AnyValue -> Key -> TomlCodec (Set a)
 arraySetOf = match . _Set
 
 -- | Parser for sets of ints. Takes converter for single value and
@@ -214,7 +212,7 @@ arrayIntSet = match _IntSet
 -- | Parser for hash sets. Takes converter for single hashable value and
 -- returns a set of hashable values.
 arrayHashSetOf
-    :: (Typeable a, Hashable a, Eq a)
+    :: (Hashable a, Eq a)
     => TomlBiMap a AnyValue
     -> Key
     -> TomlCodec (HashSet a)
@@ -222,7 +220,7 @@ arrayHashSetOf = match . _HashSet
 
 -- | Parser for non- empty lists of values. Takes converter for single value and
 -- returns a non-empty list of values.
-arrayNonEmptyOf :: (a ~ Value t, Typeable a) => TomlBiMap a AnyValue -> Key -> TomlCodec (NonEmpty a)
+arrayNonEmptyOf :: (a ~ Value t) => TomlBiMap a AnyValue -> Key -> TomlCodec (NonEmpty a)
 arrayNonEmptyOf = match . _NonEmpty
 
 -- | Parser for tables. Use it when when you have nested objects.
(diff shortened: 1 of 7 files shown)


src/Toml/Type/AnyValue.hs
```diff
@@ -82,8 +83,14 @@ reifyAnyValues _ []                 = Right []
...
-    Left _     -> Nothing
-    Right vals -> Just $ Array (x : vals)
+    Left _     -> Left BiMapError
+    Right vals -> Right $ Array (x : vals)
+
+----------------------------------------------------------------------------
+-- BiMap error type
+----------------------------------------------------------------------------
+
+data BiMapError = BiMapError
```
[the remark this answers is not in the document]
jiegillet: But all these function should return `Maybe, not `Either`, because if they fail, we know why, it's because of a type mismatch, we can use that knowledge in `BiMap.hs`.

willbasky:
@jiegillet To make a part of function with `Either` and another one with `Maybe` we need two `BiMapEither` and `BiMapMaybe` with all supported functions.

src/Toml/Type/AnyValue.hs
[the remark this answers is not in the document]
willbasky: But if `BiMap` are `Either` implemented then it ask all functions used `BiMap` to be `Either` or to add convertor from `Maybe` to `Either`.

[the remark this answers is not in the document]
jiegillet: Look at how `matchText` is implemented:
```haskell
matchText :: Value t -> Either BiMapError Text
matchText (Text s) = Right s
matchText _        = Left BiMapError
```
It can either succeed or fail, but it doesn't even know why it failed because we are pattern matching with `_`. A junction that fails without knowing why should always return `Nothing`.  Then in `BiMap.hs` you can do something fancy like
```
_Text :: BiMap BiMapError Text AnyValue
_Text = mkAnyValueBiMap 
    (\a@(AnyValue v) -> either (Left BiMapErrorTypeMistatch Text (valueType v)) Right $ matchText a) 
    Text
```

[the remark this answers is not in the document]
willbasky: Thank you.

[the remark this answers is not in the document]
willbasky: It is one-more-operation variant.
Proper is 
```
_Text :: BiMap BiMapError Text AnyValue
_Text = mkAnyValueBiMap
    (\a -> maybe (Left BiMapError) Right $ matchText a) Text
```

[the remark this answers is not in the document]
jiegillet: Right. But maybe it's best to change `mkAnyValueBiMap` instead?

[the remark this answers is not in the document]
willbasky: OMG yes!

src/Toml/BiMap.hs
```diff
@@ -57,79 +58,84 @@ import Control.Arrow ((>>>))
...
-prism :: (field -> object) -> (object -> Maybe field) -> BiMap object field
-prism review preview = BiMap preview (Just . review)
+prism :: (field -> object) -> (object -> Either error field) -> BiMap error object field
+prism review preview = BiMap preview (Right . review)
+
+----------------------------------------------------------------------------
+-- BiMap error type
+----------------------------------------------------------------------------
+
+data BiMapError = BiMapError
```
ChShersh: I guess that the most important part of this PR is the type of this error. I can see the following cases:

1. Name of the data type constructor is not what we expected.
2. Name of the `Value` constructor is not what we expected.
3. Arbitrary textual reason for other cases.

Maybe something else. But I guess we can settle to this 3 constructors for now.

src/Toml/BiMap.hs
```diff
@@ -57,79 +58,84 @@ import Control.Arrow ((>>>))
...
 
 ----------------------------------------------------------------------------
 -- General purpose bimaps
 ----------------------------------------------------------------------------
 
 -- | Bimap for 'Either' and its left type
-_Left :: BiMap (Either l r) l
-_Left = prism Left (either Just (const Nothing))
+_Left :: BiMap BiMapError (Either l r) l
```
ChShersh: Let's introduce some proper names 🙂 

1. Replace `BiMapError` with `TomlBiMapError`.
2. Introduce type alias: `type TomlBiMap = BiMap TomlBiMapError`

src/Toml/BiMap.hs
```diff
@@ -138,158 +170,169 @@ _Just = prism Just id
...
+                -> TomlBiMap a AnyValue
 mkAnyValueBiMap matchValue toValue =
-    BiMap (Just . AnyValue . toValue) (\(AnyValue value) -> matchValue value)
+    BiMap (Right . AnyValue . toValue) (\(AnyValue value) -> maybeToBiEither $ matchValue value)
 
 -- | Creates bimap for 'Text' to 'AnyValue' with custom functions
-_TextBy :: (a -> Text) -> (Text -> Maybe a) -> BiMap a AnyValue
+_TextBy :: (a -> Text) -> (Text -> Maybe a) -> TomlBiMap a AnyValue
 _TextBy toText parseText =
   mkAnyValueBiMap (matchText >=> parseText) (Text . toText)
```
ChShersh: This is a bit difficult case. `TextBy` actually can fail for multiple reasons:

1. Wrong value: expected `Text` but got something different.
2. `parseText` function failed to parse due to some reason.

I think better solution here is the following:

1. Change type of `parseText` argument from `Text -> Maybe a` to `Text -> Either Text a` where `Left` constructor will hold error messages. In most cases people should pass `readEither` function as an argument for `_TextBy`.
2. Change type of `matchValue` inside `mkAnyValueBiMap` from `forall t . Value t -> Maybe a` to `forall t . Value t -> Either TomlBiMapError a`

So this will require you to change types of all functions like:

```haskell
matchText :: Value t -> Maybe Text
```

to

```haskell
matchText :: Value t -> Either MatchError Text
```

Where `MatchError` looks like `TypeMismatchError`:

* https://github.com/kowainik/tomland/blob/4a47b14ac1b8122a6e4658a549da1a27492f7e85/src/Toml/Type/Value.hs#L209-L213

But instead of storing types, it should store type of expected value and actual value. And, I guess, instead of having two fields in `WrongValue` constructor, you can have just one field called `MatchError`.

src/Toml/BiMap.hs
```diff
@@ -138,158 +170,169 @@ _Just = prism Just id
...
+_NaturalInteger :: TomlBiMap Natural Integer
+_NaturalInteger = BiMap (Right . toInteger) eitherInteger
   where
-    maybeInteger :: Integer -> Maybe Natural
-    maybeInteger n
-      | n < 0     = Nothing
-      | otherwise = Just (fromIntegral n)
+    eitherInteger :: Integer -> Either TomlBiMapError Natural
+    eitherInteger n
+      | n < 0     = Left $ ArbitraryError "Value is below zero."
```
ChShersh: The usage of `ArbitraryError` here is correct 👍

src/Toml/BiMap.hs
```diff
@@ -138,158 +170,169 @@ _Just = prism Just id
...
+_BoundedInteger = BiMap (Right . toInteger) eitherBounded
   where
-    maybeBounded :: forall a. (Integral a, Bounded a) => Integer -> Maybe a
-    maybeBounded n
-      | n < toInteger (minBound :: a) = Nothing
-      | n > toInteger (maxBound :: a) = Nothing
-      | otherwise                     = Just (fromIntegral n)
+    eitherBounded :: forall a. (Integral a, Bounded a) => Integer -> Either TomlBiMapError a
+    eitherBounded n
+      | n < toInteger (minBound :: a) = Left $ ArbitraryError "Value is less minBound"
```
ChShersh: It's better to patch error message here to tell what is the actual value and what is the value of `minBound`

src/Toml/BiMap.hs
```diff
@@ -138,158 +170,169 @@ _Just = prism Just id
...
+      | n > toInteger (maxBound :: a) = Left $ ArbitraryError "Value is more maxBound"
```
ChShersh: Same here.

src/Toml/BiMap.hs
```diff
@@ -138,158 +170,169 @@ _Just = prism Just id
...
 -- | Helper bimap for 'Text' and lazy 'ByteString'
-_LByteStringText :: BiMap BL.ByteString Text
-_LByteStringText = prism (TL.encodeUtf8 . TL.fromStrict) maybeText
+_LByteStringText :: TomlBiMap BL.ByteString Text
+_LByteStringText = prism (TL.encodeUtf8 . TL.fromStrict) eitherText
   where
-    maybeText :: BL.ByteString -> Maybe Text
-    maybeText = either (const Nothing) (Just . TL.toStrict) . TL.decodeUtf8'
+    eitherText :: BL.ByteString -> Either TomlBiMapError Text
+    eitherText = either (const $ Left $ ArbitraryError "Decoding failed") (Right . TL.toStrict) . TL.decodeUtf8'
```
ChShersh: I think it's better to preserve the exact error of `decodeUtf8'` function here to tell what exactly went wrong.

src/Toml/BiMap.hs
```diff
@@ -138,158 +170,169 @@ _Just = prism Just id
...
+    , backward = \(AnyValue val) -> maybeToBiEither $ matchArray (eitherToMaybe . backward elementBimap) val
     }
 
 -- | Takes a bimap of a value and returns a bimap of a non-empty list of values
 -- and 'Anything' as an array. Usually used with 'nonEmptyOf' combinator.
-_NonEmpty :: BiMap a AnyValue -> BiMap (NE.NonEmpty a) AnyValue
-_NonEmpty bimap = BiMap (Just . NE.toList) NE.nonEmpty >>> _Array bimap
+_NonEmpty :: TomlBiMap a AnyValue -> TomlBiMap (NE.NonEmpty a) AnyValue
+_NonEmpty bimap = BiMap (Right . NE.toList)
+    (maybeToBiEither . NE.nonEmpty) >>> _Array bimap
```
ChShersh: Here should be also custom error saying that empty list was given but non-empty list expected

src/Toml/BiMap.hs
```diff
@@ -51,85 +53,115 @@ module Toml.BiMap
...
-prism review preview = BiMap preview (Just . review)
+prism :: (field -> object) -> (object -> Either error field) -> BiMap error object field
+prism review preview = BiMap preview (Right . review)
+
+----------------------------------------------------------------------------
+-- BiMap error types
+----------------------------------------------------------------------------
+
+data TomlBiMapError
+    = TrivialBiError
```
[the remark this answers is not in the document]
willbasky: Do you mean something like?
```
data TomlBiMapError
    = BiMapError (TomlBiMapError -> DecodeException)
    | ...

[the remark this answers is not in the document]
Comment (ChShersh) [reply src/Toml/BiMap.hs]: @willbasky I was thinking about adding extra constructor to `DecodeException` data type, basically:

```haskell
data DecodeException
    = TrivialError
    | BiMapError TomlBiMapError
    ...
```

src/Toml/Type/AnyValue.hs
```diff
@@ -43,36 +55,36 @@ instance NFData AnyValue where
 ----------------------------------------------------------------------------
 
 -- | Extract 'Bool' from 'Value'.
-matchBool :: Value t -> Maybe Bool
-matchBool (Bool b) = Just b
-matchBool _        = Nothing
+matchBool :: Value t -> Either TomlBiMapError Bool
```
ChShersh: Instead of `TomlBiMapError` these `match*` functions should return `MatchError`

src/Toml/Type/AnyValue.hs
```diff
@@ -82,8 +94,42 @@ reifyAnyValues _ []                 = Right []
...
-    Right vals -> Just $ Array (x : vals)
+    Left _     -> leftWrongValue x
+    Right vals -> Right $ Array (x : vals)
+
+---------------------------------------------------------------------------
+-- Value mismatch error and TomlBiMapError
+---------------------------------------------------------------------------
+
+data MatchError = MatchError
+    { valueExpected :: Text
```
ChShersh: Let's use `TValue` here instead of `Text` to restrict our types even further. In this field there couldn't be any other value except `TValue`

test/Test/Toml/Gen.hs
```diff
@@ -216,10 +215,11 @@ noneArrayList =
...
 
 genArrayFrom :: MonadGen m => m AnyValue -> m (Value 'TArray)
-genArrayFrom noneArray
-    = fromMaybe (error "Error in genArrayFrom")
-    . toMArray
-  <$> Gen.list (Range.constant 0 5) noneArray
+genArrayFrom noneArray = do
+    eVal <- toMArray <$> Gen.list (Range.constant 0 5) noneArray
+    case eVal of
+        Left err -> error $ Text.unpack $ prettyBiMapError err
```
ChShersh: 👍

src/Toml/BiMap.hs
```diff
@@ -51,245 +55,270 @@ module Toml.BiMap
...
+-- BiMap error types
+----------------------------------------------------------------------------
+
+-- | Abstraction on BiMap with error.
+type TomlBiMap = BiMap TomlBiMapError
+
+prettyBiMapError :: TomlBiMapError -> Text
+prettyBiMapError = \case
+    WrongConstructor expect actual ->
+        "Invalid type constrictor. Expected " <> expect <> " Actual constructor: " <> actual
```
ChShersh: Let's write prettier error messages, like:

```haskell
WrongConstructor expected actual -> unlines
    [ "Invalid constructor"
    , "  * Expected: " <> expected
    , "  * Actual:   " <> actual
    ]
```

src/Toml/BiMap.hs
```diff
@@ -51,245 +55,270 @@ module Toml.BiMap
...
 -- General purpose bimaps
 ----------------------------------------------------------------------------
 
 -- | Bimap for 'Either' and its left type
-_Left :: BiMap (Either l r) l
-_Left = prism Left (either Just (const Nothing))
+_Left :: (Show l, Show r) => TomlBiMap (Either l r) l
+_Left = prism Left $ \case
+    Left l -> Right l
+    x -> Left $ WrongConstructor "Left" (toTxt x)
```
ChShersh: Let's create and expose `wrongConstructor` helper function to allow users to write such matches easily.

src/Toml/BiMap.hs
```diff
@@ -51,245 +55,270 @@ module Toml.BiMap
...
 
 ----------------------------------------------------------------------------
 -- General purpose bimaps
 ----------------------------------------------------------------------------
 
 -- | Bimap for 'Either' and its left type
-_Left :: BiMap (Either l r) l
-_Left = prism Left (either Just (const Nothing))
+_Left :: (Show l, Show r) => TomlBiMap (Either l r) l
+_Left = prism Left $ \case
```
ChShersh: Ooo, the usage of LambdaCase here is soooo nice!

src/Toml/BiMap.hs
```diff
@@ -51,245 +55,270 @@ module Toml.BiMap
...
+                -> TomlBiMap a AnyValue
+mkAnyValueBiMap matchValue toValue = BiMap (Right . AnyValue . toValue)
+    (\(AnyValue value) -> either (\_ -> leftWrongValue value) Right $ matchValue value)
 
 -- | Creates bimap for 'Text' to 'AnyValue' with custom functions
-_TextBy :: (a -> Text) -> (Text -> Maybe a) -> BiMap a AnyValue
-_TextBy toText parseText =
-  mkAnyValueBiMap (matchText >=> parseText) (Text . toText)
+_TextBy :: (a -> Text) -> (Text -> Either TomlBiMapError a) -> TomlBiMap a AnyValue
+_TextBy toText parseText = mkAnyValueBiMap (matchText >=> parseText) (Text . toText)
```
ChShersh: I don't expect users to pass parsing function from `Text` to `Either TomlBiMapError` here 🙂 

It's more convenient if this function has type: `Text -> Either Text a)` and you can wrap this text into `TomlBiMapError`

src/Toml/BiMap.hs
```diff
@@ -51,245 +55,270 @@ module Toml.BiMap
...
-_String :: BiMap String AnyValue
+_String :: TomlBiMap String AnyValue
 _String = _StringText >>> _Text
 
 -- | Helper bimap for 'String' and types with 'Read' and 'Show' instances.
-_ReadString :: (Show a, Read a) => BiMap a String
-_ReadString = BiMap (Just . show) readMaybe
+_ReadString :: (Show a, Read a) => TomlBiMap a String
+_ReadString = BiMap (Right . show)
+    (\value -> either (\_ -> leftWrongValue . Text $ toTxt value) Right . readEither $ value)
```
ChShersh: It's good idea to keep error returned by `readEither` function and tell this error.

src/Toml/Bi/Combinators.hs
```diff
@@ -217,12 +213,16 @@ arrayIntSet = match _IntSet
...
+    :: (Typeable a, Hashable a, Eq a)
+    => TomlBiMap a AnyValue
+    -> Key
+    -> TomlCodec (HashSet a)
 arrayHashSetOf = match . _HashSet
 
 -- | Parser for non- empty lists of values. Takes converter for single value and
 -- returns a non-empty list of values.
-arrayNonEmptyOf :: Typeable a => BiMap a AnyValue -> Key -> TomlCodec (NonEmpty a)
+arrayNonEmptyOf :: (a ~ Value t, Typeable a) => TomlBiMap a AnyValue -> Key -> TomlCodec (NonEmpty a)
```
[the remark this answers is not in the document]
willbasky: Otherwise an error returns:
```
Couldn't match type ‘a’ with ‘Value t0’
    arising from a use of ‘_NonEmpty’
...
```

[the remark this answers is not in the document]
ChShersh: Hmm, that's strange :thinking: I will try to investigate this.

src/Toml/BiMap.hs
jiegillet: I think even expecting a `Text -> Either Text a` from users is a lot. Can we keep it `Text -> Maybe a`?

ChShersh: Well, the idea here is that in most cases users should pass here some function that outputs proper error in case of failed parsing. At least it should be `readEither`. If you have function of type `Text -> Maybe a` then you can easily convert it to `Text -> Either Text a` by providing some error message for `Nothing`. It should be users responsibility to tell proper error message.

src/Toml/Bi/Combinators.hs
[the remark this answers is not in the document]
ChShersh: @willbasky Okay, I've checked. This is because of type of `_NonEmpty`:

```haskell
_NonEmpty :: a ~ Value t => TomlBiMap a AnyValue -> TomlBiMap (NE.NonEmpty a) AnyValue
```

The `a ~ Value t` constraint is wrong here. The implementation of `_NonEmpty` should be look kike `_Array`

src/Toml/BiMap.hs
```diff
@@ -51,245 +56,314 @@ module Toml.BiMap
...
-_String :: BiMap String AnyValue
+_String :: TomlBiMap String AnyValue
 _String = _StringText >>> _Text
 
 -- | Helper bimap for 'String' and types with 'Read' and 'Show' instances.
-_ReadString :: (Show a, Read a) => BiMap a String
-_ReadString = BiMap (Just . show) readMaybe
+_ReadString :: (Show a, Read a) => TomlBiMap a String
+_ReadString = BiMap (Right . show)
+    (\value -> either (\_ -> leftWrongValue . Text $ tShow value) Right . readEither $ value)
```
ChShersh: Here should be used `ArbitraryError` constructor, like this:

```haskell
_ReadString = BiMap (Right . show) (first (ArbitraryError . T.pack) . readEither)
```

src/Toml/BiMap.hs
```diff
@@ -51,245 +56,314 @@ module Toml.BiMap
...
+        , "  * Expected: " <> expected
+        , "  * Actual:   " <> actual
+        ]
+    WrongValue (MatchError expect actual) -> T.unlines
+        [ "Invalid value constructor. Expected "
+        , tShow expect
+        , " Actual value: "
+        , tShow actual
+        ]
+    ArbitraryError text  -> "Invalid value: " <> text
```
ChShersh: Just `Text` is enough here
```suggestion
    ArbitraryError text  -> text
```

src/Toml/BiMap.hs
```diff
@@ -51,245 +56,314 @@ module Toml.BiMap
...
-    maybeBounded n
-      | n < toInteger (minBound :: a) = Nothing
-      | n > toInteger (maxBound :: a) = Nothing
-      | otherwise                     = Just (fromIntegral n)
+    eitherBounded :: forall a. (Integral a, Bounded a, Show a) => Integer -> Either TomlBiMapError a
+    eitherBounded n
+      | n < toInteger (minBound :: a) =
+            Left $ ArbitraryError $ "Value " <> tShow n <> " is less minBound" <> tShow (minBound :: a)
+      | n > toInteger (maxBound :: a) =
+            Left $ ArbitraryError $ "Value " <> tShow n <> " is more maxBound" <> tShow (maxBound :: a)
```
ChShersh: Nice error messages here!

src/Toml/BiMap.hs
```diff
@@ -51,245 +56,314 @@ module Toml.BiMap
...
 
 -- | Helper bimap for 'Text' and strict 'ByteString'
-_ByteStringText :: BiMap ByteString Text
-_ByteStringText = prism T.encodeUtf8 maybeText
-  where
-    maybeText :: ByteString -> Maybe Text
-    maybeText = either (const Nothing) Just . T.decodeUtf8'
+_ByteStringText :: TomlBiMap ByteString Text
+_ByteStringText = prism T.encodeUtf8
+    (\value -> either (\_ -> leftWrongValue . Text $ tShow value) Right $ T.decodeUtf8' value)
```
ChShersh: It's a good idea to do here the same with `ArbitraryError` as you did in `_LByteStringText`

src/Toml/BiMap.hs
```diff
@@ -51,245 +56,314 @@ module Toml.BiMap
...
     }
 
 -- | Takes a bimap of a value and returns a bimap of a non-empty list of values
 -- and 'Anything' as an array. Usually used with 'nonEmptyOf' combinator.
-_NonEmpty :: BiMap a AnyValue -> BiMap (NE.NonEmpty a) AnyValue
-_NonEmpty bimap = BiMap (Just . NE.toList) NE.nonEmpty >>> _Array bimap
+_NonEmpty :: TomlBiMap a AnyValue -> TomlBiMap (NE.NonEmpty a) AnyValue
+_NonEmpty elementBimap = BiMap
+    { forward = mapM (forward elementBimap) >=>
+        fmap AnyValue . toTomlBiMapError . toMArray . NE.toList
```
ChShersh: I propose to remove some code duplication between this function and `_Array`. You can implement:

```haskell
_NonEmptyArray :: TomlBiMap (NonEmpty a) [a]
```

where you handle error of empty list. And then you can implement:

```haskell
_NonEmpty bimap = _NonEmptyArray >>> _Array bimap
```

src/Toml/Type/AnyValue.hs
```diff
@@ -82,8 +110,23 @@ reifyAnyValues _ []                 = Right []
...
+tShow :: Show a => a -> Text
+tShow = T.pack . show
+
+-- | Expacted value
+typeName :: forall a . Typeable a => TValue
+typeName = read . show . typeRep $ Proxy @a
+
+-- | Left error part of MatchError.
+leftMatchError :: forall (t :: TValue) b . Value t -> Either MatchError b
+leftMatchError = Left . MatchError (typeName @TValue) . AnyValue
```
ChShersh: The same for this function applies as for `leftMatchError` in `BiMap` module.

src/Toml/Type/AnyValue.hs
```diff
@@ -82,8 +110,23 @@ reifyAnyValues _ []                 = Right []
...
+----------------------------------------------------------------------
+-- Useful functions
+----------------------------------------------------------------------
+
+tShow :: Show a => a -> Text
+tShow = T.pack . show
+
+-- | Expacted value
+typeName :: forall a . Typeable a => TValue
+typeName = read . show . typeRep $ Proxy @a
```
ChShersh: This function is kinda unsafe. As I understand, you want to convert type-level `TValue` tag to `Text`. In that case it will be safer to have `forall (t :: TValue)` and to use `readMaybe` returning `<unknown>` on `Nothing`.

src/Toml/Type/AnyValue.hs
```diff
@@ -38,41 +53,54 @@ instance Eq AnyValue where
...
 
 -- | Extract list of elements of type @a@ from array.
-matchArray :: (AnyValue -> Maybe a) -> Value t -> Maybe [a]
+matchArray :: (AnyValue -> Either MatchError a) -> Value t -> Either MatchError [a]
 matchArray matchValue (Array a) = mapM (liftMatch matchValue) a
-matchArray _            _       = Nothing
+matchArray _          value     = leftMatchError value
+
+-- | Extract NonEmpty list of elements of type @a@ from array.
+matchArrayNE :: (AnyValue -> Either MatchError a) -> Value t -> Either MatchError (NE.NonEmpty a)
```
ChShersh: This function shouldn't be in `Type` directory. This module only for basic and primitive stuff. It's better to handle `NonEmpty` error in `BiMap` module.

src/Toml/BiMap.hs
```diff
@@ -51,245 +56,314 @@ module Toml.BiMap
...
+        ]
+    ArbitraryError text  -> "Invalid value: " <> text
+
+-- | Helper to construct WrongConstuctor error.
+wrongConstructor :: Show a => Text -> a -> Either TomlBiMapError b
+wrongConstructor constructor x = Left $ WrongConstructor constructor (tShow x)
+
+-- | Left error part of TomlBiMapError.
+leftWrongValue :: forall (t :: TValue) b . Value t -> Either TomlBiMapError b
+leftWrongValue = Left . WrongValue . MatchError (typeName @TValue) . AnyValue
```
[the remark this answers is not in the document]
willbasky: if `typeName @t`
then `Expected a type, but ‘t’ has kind ‘TValue’`

willbasky:
`toMatchError` stayed untouched.

ChShersh:
@willbasky If you're using this `typeName` only for `TValue`, then it's better to constrain the kind of `a` type variable, like this:

```haskell
typeName :: forall (a :: TValue) . Typeable a => TValue
```

So you will guarantee by this that `typeName` never fails. Though, you still need `error` in the implementation.

Regarding `matchError`: the function should be defined like this:

```haskell
matchError :: forall (t :: TValue) b . Value t -> Either MatchError b
matchError = Left . MatchError (typeName @t) . AnyValue
```

Though, instead of `typeName @t` it's even better to use `valueType` function which is much safer:

https://github.com/kowainik/tomland/blob/8f4009ab2fe2be923dd0b1e7b56e92ddf1f87ee7/src/Toml/Type/Value.hs#L144-L152

In that case you won't even need `typeName` that extracts `TValue` from type. And it looks like `typeName` should return just `Text`. As I can see from current code, `typeName` is used only in one place: to tell the name of the type that has matching error:

https://github.com/kowainik/tomland/blob/fe87ad5ffc29d07afe4a9d4a9e7411d65396d212/src/Toml/Bi/Combinators.hs#L90

So there is slight chance that `typeName` won't be needed anymore after this refactoring.

PR #156, opened 2018-11-20
[#152] Remove mdimap

### LINKED ISSUE #152: Remove mdimap
It is obsolete, having the same use case as `prism`.

jiegillet:
Resolves #152

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,9 @@ The change log is available [on GitHub][2].
   Adding EDSL support for arrays of tables
 * [#149](https://github.com/kowainik/tomland/issues/149):
   Removing records syntax from `PrefixTree`.
-  
+* [#152](https://github.com/kowainik/tomland/issues/152):
+  Removing `mdimap`.
+
 ## 0.5.0 — Nov 12, 2018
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,7 +15,9 @@ The change log is available [on GitHub][2].
   Adding EDSL support for arrays of tables
 * [#149](https://github.com/kowainik/tomland/issues/149):
   Removing records syntax from `PrefixTree`.
-  
+* [#152](https://github.com/kowainik/tomland/issues/152):
+  Removing `mdimap`.
```
ChShersh: This is a breaking change. So let's write it in style of `**Breaking change:**` and `Migration guide`.

Also, could you please update version here from `## 0.6.0` to `## Unreleased: 1.0.0`?

src/Toml/Bi/Combinators.hs
```diff
@@ -95,43 +93,6 @@ match BiMap{..} key = Codec input output
         anyVal <- MaybeT $ pure $ forward a
         a <$ modify (insertKeyAnyVal key anyVal)
 
-{- | Almost same as 'dimap'. Useful when you want to have fields like this
```
ChShersh: 👍

PR #157, opened 2018-11-22
 [#154] Implement Generic instance for BiMap a AnyValue

### LINKED ISSUE #154: Implement codec for `Generic`
I think it would be nice to have something like:

```haskell
generic :: Generic a => TomlCodec a
```

And in future this can be expanded to:

```haskell
genericWithOptions :: Generic a => TomlGenericOptions -> TomlCodec a
```

Will save a lot of boilerplate 😄

jiegillet:
Related to #154.

### :white_check_mark: Check list
- [x] CHANGELOG
- [x] New/fixed features work as expected (No unit tests, to be addressed in #119).
- [x] There are no warnings during compilation.
- [x] `hlint` output on touch files is: _No Hints_ .
- [x] The code is formatted with the `stylish-haskell` tool 
      using `stylish-haskell.yaml` file in the repository.
- [x] The code style of the files you changed is preserved.
- [x] Commit messages are in the proper format.
      Start the first line of the commit with the issue number in square parentheses.

Three days ago I didn't know what `Generics` were, this has been so much fun!!!

Small concerns/questions:
- I didn't add all the possible instances we have at the moment, but it's a matter of copy/paste. Should I add everything (`Natural`, `IntMap`...) or leave users to mimic the source(each instance is 3 trivial lines)?
- I have no doubt that this feature is awesome, but the `.toml` output is not very human friendly, which may work against the core principle of TOML (Tom's Obvious, Minimal Language).
- The docs of `GHC.Generics` say that the "encoding strategy may not be reliable across different versions of GHC", meaning that the TOML decoding may fail for different compilers. I haven't tested for that.


ChShersh:
@jiegillet Huge work! But I think it makes more sense to wait until full support for array of tables before merging this because list is a very common data structure :+1:

jiegillet:
In the current state, this can only encode/decode data structures at the `AnyValue` level, it doesn't know about keys or tables, so it's unrelated to array of table. 
I think I've implemented something a little different than you had in mind.

ChShersh:
@jiegillet Generic typeclasses for `AnyValues` are probably needed as well. The idea I had in mind is that people write something like this:

```haskell
data User = User
    { name :: Text
    , age  :: Int
    } deriving (Generic)

userCodec :: TomlCodec User
userCodec = genericCodec
```

And the last codec is equivalent to:

```haskell
userCodec :: TomlCodec User
userCodec = User
    <$> Toml.text "name" .= name
    <*> Toml.int "age" .= age
```

So just saves some boilerplate

src/Toml/BiMap.hs
```diff
@@ -293,3 +299,148 @@ _HashSet bimap = iso HS.toList HS.fromList >>> _Array bimap
...
+      go b (AnyValue v) = Just $ AnyValue $ Array [Array [Bool b], Array [v]]
+
+  decode' (AnyValue (Array [Array [Bool False], Array [v]])) =
+               L1 <$> decode' (AnyValue v)
+  decode' (AnyValue (Array [Array [Bool  True], Array [v]])) =
+               R1 <$> decode' (AnyValue v)
+  decode' _ = Nothing
+
+-- | Instance for product types.
+instance (GenBiMap' f, GenBiMap' g) => GenBiMap' (f :*: g) where
```
ChShersh: At this point this typeclass looks more like `ToValue/FromValue` as was discussed earlier. After looking closely at the implementation, I don't think there's big value (no pun intended) in such generic instances for `BiMap`. Because instance for `:*:` is a product of types, basically, records. And you shouldn't use `BiMap` with complicated data types anyways.

jiegillet:
I don't disagree with you, I'm not sure a generic `BiMap` is very appropriate... I rushed into it without question anything, learned a lot and had a ton of fun. Now I realize it's not a good feature :)
That being said, I think I've learned enough to implement generic records, we can add that feature whenever appropriate.

ChShersh:
@jiegillet Thank you for your understanding! I will close this PR, probably it will be easier to start from scratch when we get to this issue :slightly_smiling_face:

PR #158, opened 2018-11-30
[#119] Add property-based tests for BiMap

### LINKED ISSUE #119: Add property-based tests for `BiMap`

jiegillet:
Resolves #119.

It's almost my first time playing with properties in hedgehog, let me know if I'm missing something.

I don't know how to test lazy bytestrings.

### DIFF
--- a/test/Test/Toml/BiMap/Property.hs
+++ b/test/Test/Toml/BiMap/Property.hs
@@ -8,6 +8,7 @@ import Hedgehog (Gen, PropertyT, Range, assert, forAll, tripping, (===))
 
 import Data.Time (Day, ZonedTime (..))
 import GHC.Natural (Natural)
+import Test.Tasty (testGroup)
 import Test.Toml.Gen (PropertyTest, prop)
 import Toml.Bi.Map (BiMap (..), TomlBiMap)
 
@@ -27,7 +28,7 @@ range100 :: Range Int
 range100 = Range.constant 0 100
 
 genNatural :: Gen Natural
-genNatural = Gen.integral (Range.constant 0 (2^(65 :: Int)))
+genNatural = fromIntegral <$> Gen.word Range.constantBounded
 
 genInSet :: Gen IS.IntSet
 genInSet = IS.fromList <$> Gen.list range100 (Gen.int Range.constantBounded)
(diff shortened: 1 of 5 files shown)


ChShersh:
@jiegillet Heh, this is a huge change and it conflicts a lot with what @willbasky is doing now. If tests got merged before refactoring `BiMap` to `Either` from `Maybe`, then it looks like tests should be rewritten. So I propose to wait before @willbasky finishes refactoring with `Either` and then we can work on tests.

jiegillet:
Sure, let's wait for the refactoring. It won't be too hard to adapt, the tedious part was gathering all the bimahs.

ChShersh:
@jiegillet Green light for this PR is given :green_heart: You can now rebase on `master` and finish tests with `BiMap` which is based on `Either` instead of `Maybe`

jiegillet:
Rebasing finished. 
It turned out to be quite simple to adapt to the new `BiMap`, I think it's the sign of a refactoring well done. Good job @willbasky!

Should we add tests based on failures? Those would probably not be property-based though.

test/Test/Toml/BiMap/Property.hs
```diff
@@ -0,0 +1,76 @@
...
+    <> prop "Word" (testBiMap B._Word (Gen.word Range.constantBounded))
+    <> prop "Int" (testBiMap B._Int (Gen.int Range.constantBounded))
+    <> prop "Float"
+        (testBiMap B._Float (Gen.float $ Range.constant (-1000000.0) 1000000.0))
+    <> prop "String"
+        (testBiMap B._String (Gen.string (Range.constant 0 100) Gen.unicode))
+    <> prop "Read (Integer)" (testBiMap B._Read G.genInteger )
+    <> prop "Natural"
+        (testBiMap B._Natural (fromIntegral <$> Gen.integral
+                             (Range.constant @Integer 0 (2^(65 :: Integer)))))
```
ChShersh: This generator looks complex enough. I like these tests when they look simple, like:

```haskell
prop "Day" (testBiMap B._Day G.genDay)
```

It would be nice to move this part into separate generator in this module so these test can be written like this:

```haskell
prop "Natural" $ testBiMap B._Natural genNatural
```

So something like `(Gen.word Range.constantBounded)` is already ok, no need to implement `genWord`, but this generator for `Natural` looks big enough to move into separate function.

P.S. Btw, `hedgehog` should contain generator for `Natural`. I can recommend open issue to this library.

test/Test/Toml/BiMap/Property.hs
```diff
@@ -0,0 +1,76 @@
...
+        (testBiMap (B._Set B._Day) (Gen.set (Range.constant 0 100) G.genDay))
+    <> prop "NonEmpty (Day)"
+        (testBiMap (B._NonEmpty B._Day)
+            (Gen.nonEmpty (Range.constant 1 100) G.genDay))
+    <> prop "HashSet (Integer)"
+        (testBiMap (B._HashSet B._Integer) (HS.fromList <$> Gen.list
+                                       (Range.constant 0 100) G.genInteger))
+
+
+-- TODO _LByteString
```
ChShersh: For something like `LByteString`: you can write the following generator:

```haskell
genLByteString :: Gen LByteString
genLByteString = toLazy <$> Gen.utf8 (Range.constant 0 100) Gen.unicodeAll
```

jiegillet:
Thanks for the review, I think I covered most things except reorganizing the tree of properties, I'm not too sure where to start, I'll take a closer look tomorrow.

test/Test/Toml/BiMap/Property.hs
```diff
@@ -0,0 +1,97 @@
...
+test_Double :: PropertyTest
+test_Double =  prop "Double" $ do
+    x   <- forAll G.genDouble
+    if isNaN x
+      then assert $
+          fmap isNaN (forward B._Double x >>= backward B._Double) == Right True
+      else (forward B._Double x >>= backward B._Double) === Right x
+
+test_BiMaps :: PropertyTest
+test_BiMaps = mconcat
```
ChShersh: In order to organize all tests into single group you just need to replace this `mconcat` with `testGroup`:

* https://hackage.haskell.org/package/tasty-1.2/docs/Test-Tasty.html#v:testGroup

Like this:

```haskell
test_BiMaps = testGroup "BiMap roundtrip tests"
    [ ...
```

jiegillet: Thank you!

PR #161, opened 2019-01-05
[#43] Introduce LHS readme with usage example

### LINKED ISSUE #43: Write 'tomland' tutorial
This tutorial should cover:

1. What is bidirectional conversion?
2. How to write bidirectional converter for simple data type with fields `Integer`, `Text` and `[Double]`?
3. How to use specification value to encode/decode toml?
4. How to parse tables?
5. How to parse nested objects?
6. How to parse list of objects (using array of tables)?

Something else we need to cover?..

ChShersh:
Resolves #43 

I'm currently working on the blog post about `tomland`. So this README with the blog post should resolve the issue. Once blog post is published, I will insert links to it everywhere. And once array of tables are fully implemented by @jiegillet, I will update the example to use them.


README.md
```diff
@@ -1,8 +1,132 @@
...
+_Client = Toml.prism Client $ \case
+    Client n -> Right n
+    other    -> Toml.wrongConstructor "Client" other
+
+userCodec :: TomlCodec User
+userCodec =
+        Toml.match (_Admin >>> Toml._Integer) "id"
+    <|> Toml.match (_Client >>> Toml._Text) "name"
+```
+
```
vrom911: Probably add some comment before the code about some functions that would be used in here, like `diwrap`, `match`. More about which is suitable for which kind of data types: sum types, type aliases etc. What do you think?

README.md
```diff
@@ -1,8 +1,132 @@
...
+## Data type: parsing and printing
+
+Consider the following TOML configuration specified as interpolated text via
+`neat-interpolation` quasi-quoter:
+
+```haskell
+tomlExample :: Text
+tomlExample = [Interpol.text|
```
vrom911: I wonder if it would be useful to have it in the separate toml file, so you won't need to recompile when playing with this example?

README.md
```diff
@@ -1,8 +1,132 @@
...
+```
+
+This static configuration is captured by the following Haskell data type:
+
+```haskell
+data Settings = Settings
+    { settingsPort        :: !Port
+    , settingsDescription :: !Text
+    , settingsMail        :: !Mail
+    , settingsUsers       :: ![User]
```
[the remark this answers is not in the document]
ChShersh: For now: yes. I will add this info after array of tables are fully supported.

README.md
ChShersh: Nice idea! This indeed will be very helpful to understand the code :slightly_smiling_face:

README.md
ChShersh: Hmm.. :thinking: I see value in having this TOML in separate file:

1. Simpler README (no QuasiQuoters and new libraries)
2. Simpler `main` function

Nice idea! Will move this TOML to separate file. Do you think `examples/` directory works as a fine place for this file?

vrom911: Yes, The only issue that I see is it won't be immediately visible to users. Not sure how to solve this, and what is more important

vrom911:
Oh, by the way, what about logo? :wink:

ChShersh:
@vrom911 You can start choosing logo :palm_tree: :smirk: 

* https://www.flaticon.com/search?word=island

README.md
ChShersh: I guess it's better to put in separate file. I think users can follow the link and press back button to return :slightly_smiling_face: Not everyone is familiar with quasi-quoters, and `tomland` already uses a lot of non-beginner features... So let's try to make tutorial simpler.

ChShersh:
@vrom911 I've updated README. Please, have a look again :slightly_smiling_face:

README.md
```diff
@@ -1,8 +1,144 @@
...
+   this data type as TOML table in `.toml` file.
+4. If you have lists of custom data types, use `Toml.list`. Such lists are
+   represented as array of tables in TOML. If you have lists of primitive types
+   like `Int`, `Bool`, `Double`, `Text` or time types, that you can use
+   `Toml.arrayOf` and parse arrays of values.
+5. `tomland` separates conversion between Haskell types and TOML values from
+   matching values by keys. Converters between types and values have type
+   `TomlBiMap` and are named with capital letter started with underscore. Main
+   type for TOML codecs is called `TomlCodec`. To lift `TomlBiMap` to
+   `TomlCodec` you need to use `Toml.match` function.
```
vrom911: Very nicely structured information!

PR #166, opened 2019-01-07
Add benchmarks with toml-parser library

ChShersh:
Added benchmarks with `toml-parser` library. `tomland` actually is not that fast comparing to other libraries... But, again, we can improve performance later :slightly_smiling_face:

### DIFF
--- a/benchmark/Benchmark/HtomlMegaparsec.hs
+++ b/benchmark/Benchmark/HtomlMegaparsec.hs
@@ -3,17 +3,24 @@
 {-# LANGUAGE PackageImports #-}
 
 module Benchmark.HtomlMegaparsec
-       ( convert
-       , decode
+       ( decode
        , parse
+       , convert
        ) where
 
 import Data.Aeson.Types (ToJSON, parseEither, parseJSON, toJSON)
+import Data.Semigroup ((<>))
 import Data.Text (Text)
+
 import "htoml-megaparsec" Text.Toml (Node (..), Table, TomlError, parseTomlDoc)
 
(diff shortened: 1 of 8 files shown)


benchmark/Benchmark/HtomlMegaparsec.hs
```diff
@@ -3,17 +3,24 @@
...
+import Data.Semigroup ((<>))
 import Data.Text (Text)
+
 import "htoml-megaparsec" Text.Toml (Node (..), Table, TomlError, parseTomlDoc)
 
 import Benchmark.Type (HaskellType)
 
+-- | Decode toml file to Haskell type.
+decode :: Text -> Either String HaskellType
+decode txt = case parseTomlDoc "log" txt of
```
vrom911: I see these functions in every module, can we make them polymorphic?

ChShersh: Hmm, I haven't thought about this :thinking: It's not that simple because the use `convert` and `parse` specific to every library. But maybe it's possible to somehow remove code duplication...

willbasky:
> Added benchmarks with toml-parser library. tomland actually is not that fast comparing to other libraries... But, again, we can improve performance later

What statistic on benchmarking  is now?

ChShersh:
@willbasky Results for full decoding:

| lib | decode |
|---|---|
| tomland | 394.2 μs |
| htoml | 839.6 μs |
| htoml-megaparsec | 353.6 μs |
| toml-parser | 162.3 μs |

PR #168, opened 2019-01-07
[#163] Move dates from DateTime

### LINKED ISSUE #163: Move all time data types from nested `DateTime` to `Value`
According to TOML spec it looks like we need to move all time types from nested `DateTime` in single `Value` constructor to 4 `Value` constructors. Because it doesn't make sense to have mixed array of `ZonedTime` and `TimeOfDay`.

vrom911:
Note that parsing should be improved to make it more correct.
Also, I didn't updated the Changelog yet, should I do that?

Resolves #163 ?

### DIFF
--- a/src/Toml/Bi/Map.hs
+++ b/src/Toml/Bi/Map.hs
@@ -74,9 +74,9 @@ import GHC.Generics (Generic)
 import Numeric.Natural (Natural)
 import Text.Read (readEither)
 
-import Toml.Type (AnyValue (..), DateTime (..), MatchError (..), TValue (..), Value (..),
-                  applyAsToAny, matchBool, matchDate, matchDouble, matchInteger, matchText,
-                  mkMatchError, toMArray)
+import Toml.Type (AnyValue (..), MatchError (..), TValue (..), Value (..), applyAsToAny, matchBool,
+                  matchDay, matchDouble, matchHours, matchInteger, matchLocal, matchText,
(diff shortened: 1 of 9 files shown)


src/Toml/Bi/Map.hs
```diff
@@ -244,31 +244,19 @@ _Text = mkAnyValueBiMap matchText Text
 
 -- | Zoned time bimap for 'AnyValue'. Usually used with 'zonedTime' combinator.
 _ZonedTime :: TomlBiMap ZonedTime AnyValue
-_ZonedTime = mkAnyValueBiMap (matchDate >=> getTime) (Date . Zoned)
-  where
-    getTime (Zoned z) = Right z
-    getTime value     = mkMatchError TDate $ Date value
+_ZonedTime = mkAnyValueBiMap matchZoned Zoned
```
ChShersh: These `BiMap`s became much simpler, thank you!

src/Toml/Parser/Value.hs
```diff
@@ -20,52 +20,51 @@ import Data.Char (chr, isControl)
...
+-- | Parser for TOML text.
 textP :: Parser Text
 textP = multilineBasicStringP
-     <|> multilineLiteralStringP
-     <|> literalStringP
-     <|> basicStringP
-     <?> "text"
-
+    <|> multilineLiteralStringP
+    <|> literalStringP
```
ChShersh: Thank you for aligning :+1:

PR #172, opened 2019-01-09
[#145] Support for inline table arrays

### LINKED ISSUE #145: Support array of inline tables
Add parser + tests for inline tables

jiegillet:
Resolves #145 
Added tests as well.


ChShersh:
@jiegillet Probably I'm missing something, but the following TOML file doesn't parse with the parser in this PR.

```toml
bar = [ { x = 5 } ]
```

I see the following error:

```haskell
Left (ParseException "1:9:\n  |\n1 | bar = [ { x = 5 } ]\n  |         ^\nunexpected '{'\nexpecting ']', array, bool, datetime, double, integer, or text\n")
```

jiegillet:
Right, there is something wrong. I'll fix it, thank you for checking!

chshersh:
@jiegillet Hi! Sorry for bothering you, I just would like to know, whether you are still interested in working on this issue? 🙂

jiegillet:
Hi! I've been in a programming funk these last few months, sorry about abandoning you. I'm getting better now, I think I'll get back to it. Please leave the issue open :)

jiegillet:
I now have a local version that passes the tests, but it is built on top of #197 so I'll wait to see the status to push it. It needs some cleaning out, but the performance is the same as #197.

jiegillet:
Small new feature, but it forced me to refactor the code. It's more readable now, I think. I think the performance is slightly less, but not by too much.

src/Toml/Parser/TOML.hs
```diff
@@ -56,54 +52,73 @@ tableNameP = between (text "[") (text "]") keyP
 tableArrayNameP :: Parser Key
 tableArrayNameP = between (text "[[") (text "]]") keyP
 
--- | Parser for lines starting with 'key =', either values or inline tables.
-hasKeyP :: Parser (Key, Either AnyValue TOML)
-hasKeyP = (,) <$> keyP <* text "=" <*> eitherP anyValueP inlineTableP
+-- Helper functions for building TOML
+tomlKV :: Key -> AnyValue -> TOML
+tomlKV k v = mempty { tomlPairs = HashMap.singleton k v}
```
chshersh: Let's add spaces around `{` and `}` in all three functions for consistentcy.

src/Toml/Parser/TOML.hs
```diff
@@ -56,54 +52,73 @@ tableNameP = between (text "[") (text "]") keyP
...
+
+tomlA :: Key -> NonEmpty TOML -> TOML
+tomlA k a = mempty {tomlTableArrays = HashMap.singleton k a}
+
+-- | Parser for lines starting with 'key =', either values, inline tables or
+-- inline arrays of tables.
+hasKeyP :: Maybe Key -> Parser TOML
+hasKeyP key = do
+    k <- keyP <* text "="
+    try (tableArray k) <|> try (table k) <|> (tomlKV k <$> anyValueP)
```
chshersh: Will it work without these two `try`s here? THere is a `try` inside `childKeyP`

jiegillet: Mmh, maybe the one around `table` isn't needed because only an inline table starts with `{`, the one around `tableArray` is needed. Also note that here `childKeyP` will never consume any input because the parser is `pure k`, so the `try` inside is useless.

jiegillet:
It won't parse nested inline arrays. I don't think it should, the documentation says if you start being tempted by this, you should use a table.

PR #173, opened 2019-01-10
[#50] Add decode.encode = id tests

### LINKED ISSUE #50: Property-based tests for encoder and decoder
How this test should looks like:

1. Create big data type with all possible types supported by `tomland`.
2. Write converter for this type.
3. Ensure `decode . encode == id`.

vrom911:
Resolves #50

### DIFF
--- a/test/Test/Toml/BiCode/Property.hs
+++ b/test/Test/Toml/BiCode/Property.hs
@@ -3,7 +3,6 @@ module Test.Toml.BiCode.Property where
 import Control.Applicative ((<|>))
 import Control.Category ((>>>))
 import Data.ByteString (ByteString)
-import Data.Hashable (Hashable)
 import Data.HashSet (HashSet)
 import Data.IntSet (IntSet)
 import Data.List.NonEmpty (NonEmpty)
@@ -17,11 +16,11 @@ import Numeric.Natural (Natural)
 import Toml (TomlBiMap, TomlCodec, (.=))
 import Toml.Bi.Code (decode, encode)
 
-import Test.Toml.Gen (PropertyTest, genBool, genDay, genDouble, genHours, genInteger, genLocal,
-                      genText, genZoned, prop)
+import Test.Toml.Gen (PropertyTest, genBool, genByteString, genDay, genDouble, genFloat, genHashSet,
+                      genHours, genInt, genIntSet, genInteger, genLByteString, genLocal, genNatural,
+                      genNonEmpty, genString, genText, genWord, genZoned, prop)
 
 import qualified Data.ByteString.Lazy as L
-import qualified Data.Text as T
 import qualified Hedgehog.Gen as Gen
 import qualified Hedgehog.Range as Range
 import qualified Toml
@@ -144,8 +143,8 @@ genBigType = do
     btFloat         <- Batman <$> genFloat
     btText          <- genText
     btString        <- genString
-    btBS            <- genBS
-    btLazyBS        <- L.fromStrict <$> genBS
+    btBS            <- genByteString
+    btLazyBS        <- genLByteString
     btLocalTime     <- genLocal
     btDay           <- genDay
     btTimeOfDay     <- genHours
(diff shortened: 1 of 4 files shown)


test/Test/Toml/BiCode/Property.hs
```diff
@@ -0,0 +1,191 @@
...
+    btNewtype       <- genNewType
+    btSum           <- genSum
+    btRecord        <- genRec
+    pure BigType {..}
+
+genInt :: Gen Int
+genInt = Gen.int (Range.constant 0 256)
+
+genWord :: Gen Word
+genWord = Gen.word (Range.constant 1 256)
```
ChShersh: You can use `constantBounded` generator for `Int` and `Word` to avoid specifying boundaries manually. Unfortunately, won't fork for `Double` and `Float`, they don't have instances...

* https://hackage.haskell.org/package/hedgehog-0.6.1/docs/Hedgehog-Internal-Range.html#v:constantBounded

test/Test/Toml/BiCode/Property.hs
```diff
@@ -0,0 +1,191 @@
...
+    , btNewtype       :: BigTypeNewtype
+    , btSum           :: BigTypeSum
+    , btRecord        :: BigTypeRecord
+    } deriving (Show, Eq)
+
+newtype BigTypeNewtype = BigTypeNewtype ZonedTime
+    deriving (Show)
+
+instance Eq BigTypeNewtype where
+    (BigTypeNewtype a) == (BigTypeNewtype b) = zonedTimeToUTC a == zonedTimeToUTC b
```
ChShersh: Very nice idea to move `ZonedTime` inside `newtype`! With this trick you can minimize the size of `Eq` instance and test both `diwrap` and `zonedTime`!

test/Test/Toml/BiCode/Property.hs
```diff
@@ -0,0 +1,191 @@
...
+    , btInteger       :: Integer
+    , btNatural       :: Natural
+    , btInt           :: Int
+    , btWord          :: Word
+    , btDouble        :: Double
+    , btFloat         :: Float
+    , btText          :: Text
+    , btString        :: String
+    , btBS            :: ByteString
+    , btLazyBS        :: L.ByteString
```
ChShersh: Lol, I just noticed that in `tomland` we have `Text`, `ByteString`, `LByteString` and `String` but not `LText`. Heh, that's awkward. Probably need to add `LText` to `tomland` as well. I've created issue for this:

* https://github.com/kowainik/tomland/issues/174

test/Test/Toml/BiCode/Property.hs
ChShersh: @vrom911 Yeah, I think it's a good idea to move general-purpose generators (like `genWord`) into single module, because they are probably going to be reusable in several places (for example, I can see that `BiMap.Property` uses most of the same generators. It's a good idea to keep all basic generators in single place :slightly_smiling_face:

vrom911:
My laptop can't run these tests now, so I leave this on CI. I hope that there is an error in there that makes it hang, otherwise, I don't know how I will run tests..

test/Test/Toml/Gen.hs
```diff
@@ -11,40 +11,72 @@ module Test.Toml.Gen
...
          -- * Generators
-       , genVal
-       , genKey
-       , genPrefixMap
-       , genToml
-       , genBool
+         -- ** Primitive
+       , genInt
        , genInteger
+       , genDouble
```
ChShersh: I like separation of reexports in this export section! :heart_eyes:

test/Test/Toml/Gen.hs
```diff
@@ -156,20 +191,49 @@ genZoned = do
...
+    , (1, Gen.constant $ 1/0)
+    , (1, Gen.constant $ -1/0)
+    , (1, Gen.constant $ 0/0)
     ]
 
+genWord :: MonadGen m => m Word
+genWord = Gen.word Range.constantBounded
+
+genNatural :: MonadGen m => m Natural
+genNatural = fromIntegral <$> Gen.word Range.constantBounded
```
ChShersh: Can use our functions here as well :slightly_smiling_face: 

```suggestion
genNatural = fromIntegral <$> genWord
```

Issue #177, opened 2019-01-13
Implement better generator for key `Piece`


ChShersh:
Currently generator for `Piece` just generates simple bare keys.

https://github.com/kowainik/tomland/blob/90af0bdacf81ffe5a156d214da10dc0568bf1f6f/test/Test/Toml/Gen.hs#L107-L108

But what it should do:

1. Generate bare keys with ASCII letters, ASCII digits, underscores, and dashes (`A-Za-z0-9_-`)
2. Generate quoted keys (both basic and literal strings) with unicode inside.

This may take quite some time and effort to implement such generator, so putting it into separate issue.

cronokirby:
I believe this is resolved as of: https://github.com/kowainik/tomland/pull/184

PR #184, opened 2019-02-04
Implement a better generator for `Piece`

cronokirby:
This PR addresses #177.

This modifies the generator for `Piece` to either generate a bare piece, with alphanumeric and dashes, otherwise it generates either single quoted `''` or double quoted `""` strings containing arbitrary unicode.

All existing tests pass, thankfully.

I've updated the changelog with version 1.0.1, not sure if this is the right one to have there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 tomland uses [PVP Versioning][1].
 The changelog is available [on GitHub][2].
 
+## 1.0.1
+* [#177](https://github.com/kowainik/tomland/issues/177):
+  Add a more extensive property generator for `Piece`.
+
 ## 1.0.0 — Jan 14, 2019
 
 * [#13](https://github.com/kowainik/tomland/issues/13):
--- a/cabal.project
+++ b/cabal.project
@@ -1,0 +1,1 @@
+allow-newer: htoml-megaparsec:unordered-containers
--- a/cabal.project
+++ b/cabal.project
@@ -1,1 +1,3 @@
+packages: .
+
 allow-newer: htoml-megaparsec:unordered-containers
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 tomland uses [PVP Versioning][1].
 The changelog is available [on GitHub][2].
 
+## 1.0.1
```
ChShersh: Version bump up is correct :+1: Though let's add `Unreleased` tag for clarity
```suggestion
## Unreleased: 1.0.1
```

cronokirby: Makes sense

cronokirby:
For 610.3 and 610.1, the build error doesn't even seem to be in the code from this PR :thinking: 

As for the failing test case in 610.2, it seems to be that "\NUL" fails to parse as a key. I'm not even sure if "\NUL" is a valid quoted key. In theory unicode is valid, but having \NUL in keys would make parsing toml very tricky in C :thinking:

ChShersh:
@cronokirby Yeah, `"\NUL"` is invalid inside TOML keys. I already had some troubles with such values. Generators should be smart enough to not produce `"\NUL"`. Unfortunately, I don't see standard generator in the `hedgehog` library that does that. Which means that we need to patch our generators 😞 

> For 610.3 and 610.1, the build error doesn't even seem to be in the code from this PR

Yeah, that's true. Most likely new version of some library (in this case `unordered-containers`) came up and broke some code. Looks like in this case `htoml-megaparsec` has stricter constraints for `unordered-containers` (dunno why though). What I can propose to try is to add the following line to `cabal.project`:

```
allow-newer: htoml-megaparsec:unordered-containers
```

cabal.project
```diff
@@ -0,0 +1 @@
+allow-newer: htoml-megaparsec:unordered-containers
```
ChShersh: Oh, I see why the build is failing. You also need to specify `packages` in `cabal.project`

```suggestion
packages: .

allow-newer: htoml-megaparsec:unordered-containers
```

cronokirby: aha

Issue #190, opened 2019-05-21
Codec for nullary sum types 


mxxo:
Take a simple sum data type like this:

```haskell
data Color = Blue | Yellow | Green | Red
```
Without any fields, I'm not sure how to make the existing sum type parsing approach using prisms work here (as shown in the Readme).

PR #191, opened 2019-05-21
[#190] Nullary sum types

mxxo:
Possible fix for #190. I want to use many nullary types as configuration options and this patch lets me add them without much boilerplate. 

The requirement is making your sum types derive Read, Show, Enum, and Bounded. 

```haskell
-- | Requested CPU resources
data CpuType = Serial
             | Parallel
             deriving (Read, Show, Enum, Bounded)

-- | Output options
data RunType = Standard
             | Trial
             | Movie
             deriving (Read, Show, Enum, Bounded)
```

Then, optionally, you can claim a special key for your type: 

```haskell

runType :: TomlCodec RunType
runType = sumType "type"

cpuType :: TomlCodec CpuType
cpuType = sumType "cpu"

-- | Control block determines what type of simulation is run
data ControlBlock = ControlBlock
    { controlRunType :: RunType
    , cpu            :: CpuType
    } deriving (Show)

controlBlockCodec :: TomlCodec ControlBlock
controlBlockCodec = flip Toml.table "control" $
    ControlBlock
    <$> runType  .= controlRunType
    <*> cpuType  .= cpu
```

All the logic, including error handling, is updated automatically when you add or subtract values:
```shell
--- reading run control file
tomland decode error:  Value is Cereal but expected one of: Serial, Parallel
```
```haskell
-- add more options to a type
data CpuType = Cereal 
             | Serial
             | Parallel
             | Skynet
             deriving (Read, Show, Enum, Bounded)
```
```shell
--- reading run control file

[control]
  cpu = "Cereal"
  type = "Standard"


--- reading run control file
tomland decode error:  Value is ereal but expected one of: Cereal, Serial, Parallel, Skynet

```

As a possible extension, we could use the _TextBy codec to specify custom conversion functions if there's some reason that the sum data format "True, False, Red, Yellow, ..." isn't enough.


examples/Playground.hs
```diff
@@ -58,6 +64,7 @@ testT = Test
     <*> Toml.table insideCodec "testX" .= testX
     <*> Toml.dioptional ((Toml.table insideCodec) "testY") .= testY
     <*> Toml.diwrap (Toml.text "testN") .= testN
+    <*> Toml.sumType "testC" .= testC
```
chshersh: Technically speaking, `tomland` already has a way to write bidirectional serializers for enum types. You could just use [`Toml.read`](http://hackage.haskell.org/package/tomland-1.0.1.0/docs/Toml-Bi-Combinators.html#v:read). But I see that your `sumType` provides much better error messages, and it's always good to have better error messages!

src/Toml/Bi/Combinators.hs
```diff
@@ -162,6 +163,11 @@ string = match _String
 read :: (Show a, Read a) => Key -> TomlCodec a
 read = match _Read
 
+-- | Codec for general nullary sum data types with a 'Bounded', 'Enum', 'Read',
+-- and 'Show' instance.
+sumType :: (Bounded a, Enum a, Read a, Show a) => Key -> TomlCodec a
+sumType = match _SumType
```
chshersh: `sumType` is a quite generic name. This codec doesn't support general sum types. So let's rename to `enumBounded` so the name reflects the idea. And also let's rename `_SumType` to `_EnumBounded`.

src/Toml/Bi/Combinators.hs
```diff
@@ -162,6 +163,11 @@ string = match _String
 read :: (Show a, Read a) => Key -> TomlCodec a
 read = match _Read
 
+-- | Codec for general nullary sum data types with a 'Bounded', 'Enum', 'Read',
+-- and 'Show' instance.
```
chshersh:
```suggestion
-- and 'Show' instance. This codec provides much better error messages than 'read' for nullary sum types.
```

src/Toml/Bi/Map.hs
```diff
@@ -381,6 +383,29 @@ _BoundedInteger = BiMap (Right . toInteger) eitherBounded
...
+    { forward  = Right . tShow
+    , backward = toSumType
+    }
+  where
+    toSumType :: forall a. (Read a, Show a, Enum a, Bounded a) => Text -> Either TomlBiMapError a
+    toSumType value
+      | value `elem` options =
+          Right (fromText value)
+      | otherwise =
+          let msg = "Value is " <> value <> " but expected one of: " <> T.intercalate ", " options
```
chshersh:
```suggestion
          let msg = "Value is '" <> value <> "' but expected one of: " <> T.intercalate ", " options
```

src/Toml/Bi/Map.hs
```diff
@@ -381,6 +383,29 @@ _BoundedInteger = BiMap (Right . toInteger) eitherBounded
...
       | otherwise = Right (fromIntegral n)
 
+-- | Helper bimap for 'SumType' and 'Data.Text.Text'.
+_SumTypeText :: (Read a, Show a, Enum a, Bounded a) => TomlBiMap a Text
+_SumTypeText = BiMap
+    { forward  = Right . tShow
+    , backward = toSumType
+    }
+  where
+    toSumType :: forall a. (Read a, Show a, Enum a, Bounded a) => Text -> Either TomlBiMapError a
```
chshersh: You don't need to write all constraints and a `forall` here. You only need to write `forall` on the top-level for `_SumTypeText` (to be renamed to `_EnumBoundedText`). Since `-XScopedTypeVariables` is already enabled for the whole project.
```suggestion
    toSumType :: Text -> Either TomlBiMapError a
```

src/Toml/Bi/Map.hs
```diff
@@ -381,6 +383,29 @@ _BoundedInteger = BiMap (Right . toInteger) eitherBounded
...
+    toSumType value
+      | value `elem` options =
+          Right (fromText value)
```
chshersh: I think that this implementation can be slightly efficient and more robust using common programming patterns from Haskell. Currently, you are checking the text value and then parsing it. Let's just get rid of parsing 🙂 

You can create value of type `[(Text, a)]` where the first element of a pair is a shown enum value and the second element is a value itself. So instead of `elem + fromText` you can just do [`lookup`](https://hackage.haskell.org/package/base-4.12.0.0/docs/Prelude.html#v:lookup) and pattern-match on `Maybe`.

mxxo:
Thank you very much for your help and the great library!

Issue #189, opened 2019-05-17
Support custom sorting of keys during pretty-printing


chshersh:
Sorting is done via `PrintOptions`:

https://github.com/kowainik/tomland/blob/e7f36914311e6c1fe50ed50606c4fe9b2ed8aef6/src/Toml/Printer.hs#L27-L31

However, custom sorting of keys is not supported. In order to support it this data type should be generalised as follows:

```haskell
data PrintOptions = PrintOptions
    { printOptionsSorting :: !(Maybe (Key -> Key -> Ordering))
    , printOptionsIndent  :: !Int
    } 
```

Couple notes:

1. Fields are renamed according to our style guide.
2. Fields are strict now to make code more robust.
3. Instead of boolean flag we now have optional comparison function.

cc @ramanshah

ramanshah:
Thanks so much for the writeup (and pointers)! This is exactly the enhancement I was interested in. I'd be happy to try implementing this or helping any way that I can. Let me know.

chshersh:
@ramanshah Sure, feel free to work on that issue! I described the beginning of the implementation. Let us know if something is not clear :slightly_smiling_face:

ramanshah:
On it! Talk soon :)

ramanshah:
I'm a little stuck with the build system. This is on Mac OS X Sierra 10.12.6.

```
master ~/tomland> stack --version
Version 1.7.1, Git revision 681c800873816c022739ca7ed14755e85a579565 (5807 commits) x86_64 hpack-0.28.2
master ~/tomland> stack ghc -- --version
The Glorious Glasgow Haskell Compilation System, version 8.6.5
master ~/tomland> cabal --version
cabal-install version 2.4.1.0
compiled using version 2.4.1.0 of the Cabal library
master ~/tomland> stack build
Unable to parse cabal file ~/tomland/tomland.cabal

- 0:0: Unsupported cabal-version. See https://github.com/haskell/cabal/issues/4899.

exit 1
```

I've tried to Google my way out of this at some length but am still stumped. Any pointers? I'd be happy to update `CONTRIBUTING.md` if there's some prerequisite step that newcomers may not know to do.

ramanshah:
Just figured it out! I learned that the relevant `cabal` (library) is baked into the `stack` version. Running `stack upgrade` to bump my `stack` to 1.9.3 got me a clean build and passing test suite.

Is there a good place to annotate this dependency?

ramanshah:
[...]

PR #192, opened 2019-05-28
[#189] Implement custom table key sorting

ramanshah:
@chshersh As discussed, this code does not yet compile, but I'm opening this PR to facilitate discussion. Thanks!

### DIFF
--- a/src/Toml/Printer.hs
+++ b/src/Toml/Printer.hs
@@ -26,9 +26,9 @@ import qualified Data.Text as Text
 
 {- | Configures the pretty printer. -}
 data PrintOptions = PrintOptions
-    {- | How table keys should be sorted, if at all. -}
-    { printOptionsSorting :: !(Maybe (Key -> Key -> Ordering))
-    {- | Number of spaces by which to indent. -}
+    { -- | How table keys should be sorted, if at all.
+      printOptionsSorting :: !(Maybe (Key -> Key -> Ordering))
+      -- | Number of spaces by which to indent.
     , printOptionsIndent  :: !Int
     }
 
@@ -80,7 +80,7 @@ prettyTomlInd :: PrintOptions -- ^ Printing options
               -> Int          -- ^ Current indentation
               -> Text         -- ^ Accumulator for table names
               -> TOML         -- ^ Given 'TOML'
-              -> [Text]         -- ^ Pretty result
+              -> [Text]       -- ^ Pretty result
 prettyTomlInd options i prefix TOML{..} = concat
     [ prettyKeyValue    options i tomlPairs
     , prettyTables      options i prefix tomlTables
@@ -173,7 +173,7 @@ prettyTableArrays options i pref = mapOrdered arrText options
 
 -- Returns an indentation prefix
 tabWith :: PrintOptions -> Int -> Text
-tabWith options n = Text.replicate (n * printOptionsIndent options) " "
+tabWith PrintOptions{..} n = Text.replicate (n * printOptionsIndent) " "
 
 -- Returns a proper sorting function
 mapOrdered :: ((Key, v) -> [t]) -> PrintOptions -> HashMap Key v -> [t]
@@ -181,9 +181,9 @@ mapOrdered f options = case printOptionsSorting options of
     Just sorter -> concatMap f . sortBy (applyToFirst sorter) . HashMap.toList
     Nothing     -> concatMap f . HashMap.toList
 
--- Compares pairs of tuples by their first elements with a custom sorter
-applyToFirst :: Ord a => (a -> a -> Ordering) -> (a, b) -> (a, b) -> Ordering
-applyToFirst sorter x y = sorter (fst x) (fst y)
+-- Applies a binary function to the first elements of tuples
+applyToFirst :: (a -> b -> c) -> (a, x) -> (b, y) -> c
+applyToFirst f x y = f (fst x) (fst y)
 
 -- Adds next part of the table name to the accumulator.
 addPrefix :: Key -> Text -> Text
--- a/src/Toml/Printer.hs
+++ b/src/Toml/Printer.hs
@@ -11,7 +11,7 @@ module Toml.Printer
        ) where
 
 import Data.HashMap.Strict (HashMap)
-import Data.List (sortOn, splitAt)
+import Data.List (sortBy, splitAt)
 import Data.List.NonEmpty (NonEmpty)
 import Data.Monoid ((<>))
 import Data.Text (Text)
@@ -26,9 +26,11 @@ import qualified Data.Text as Text
 
 {- | Configures the pretty printer. -}
 data PrintOptions = PrintOptions
-    { shouldSort :: Bool  -- ^ should table keys be sorted or not
-    , indent     :: Int   -- ^ indentation size
-    } deriving (Show)
+    {- | How table keys should be sorted, if at all. -}
+    { printOptionsSorting :: !(Maybe (Key -> Key -> Ordering))
+    {- | Number of spaces by which to indent. -}
+    , printOptionsIndent  :: !Int
+    }
 
 {- | Default printing options.
 
@@ -36,7 +38,7 @@ data PrintOptions = PrintOptions
 2. Indents with 2 spaces.
 -}
 defaultOptions :: PrintOptions
-defaultOptions = PrintOptions True 2
+defaultOptions = PrintOptions (Just compare) 2
 
 {- | Converts 'TOML' type into 'Data.Text.Text' (using 'defaultOptions').
 
@@ -171,13 +173,17 @@ prettyTableArrays options i pref = mapOrdered arrText options
 
 -- Returns an indentation prefix
 tabWith :: PrintOptions -> Int -> Text
-tabWith options n = Text.replicate (n * indent options) " "
+tabWith options n = Text.replicate (n * printOptionsIndent options) " "
 
 -- Returns a proper sorting function
-mapOrdered :: Ord k => ((k, v) -> [t]) -> PrintOptions -> HashMap k v -> [t]
-mapOrdered f options
-    | shouldSort options = concatMap f . sortOn fst . HashMap.toList
-    | otherwise          = concatMap f . HashMap.toList
+mapOrdered :: ((Key, v) -> [t]) -> PrintOptions -> HashMap Key v -> [t]
+mapOrdered f options = case printOptionsSorting options of
+    Just sorter -> concatMap f . sortBy (applyToFirst sorter) . HashMap.toList
+    Nothing     -> concatMap f . HashMap.toList
+
+-- Compares pairs of tuples by their first elements with a custom sorter
+applyToFirst :: Ord a => (a -> a -> Ordering) -> (a, b) -> (a, b) -> Ordering
+applyToFirst sorter x y = sorter (fst x) (fst y)
 
 -- Adds next part of the table name to the accumulator.
 addPrefix :: Key -> Text -> Text
--- a/src/Toml/Printer.hs
+++ b/src/Toml/Printer.hs
@@ -130,8 +130,14 @@ prettyKeyValue options i = mapOrdered (\kv -> [kvText kv]) options
 
 -- | Returns pretty formatted tables section of the 'TOML'.
 prettyTables :: PrintOptions -> Int -> Text -> PrefixMap TOML -> [Text]
-prettyTables options i pref = mapOrdered (prettyTable . snd) options
+prettyTables options i pref asPieces = mapOrdered (prettyTable . snd) options asKeys
   where
+    asKeys :: HashMap Key (PrefixTree TOML)
+    asKeys = HashMap.fromList . map pieceToKey $ HashMap.toList asPieces
+
+    pieceToKey :: (Piece, a) -> (Key, a)
+    pieceToKey (piece, x) = (Key (pure piece), x)
+
     prettyTable :: PrefixTree TOML -> [Text]
     prettyTable (Leaf k toml) =
         let name = addPrefix k pref
--- a/src/Toml/Printer.hs
+++ b/src/Toml/Printer.hs
@@ -137,7 +137,7 @@ prettyTables options i pref asPieces = mapOrdered (prettyTable . snd) options as
     asKeys = map (first pieceToKey) $ HashMap.toList asPieces
 
     pieceToKey :: Piece -> Key
-    pieceToKey piece = Key (pure piece)
+    pieceToKey = Key . pure
 
     prettyTable :: PrefixTree TOML -> [Text]
     prettyTable (Leaf k toml) =
--- a/src/Toml/Printer.hs
+++ b/src/Toml/Printer.hs
@@ -11,6 +11,7 @@ module Toml.Printer
        ) where
 
 import Data.Bifunctor (first)
+import Data.Function (on)
 import Data.HashMap.Strict (HashMap)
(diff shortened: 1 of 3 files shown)


jiegillet:
I agree with you assessment of the situation in the discussion. The previous ordering worked both for the lists of key/value pairs `[(Key, v)]` and for ordering the tables `[(Piece, v)]`, but if you want a specific ordering, you need to have two orderings and split `mapOrdered` in `mapOrderedKV` and `mapOrderedTables` or something like that.

src/Toml/Printer.hs
```diff
@@ -171,13 +173,17 @@ prettyTableArrays options i pref = mapOrdered arrText options
...
 tabWith :: PrintOptions -> Int -> Text
-tabWith options n = Text.replicate (n * indent options) " "
+tabWith options n = Text.replicate (n * printOptionsIndent options) " "
 
 -- Returns a proper sorting function
-mapOrdered :: Ord k => ((k, v) -> [t]) -> PrintOptions -> HashMap k v -> [t]
-mapOrdered f options
-    | shouldSort options = concatMap f . sortOn fst . HashMap.toList
-    | otherwise          = concatMap f . HashMap.toList
+mapOrdered :: ((Key, v) -> [t]) -> PrintOptions -> HashMap Key v -> [t]
```
chshersh: Okay, I see the problem now. I don't think we should introduce a separate comparator for `Key` and `Piece`, because `Piece` can be considered as a special case of `Key`. So here in this function we can pass a list of type `[(Key, v)]` instead of `HashMap Key v` and convert to this list on all call-sites.

src/Toml/Printer.hs
```diff
@@ -171,13 +173,17 @@ prettyTableArrays options i pref = mapOrdered arrText options
...
+mapOrdered f options = case printOptionsSorting options of
+    Just sorter -> concatMap f . sortBy (applyToFirst sorter) . HashMap.toList
```
chshersh: You don't need to write `applyToFirst`, you can use standard function `comparing` here:

* https://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Ord.html#v:comparing

src/Toml/Printer.hs
```diff
@@ -171,13 +173,17 @@ prettyTableArrays options i pref = mapOrdered arrText options
 
 -- Returns an indentation prefix
 tabWith :: PrintOptions -> Int -> Text
-tabWith options n = Text.replicate (n * indent options) " "
+tabWith options n = Text.replicate (n * printOptionsIndent options) " "
```
chshersh: We are using `RecordWildCards` so such code usually can be written a little bit nicer 🙂 I think it looks a bit cleaner.
```suggestion
tabWith PrintOptions{..} n = Text.replicate (n * printOptionsIndent) " "
```

src/Toml/Printer.hs
```diff
@@ -26,17 +26,19 @@ import qualified Data.Text as Text
 
 {- | Configures the pretty printer. -}
 data PrintOptions = PrintOptions
-    { shouldSort :: Bool  -- ^ should table keys be sorted or not
-    , indent     :: Int   -- ^ indentation size
-    } deriving (Show)
+    {- | How table keys should be sorted, if at all. -}
+    { printOptionsSorting :: !(Maybe (Key -> Key -> Ordering))
+    {- | Number of spaces by which to indent. -}
+    , printOptionsIndent  :: !Int
```
chshersh: Oh, I just noticed. Documentation here is formatted a little bit wrong. It should be:

```haskell
{- | Configures the pretty printer. -}
data PrintOptions = PrintOptions
    { -- | How table keys should be sorted, if at all. 
    , printOptionsSorting :: !(Maybe (Key -> Key -> Ordering))

      -- | Number of spaces by which to indent.
    , printOptionsIndent  :: !Int
    }
```

src/Toml/Printer.hs
ramanshah: Perhaps I'm missing something? `comparing` doesn't accept a user-provided `a -> a -> Ordering` sorter. I couldn't find anything in the standard library that applies a custom sorter to tuples, either. (I attempted at some length to discover some slick `Functor` or `Applicative` thing to "lift" the user-provided sorter to tuples, but I failed and figured this one-line helper would be more readable anyway.)

src/Toml/Printer.hs
ramanshah: Awesome - this was exactly the insight I was needing! Let me work on fleshing it out :)

src/Toml/Printer.hs
chshersh: @ramanshah ah, sorry, that was an oversight from my side. Here you indeed can't use `comparing`. I also don't see any elegant implementation using some standard functions. In that case, I can only recommend changing the type signature of the `applyToFirst` function further using _generalisation programming pattern_ in Haskell to the following (so there is only one possible implementation without any chance to introduce error):

```haskell
applyToFirst :: (a -> b -> c) -> (a, x) -> (b, y) -> c
```

ramanshah: Very cool idea. I'm learning a lot on in this PR :)

ramanshah:
I got us a clean compilation and clean test suite!

I'm going to do a bit of refactoring (to respect the requested type signature for `mapOrdered`) and add test coverage of the custom sorting feature. Stay tuned. Are there any docs I need to build or update?

src/Toml/Printer.hs
```diff
@@ -171,13 +179,17 @@ prettyTableArrays options i pref = mapOrdered arrText options
...
 tabWith :: PrintOptions -> Int -> Text
-tabWith options n = Text.replicate (n * indent options) " "
+tabWith PrintOptions{..} n = Text.replicate (n * printOptionsIndent) " "
 
 -- Returns a proper sorting function
-mapOrdered :: Ord k => ((k, v) -> [t]) -> PrintOptions -> HashMap k v -> [t]
-mapOrdered f options
-    | shouldSort options = concatMap f . sortOn fst . HashMap.toList
-    | otherwise          = concatMap f . HashMap.toList
+mapOrdered :: ((Key, v) -> [t]) -> PrintOptions -> HashMap Key v -> [t]
```
chshersh: `mapOrdered` function should take `[(Key, v)]` instead of `HashMap`. This will make the function more performant. In the implementation of this function you're calling `toList` function anyways. But when you convert `HashMap Piece v` to `HashMap Key v` you spend extra time converting back to `HashMap` which is redundant, since you are going to work with the list anywas, so you can convert `HashMap Piece v` to `[(Key, v)]`

src/Toml/Printer.hs
```diff
@@ -128,8 +130,14 @@ prettyKeyValue options i = mapOrdered (\kv -> [kvText kv]) options
...
 -- | Returns pretty formatted tables section of the 'TOML'.
 prettyTables :: PrintOptions -> Int -> Text -> PrefixMap TOML -> [Text]
-prettyTables options i pref = mapOrdered (prettyTable . snd) options
+prettyTables options i pref asPieces = mapOrdered (prettyTable . snd) options asKeys
   where
+    asKeys :: HashMap Key (PrefixTree TOML)
+    asKeys = HashMap.fromList . map pieceToKey $ HashMap.toList asPieces
+
+    pieceToKey :: (Piece, a) -> (Key, a)
+    pieceToKey (piece, x) = (Key (pure piece), x)
```
chshersh: You can implement `pieceToKey` using `first` function from `Data.Bifunctor`. There is `Bifunctor` instance for tuple.

src/Toml/Printer.hs
ramanshah: Got it. Sorry, this code wasn't ready for re-review :) I will definitely do this!

src/Toml/Printer.hs
ramanshah: Learned something new here :)

ramanshah:
All set! I addressed the above comments and added a "golden" test of the custom sorting. I'm happy to add documentation as needed (and also would love comments on how to write a custom `Key -> Key -> Ordering` sorter nicely).

test/Test/Toml/Printer/Golden.hs
```diff
@@ -34,17 +35,31 @@ example = mkToml $ do
...
+spamEggDecorate k
+    | k == fromText "spam" = (0, fromText "spam")
+    | k == fromText "egg" = (1, fromText "egg")
+    | otherwise = (2, k)
+  where
+    fromText :: Text -> Key
+    fromText x = Key $ pure $ Piece x
+
+spamEgg :: Key -> Key -> Ordering
+spamEgg k1 k2 = compare (spamEggDecorate k1) (spamEggDecorate k2)
```
chshersh: This function can be written a little bit shorter:
```suggestion
spamEgg = comparing spamEggDecorate
```

test/Test/Toml/Printer/Golden.hs
```diff
@@ -34,17 +35,31 @@ example = mkToml $ do
...
-    , indent     = 0
+    { printOptionsSorting = Nothing
+    , printOptionsIndent  = 0
     }
 
+-- | Decorate keys as tuples so spam comes before egg
+spamEggDecorate :: Key -> (Int, Key)
+spamEggDecorate k
+    | k == fromText "spam" = (0, fromText "spam")
+    | k == fromText "egg" = (1, fromText "egg")
```
chshersh: You don't need to call `fromText` function here. There is `IsString` instance for `Key`. And we have `OverloadedStrings` language extension enabled project-wide. So you can just write the following and it will work:

```haskell
k == "egg" = (1, "egg")
```

src/Toml/Printer.hs
```diff
@@ -128,8 +131,14 @@ prettyKeyValue options i = mapOrdered (\kv -> [kvText kv]) options
...
 -- | Returns pretty formatted tables section of the 'TOML'.
 prettyTables :: PrintOptions -> Int -> Text -> PrefixMap TOML -> [Text]
-prettyTables options i pref = mapOrdered (prettyTable . snd) options
+prettyTables options i pref asPieces = mapOrdered (prettyTable . snd) options asKeys
   where
+    asKeys :: [(Key, PrefixTree TOML)]
+    asKeys = map (first pieceToKey) $ HashMap.toList asPieces
+
+    pieceToKey :: Piece -> Key
+    pieceToKey piece = Key (pure piece)
```
chshersh: I think eta-reduaction is okay here
```suggestion
    pieceToKey = Key . pure
```

test/Test/Toml/Printer/Golden.hs
ramanshah: Oh that's so much better!

ramanshah:
All set!

test/Test/Toml/Printer/Golden.hs
```diff
@@ -34,17 +35,28 @@ example = mkToml $ do
...
 
+-- | Decorate keys as tuples so spam comes before egg
+spamEggDecorate :: Key -> (Int, Key)
+spamEggDecorate k
+    | k == "spam" = (0, "spam")
+    | k == "egg" = (1, "egg")
+    | otherwise = (2, k)
+
+spamEgg :: Key -> Key -> Ordering
+spamEgg = comparing spamEggDecorate
```
jiegillet: Maybe a slightly more simple example of defining a custom ordering could be `spamEgg = compare on length` or something like that.

src/Toml/Printer.hs
```diff
@@ -171,13 +180,17 @@ prettyTableArrays options i pref = mapOrdered arrText options
...
+tabWith PrintOptions{..} n = Text.replicate (n * printOptionsIndent) " "
 
 -- Returns a proper sorting function
-mapOrdered :: Ord k => ((k, v) -> [t]) -> PrintOptions -> HashMap k v -> [t]
-mapOrdered f options
-    | shouldSort options = concatMap f . sortOn fst . HashMap.toList
-    | otherwise          = concatMap f . HashMap.toList
+mapOrdered :: ((Key, v) -> [t]) -> PrintOptions -> [(Key, v)] -> [t]
+mapOrdered f options = case printOptionsSorting options of
+    Just sorter -> concatMap f . sortBy (applyToFirst sorter)
```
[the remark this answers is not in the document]
ramanshah: Oh very nice! That's the elegant thing we were looking for previously :)

test/Test/Toml/Printer/Golden.hs
ramanshah: Realistic use-cases of this feature involve key orders chosen semantically to serve human comprehension. They are usually arbitrary with respect to the key text. I'd like there to be some advertisement of how to to pretty-print a textually arbitrary key order. I think tuple decoration is a straightforward way to do so. This test is the only such advertisement so far. I'd personally have trouble reverse-engineering `compare \`on\` length` to design a semantically relevant sorter. I suffered an extreme amount of this type of obscurity among the Haskell YAML options, finally migrating a project to TOML because `tomland`'s codebase and docs are actually comprehensible. This type of comprehensibility matters a lot to me.

If it's important to keep the test code concise, we can go this way. But if so, let's discuss where to present a realistic pattern for a custom sort, either in the package documentation or in the blogosphere.

src/Toml/Printer.hs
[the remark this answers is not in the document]
jiegillet: Right, I missed that discussion. Well, better late than never ^^

test/Test/Toml/Printer/Golden.hs
jiegillet: I often use the pattern ``compare `on` f`` so to me the meaning is super clear: sort by string length. I understand if you are less used to it and you prefer more explicit, semantic functions, it's OK to do that. 

However, I feel a bit uneasy about the signature `spamEggDecorate :: Key -> (Int, Key)`. You are doing a function to sort `Key`, so you transform it to an `Int` but you still keep the `Key`, if two keys have the same integer value, you default to alphabetical order. I guess it's not such a big deal, but maybe `spamEggDecorate :: Key -> Int` would be more clear? It's up to you.

I can't really think of an ordering I'd want more than plain alphabetical to be honest, so I'll leave it up to you. As far as I'm concerned, the current PR is great.

ramanshah: I grant that `(Int, Key)` has more cardinality than actually used in the sort.

The common use-case that motivated this feature is when one field in an array of tables/objects is critical for comprehension but gets buried in an alphabetical pretty-printing. Using an example from the `toml` readme, an alphabetically sorted array of tables might look like this:

```
[[products]]
name = "Hammer"
sku = 738594937

[[products]]
color = "gray"
name = "Nail"
sku = 284758393
```

Putting the name up front makes the table easier for a human to scan:

```
[[products]]
name = "Hammer"
sku = 738594937

[[products]]
name = "Nail"
color = "gray"
sku = 284758393
```

Realistic data structures tend to suffer from a more extreme version of this problem because they tend to be bigger, often with substantial nesting; I've found alphabetically pretty-printed data structures over the course of my career surprisingly hard to read. I frequently want "alphabetical with exceptions" to make pretty-printed data more legible, and the given sorter implements this.

For the purpose of the test suite, the given sorter also corresponds to a minimal change in the golden file (transposing two lines of TOML compared to the alphabetical sort, which I was able to construct by hand), helping to verify correctness of the library code.

jiegillet: I see, you make a good case. For the purpose of wanting to get `name` on top, `(Int, Key)` is actually nice since anything non-covered explicitly will be alphabetical sorted. 

Thank you for sharing your insights on realistic uses of config files, that is helpful to me.

Issue #195, opened 2019-06-11
Exponential Edge Case in Parser


stephenmac7:
Parsing the file below takes 6 seconds on my machine, even though it's pretty small. Might be an issue with backtracking in the parser. For more complicated files, there is exponential blowup.

```
[[a]]
[[a]]
[[a]]
[[a]]
[[a]]
[[a]]
[[a]]
[[a]]
[[a]]
[[a]]
[[a]]
[[a]]
[[a]]
[[a]]
[[a]]
```

chshersh:
@stephenmac7 Thanks for reporting the problem! Parsing TOML files is indeed not that simple, because parsing should be context-sensitive, unfortunately, in the current implementation. Though I have some ideas on how to improve the situation, I will try to look into this problem and see what can be done.

jiegillet:
I played around and can confirm the issue. Unfortunately I can't think of a quick fix, but I'll look into it too.

nhibberd:
If its useful, I have made an attempt at a fix here (not complete but progress) - https://github.com/kowainik/tomland/compare/master...nhibberd:master

chshersh:
@nhibberd Thanks for your attempt to fix the problem! I would say that if tests are passing and if the example from this issue is parsed without performance issues, then the PR can be merged. We have quite exhaustive test-suite, though adding a couple more tests never hurts :slightly_smiling_face:

stephenmac7:
The fork @nhibberd posted fixed my issue, but the tests don't pass.

chshersh:
The issue is fixed by @jiegillet. It's already in `master`.

PR #197, opened 2019-06-22
[#195] Fix parser exponential behavior

jiegillet:
Fixed the exponential behavior, I've had to change a bunch of functions, but in the end it let me refactor `tomlP` nicely so I think it's worth it. Benchmarking also improved.

Since some signatures changed, I've had to modify some tests.

I've also added an `eof` to the TOML parser to make sure the whole `.toml` file is valid.

Benchmarks:
Without fix:
```
benchmarked Parse/tomland
time                 440.8 μs   (430.0 μs .. 453.9 μs)
                     0.993 R²   (0.989 R² .. 0.997 R²)
mean                 433.0 μs   (427.9 μs .. 440.4 μs)
std dev              20.27 μs   (14.42 μs .. 31.25 μs)
variance introduced by outliers: 27% (moderately inflated)
```

With fix:
```
benchmarked Parse/tomland
time                 355.8 μs   (348.9 μs .. 368.5 μs)
                     0.990 R²   (0.980 R² .. 0.998 R²)
mean                 360.6 μs   (356.1 μs .. 368.5 μs)
std dev              19.44 μs   (13.67 μs .. 27.89 μs)
variance introduced by outliers: 32% (moderately inflated)
```
That's a 15% speedup.

### DIFF
--- a/src/Toml/Parser/TOML.hs
+++ b/src/Toml/Parser/TOML.hs
@@ -69,7 +69,7 @@ inlineTableP = between
     (tomlFromInline <$> hasKeyP `sepEndBy` text ",")
 
 
--- | Parser for an array of tables.
+-- | Parser for an array of tables under a certain key.
 tableArrayP :: Key -> Parser (NonEmpty TOML)
 tableArrayP key =
     tableP (Just key) `NC.sepBy1` sameKeyP key tableArrayNameP
@@ -78,7 +78,7 @@ tableArrayP key =
 tomlP :: Parser TOML
 tomlP = sc *> tableP Nothing <* eof
 
--- | Parser for full 'TOML' under a certain key
+-- | Parser for a table under a certain key
 tableP :: Maybe Key -> Parser TOML
 tableP key = do
     (val, inline)  <- distributeEithers <$> many hasKeyP
@@ -100,8 +100,9 @@ tableP key = do
         , tomlTableArrays = HashMap.fromList array
         }
 
--- | @childKeyP key p@ returns the result of @p@ if the key returned by @p@ is
--- a child key of the @key@, and fails otherwise.
+-- | @childKeyP (Just key) p@ checks if the result of @p@ if a child key of
+-- @key@ and returns the difference of the keys and the child key.
+-- @childKeyP Nothing p@ is only called from @tomlP@ (no parent key).
 childKeyP :: Maybe Key -> Parser Key -> Parser (Key, Key)
 childKeyP Nothing parser = try $ do
   k <- parser
--- a/src/Toml/Parser/TOML.hs
+++ b/src/Toml/Parser/TOML.hs
@@ -5,7 +5,7 @@ parsers from "Toml.Parser.Value".
 module Toml.Parser.TOML
        ( keyP
        , hasKeyP
-       , tableP
+       , getToml
        , tableArrayP
        , inlineTableP
        , tomlP
@@ -56,8 +56,6 @@ tableNameP = between (text "[") (text "]") keyP
 tableArrayNameP :: Parser Key
 tableArrayNameP = between (text "[[") (text "]]") keyP
 
--- Tables
-
 -- | Parser for lines starting with 'key =', either values or inline tables.
 hasKeyP :: Parser (Key, Either AnyValue TOML)
 hasKeyP = (,) <$> keyP <* text "=" <*> eitherP anyValueP inlineTableP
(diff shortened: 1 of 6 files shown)


jiegillet:
Some checks are failing: `htoml-megaparsec` doesn't build on `8.4.4`, not sure what I can do about that.

test/Test/Toml/Parsing/Unit.hs
```diff
@@ -396,49 +396,39 @@ tableSpecs :: Spec
...
-                (makeKey ["table-1"], tomlFromKeyVal [str, int])
+          parseTable "key1 = \"some string\"\nkey2 = 123"
+                (tomlFromKeyVal [str, int])
         it "can parse an empty TOML table" $
-            parseTable "[table]" (makeKey ["table"], tomlFromKeyVal [])
-        it "allows the name of the table to be any valid TOML key" $ do
-            parseTable "[dog.\"tater.man\"]"
-                (makeKey ["dog", dquote "tater.man"], tomlFromKeyVal [])
-            parseTable "[j.\"ʞ\".'l']"
-                (makeKey ["j", dquote "ʞ", squote "l"], tomlFromKeyVal [])
```
chshersh: I don't quite understand why so many tests were deleted :thinking: They seem useful because they test unicode symbols. It's not obvious from the implementation, why these tests are redundant now.

src/Toml/Parser/TOML.hs
```diff
@@ -68,69 +68,58 @@ inlineTableP = between
...
     (table, array) <- fmap distributeEithers
         $ many
-        $ childKeyP key
-        $ eitherPairP (try tableP) tableArrayP
+        $ eitherPairP
+            (try $ do
+              (kDiff, k) <- childKeyP key tableNameP
+              t <- tableP (Just k)
+              pure (kDiff, t))
+            (do
```
chshersh: I think it's better for code readability to introduce two local variables under `let` with types and names like `tomlTableP` and `tomlTableArrayP` and use them as arguments for `eitherPairP` instead of manually wrapping `do` blocks in `()`.

src/Toml/Parser/TOML.hs
```diff
@@ -68,69 +68,58 @@ inlineTableP = between
...
+tableP :: Maybe Key -> Parser TOML
+tableP key = do
     (val, inline)  <- distributeEithers <$> many hasKeyP
     (table, array) <- fmap distributeEithers
         $ many
-        $ childKeyP key
-        $ eitherPairP (try tableP) tableArrayP
+        $ eitherPairP
+            (try $ do
+              (kDiff, k) <- childKeyP key tableNameP
```
chshersh: It looks like this parser can hit another corner case because it has backtracking due to nested `try` combinators inside self-recursive calls. My understanding is that you should put `try` as deep as possible. I see that `childKeyP` already has `try`. Would the tests still pass if you just remove that `try` from the line above (88)?

src/Toml/Parser/TOML.hs
```diff
@@ -68,69 +68,58 @@ inlineTableP = between
...
-childKeyP :: Key -> Parser (Key, a) -> Parser (Key, a)
-childKeyP key p = try $ do
-    (k, x) <- p
+-- | @childKeyP (Just key) p@ checks if the result of @p@ if a child key of
+-- @key@ and returns the difference of the keys and the child key.
+-- @childKeyP Nothing p@ is only called from @tomlP@ (no parent key).
+childKeyP :: Maybe Key -> Parser Key -> Parser (Key, Key)
+childKeyP Nothing parser = try $ do
+  k <- parser
+  pure (k, k)
```
chshersh: I would rewrite it as `try $ (\x -> (x, x) <$> parser`. I see why you did it this way: you prefer symmetry in pattern-matching. In that case, you can leave the implementation as it is, but you should use 4 spaces according to our style guide.

src/Toml/Parser/TOML.hs
jiegillet: You're right, I'll change it

src/Toml/Parser/TOML.hs
jiegillet: I'll check!

src/Toml/Parser/TOML.hs
jiegillet: I don't mind your suggestion, I didn't particularly like my way of writing it. And I'll re-check the spacing everywhere.

chshersh:
@jiegillet Ci should pass now. I cleaned the cache and restarted it. Not sure what's happening here exactly and why this error happens.

test/Test/Toml/Parsing/Unit.hs
jiegillet: I had to remove them because I've refactored the key parsing out of the table parsing function. Those exact same tests on unicode keys are tested somewhere else, so they were redundant anyway.
I will add more tests like you suggested though. There are also some corner cases that I should add.

jiegillet:
Sorry I ended up moving a lot of things around in tests. I ditched all the partial tables and arrays tests in favor of full on `tomlP` tests. I didn't remove any test, and I added one with many arrays.

PR #200, opened 2019-06-28
[#154] Implement generic codec

### LINKED ISSUE #154: Implement codec for `Generic`

chshersh:
Resolves #154

I've come with the implementation of `Generic` codecs. Should help with removing boilerplate. See how it's used in `seaweed`:

* https://github.com/kowainik/seaweed/pull/15

TODO:

- [x] Write usage example
- [x] Write better documentation
- [ ] Write some tests

### DIFF
--- a/src/Toml/Generic.hs
+++ b/src/Toml/Generic.hs
@@ -307,7 +307,7 @@ instance (HasItemCodec a, Typeable a) => HasItemCodec [a] where
         Right codec -> Right $ Toml.list codec (fromString $ typeName @a)
 
 {- | Helper typeclass for generic deriving. This instance tells how the data
-type should be coded if it's a field of other data type.
+type should be coded if it's a field of another data type.
 
 __NOTE:__ You can use this typeclass for writing your custom codecs manually but
 this is less explicit and generally not encouraged. Implement instances of this
--- a/src/Toml/Generic.hs
+++ b/src/Toml/Generic.hs
@@ -73,7 +73,7 @@ socialCodec = Social
 @
 
 However, if you derive 'Generic' instance for your data types (as we do in the
-example), you can write your codecs simpler.
+example), you can write your codecs in a simpler way.
 
 @
 userCodec :: 'TomlCodec' User
--- a/src/Toml/Generic.hs
+++ b/src/Toml/Generic.hs
@@ -92,7 +92,7 @@ Several notes about the interface:
 generic codecs).
 2. If you have a custom data type as a field of another type, you need to implement
 the instance of the 'HasCodec' typeclass.
-3. If data type appears as an element of a list, you need to implement instance
+3. If the data type appears as an element of a list, you need to implement the instance
 of the 'HasItemCodec' typeclass.
 -}
 
--- a/src/Toml/Generic.hs
+++ b/src/Toml/Generic.hs
@@ -72,7 +72,7 @@ socialCodec = Social
     \<*\> Toml.text "link" .= link
 @
 
-However, if you derived 'Generic' for your data types (as we did in the
+However, if you derive 'Generic' instance for your data types (as we do in the
 example), you can write your codecs simpler.
 
 @
--- a/src/Toml/Generic.hs
+++ b/src/Toml/Generic.hs
@@ -9,7 +9,92 @@
 {-# LANGUAGE TypeSynonymInstances #-}
 {-# LANGUAGE UndecidableInstances #-}
 
--- | This module contains implementation of the 'Generic' TOML codec.
+{- | This module contains implementation of the 'Generic' TOML codec. If your
+data types are big and you want to have codecs for them without writing a lot of
+boilerplate code, you can find this module helpful. Below you can find detailed
+explanation on how 'Generic' codecs work.
+
+Consider the following Haskell data types:
+
+@
+__data__ User = User
+    { age     :: Int
+    , address :: Address
+    , socials :: [Social]
+    } __deriving__ ('Generic')
+
+__data__ Address = Address
+    { street :: Text
+    , house  :: Int
+    } __deriving__ ('Generic')
+
+__data__ Social = Social
+    { name :: Text
+    , link :: Text
+    } __deriving__ ('Generic')
+@
+
+Value of the @User@ type represents the following TOML:
+
+@
+age = 27
+
+[address]
+  street = "Miami Beach"
+  house  = 42
+
+[[socials]]
+  name = \"Twitter\"
+  link = "https://twitter.com/foo"
+
+[[socials]]
+  name = \"GitHub\"
+  link = "https://github.com/bar"
+@
+
+Normally you would write 'TomlCodec' for this data type like this:
+
+@
+userCodec :: 'TomlCodec' User
+userCodec = User
+    \<$\> Toml.int "age" .= age
+    \<*\> Toml.table addressCodec "address" .= address
+    \<*\> Toml.list  socialCodec "socials"  .= socials
+
+addressCodec :: 'TomlCodec' Address
+addressCodec = Address
+    \<$\> Toml.text "street" .= street
+    \<*\> Toml.int  "house"  .= house
+
+socialCodec :: 'TomlCodec' Social
+socialCodec = Social
+    \<$\> Toml.text "name" .= name
+    \<*\> Toml.text "link" .= link
+@
+
+However, if you derived 'Generic' for your data types (as we did in the
+example), you can write your codecs simpler.
+
+@
+userCodec :: 'TomlCodec' User
+userCodec = 'genericCodec'
+
+__instance__ 'HasCodec' Address __where__
+    hasCodec = Toml.table 'genericCodec'
+
+__instance__ 'HasItemCodec' Social __where__
+    hasItemCodec = Right 'genericCodec'
+@
+
+Several notes about the interface:
+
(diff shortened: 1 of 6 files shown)


jiegillet:
That's really cool, well done.
While playing around, the first thing I wanted to test on was a record with a `String` but it's not supported, we should probably have that. 

I also found something strange?
```
ghci> Toml.encode genericCodec (True, True)
" = true\n"
```

Then I tried with `testT` from `Playground.hs` but `Either` is not supported. Maybe the way  `testT` handles it is good, encoding: `either.Left` and `either.Right`. Or maybe that should remain a choice from the user? If so, how do we force that interpretation of `testE1` into `testT`?

I'll keep testing things, but I have to stop now, so I'm posting this already.

chshersh:
@jiegillet 

> I also found something strange?

That's because there is a `Generic` instance for a pair. TOML currently doesn't support tuples, it shouldn't be that hard to throw compiler error when trying to decode a pair, but I'll leave it to another issue. I want to finish implementation (with docs and tests) and release `tomland-1.1.0.0` soon and create a bunch of issues about further improvements.

> but Either is not supported

Arbitrary sum types are not supported. Moreover, for `Either` there are at least two possible ways to decode it, and it's not clear which one to choose. So if you have sum types in your configuration, for now, you need to write your codec manually.

jiegillet:
> You shouldn't use `String` anyways

That sounds like rationalizing but I understand the technical difficulties.

> That's because there is a `Generic` instance for a pair.

> Arbitrary sum types are not supported. 

I think it's worth thinking about providing a default implementation of sum and product types, as long as they can be overwritten.  This is what I had done in #157 actually really. 
For example `data A = A Type1 Type2` could be be encoded `a = [[type1], [type2]]`, that's the simplest way I can think of to encode a pair in `TOML`. And the sum type `data X = A | B Int | C Bool Int` could be `x = "a"` or `x = 0` or `x = [[true], [0]]`.  But that's fine for the future (or never, if you don't like the idea).

I managed to make the `Test` type encode with `genericCodec` but it took me a while, and I had to remove the `Either` type from there because I really couldn't think of a way to include them. It all works fine, but the documentation will be very important here.

chshersh:
This PR is finished. Please, review, if you have time :slightly_smiling_face: After this PR is merged, I will prepare next `tomland` release and create issues regarding further improvements.

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE GADTs                #-}
+{-# LANGUAGE PolyKinds            #-}
+{-# LANGUAGE Rank2Types           #-}
+{-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+{- | This module contains implementation of the 'Generic' TOML codec. If your
+data types are big and you want to have codecs for them without writing a lot of
```
vrom911:
```suggestion
data types are big and nested, and you want to have codecs for them without writing a lot of
```

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+boilerplate code, you can find this module helpful. Below you can find detailed
```
vrom911:
```suggestion
boilerplate code, you can find this module helpful. Below you can find the detailed
```

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+explanation on how 'Generic' codecs work.
```
vrom911:
```suggestion
explanation on how the 'Generic' codecs work.
```

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+    \<$\> Toml.text "street" .= street
+    \<*\> Toml.int  "house"  .= house
+
+socialCodec :: 'TomlCodec' Social
+socialCodec = Social
+    \<$\> Toml.text "name" .= name
+    \<*\> Toml.text "link" .= link
+@
+
+However, if you derived 'Generic' for your data types (as we did in the
```
vrom911:
```suggestion
However, if you derive 'Generic' instance for your data types (as we do in the
```

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+example), you can write your codecs simpler.
```
vrom911:
```suggestion
example), you can write your codecs in a simpler way.
```

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+__instance__ 'HasCodec' Address __where__
+    hasCodec = Toml.table 'genericCodec'
+
+__instance__ 'HasItemCodec' Social __where__
+    hasItemCodec = Right 'genericCodec'
+@
+
+Several notes about the interface:
+
+1. Your top-level data types is always implemented as 'genericCodec' (or other
```
vrom911:
```suggestion
1. Your top-level data types are always implemented as 'genericCodec' (or other
```

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+generic codec).
```
vrom911:
```suggestion
generic codecs).
```

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+2. If you have custom data type as a field of other type, you need to implement
```
vrom911:
```suggestion
2. If you have a custom data type as a field of another type, you need to implement
```

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+instance of the 'HasCodec' typeclass.
```
vrom911:
```suggestion
the instance of the 'HasCodec' typeclass.
```

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+3. If data type appears as an element of a list, you need to implement instance
```
vrom911:
```suggestion
3. If the data type appears as an element of a list, you need to implement the instance
```

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+    genericTomlCodec :: GenericOptions -> TomlCodec (f p)
+
+instance GenericCodec f => GenericCodec (D1 d f) where
+    genericTomlCodec = Toml.dimap unM1 M1 . genericTomlCodec
+    {-# INLINE genericTomlCodec #-}
+
+type GenericSumTomlNotSupported =
+    'Text "Generic TOML deriving for arbitrary sum types is not supported currently."
+
+instance (TypeError GenericSumTomlNotSupported) => GenericCodec (f :+: g) where
```
vrom911: This is useful!

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+{- | If data type @a@ is not primitive then this instance returns codec for list
+under key equal to @a@ type name.
+-}
+instance (HasItemCodec a, Typeable a) => HasItemCodec [a] where
+    hasItemCodec = case hasItemCodec @a of
+        Left prim   -> Left $ Toml._Array prim
+        Right codec -> Right $ Toml.list codec (fromString $ typeName @a)
+
+{- | Helper typeclass for generic deriving. This instance tells how the data
+type should be coded if it's a field of other data type.
```
vrom911:
```suggestion
type should be coded if it's a field of another data type.
```

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+
+__NOTE:__ You can use this typeclass for writing your custom codecs manually but
```
vrom911: You mean, you can write instances for any custom data types, but this is not recommended way? The alternative is using one of the generic codecs?

src/Toml/Generic.hs
```diff
@@ -0,0 +1,390 @@
...
+data Person = Person
+    { personName    :: Text
+    , personAddress :: Address
+    } deriving (Generic)
+
+data Address = Address
+    { addressStreet :: Text
+    , addressHouse  :: Int
+    } deriving (Generic)
+      deriving HasCodec via TomlTable Address
```
vrom911: This is going to be cool!

src/Toml/Generic.hs
chshersh: What I meant actually, is that instead of writing:

```haskell
userCodec :: TomlCodec User
userCodec = User
    <$> Toml.text "name" .= userName
    <*> Toml.int  "age"  .= userAge
```

You can write

```
userCodec :: TomlCodec User
userCodec = User
    <$> Toml.hasCodec "name" .= userName
    <*> Toml.hasCodec "age"  .= userAge
```

But it's not recommended. Probably should write this in a more clear way :slightly_smiling_face:

chshersh:
@vrom911 Thanks for your review :heart: I've updated documentation :+1:

src/Toml/Generic.hs
chshersh: Updated docs for this typeclass.

Issue #206, opened 2019-07-10
Test on `toml-v0.4.0` example


leptonyu:
```toml
################################################################################
## Inline Table

# Inline tables provide a more compact syntax for expressing tables. They are
# especially useful for grouped data that can otherwise quickly become verbose.
# Inline tables are enclosed in curly braces `{` and `}`. No newlines are
# allowed between the curly braces unless they are valid within a value.

[table.inline]

name = { first = "Tom", last = "Preston-Werner" }
point = { x = 1, y = 2 }
```

My tests has the above inline block, it fails by using `tomland-1.1.0.0`

````
uncaught exception: IOException of type UserError
user error (ParseException "61:8:\n   |\n61 | name = { first = \"Tom\", last = \"Preston-Werner\" }\n   |        ^\n\"name\" :| [] is not a child key of \"table\" :| [\"inline\"]\n")
````

**Comment by chshersh** [comment]
@leptonyu Thanks for reporting the issue! That was unexpected 😕 We will improve our test-suite and fix the bug. Could you also provide the link to your tests, so we can test before releasing a new version to Hackage?

/cc @jiegillet

**Comment by leptonyu** [comment]
My tests file was copied from the offical toml-lang repo. `example-v0.4.0.toml`

Here are the link.
[salak.toml](https://github.com/leptonyu/salak/blob/master/salak-toml/test/salak.toml)

**Comment by leptonyu** [comment]
[Official toml file link](https://github.com/toml-lang/toml/blob/master/examples/example-v0.4.0.toml)

**Comment by chshersh** [comment]
@leptonyu Nice to have an example! I've changed the title of this issue to implement more robust testing.

**Comment by jiegillet** [comment]
Sorry about that. For this specific issue, adding a `try` at the right place fixes it. I'll do a PR, I should also probably add all official examples and make sure they are fine.

I have one uncertainty about this situation though. Should 
```toml
[table.inline]
name = { first = "Tom", last = "Preston-Werner" }
point = { x = 1, y = 2 }
```
be interpreted as one table `table.inline` containing two sub tables `name` and `point`, or should it be considered as three separate tables on the same level? The official documentation says:

> ```toml
> name = { first = "Tom", last = "Preston-Werner" }
> point = { x = 1, y = 2 }
> animal = { type.name = "pug" }
> ```
> The inline tables above are identical to the following standard table definitions:
> ```toml
> [name]
> first = "Tom"
> last = "Preston-Werner"
> 
> [point]
> x = 1
[...]

Issue #198, opened 2019-06-24
Small bugs with Gen.hs


jiegillet:
Some bugs I found when running tests:

- [x] The string `"\"` can be generated as a key, but is invalid because the escape character cannot be on its own
- [ ] It can generate  an array `a` which contains two arrays `b` and `b.c` on the same level, which pretty prints to `[[a]] \n [[a.b]] \n [[a.b.c]]` but that gets (correctly) interpreted as an array `a` that contains  an array `b` that contains an array `c`. Elements of an array can't share prefixes.

The first one should be easy to fix, the second is trickier maybe, but overall those have very small chance of appearing, and if they do, re-running the test has always done the trick for me. Maybe not worth fixing.

chshersh:
@jiegillet Let's fix the first problem because it's the easiest one. And keep the issue open. Maybe somebody would like to practice in implementing generators :slightly_smiling_face:

chshersh:
The first part was fixed. The second part will require more effort to fix and it doesn't cause any troubles so far, so I'm closing the issue.

PR #208, opened 2019-07-10
 [#206] Test on toml-v0.4.0 example

jiegillet:
Fix for #206. I apologize for letting that slip through me to begin with.

I don't have much experience with `hspec`, please let me know if I did something strange.


jiegillet:
One of the checks failed because of the issue described in #198 (same-level tables with common subkeys)

CHANGELOG.md
```diff
@@ -3,6 +3,10 @@
 tomland uses [PVP Versioning][1].
 The changelog is available [on GitHub][2].
 
+## 1.1.1.0
```
chshersh: Let's do a minor version bump and today's date. I will release it to Hackage after this PR is merged
```suggestion
## 1.1.0.1 — Jul 10, 2019
```

test/Test/Toml/Parsing/Examples.hs
```diff
@@ -0,0 +1,26 @@
...
+    mapM_ example files
+
+example :: FilePath -> Spec
+example file = it ("can parse file " ++ file) $
+      (isRight . parse <$> TIO.readFile ("test/examples/" ++ file))
+      `shouldReturn` True
+
+files :: [FilePath]
+files =
+    [ "fruit.toml"
```
chshersh: I propose to use `listDirectory` function inside tests to avoid hardcoding of test cases:

* https://hackage.haskell.org/package/directory-1.3.4.0/docs/System-Directory.html#v:listDirectory

You can use `runIO` function from `hspec` to run `IO` actions inside `Spec` monad:

* https://hackage.haskell.org/package/hspec-2.7.1/docs/Test-Hspec.html#v:runIO

jiegillet: I looked into `runIO` when I was figuring out how to read files in a `Spec` but then I realized `it` can handle it just fine, I think I'll leave it like this.

jiegillet: Never mind, I guess it's necessary if I want to use `listDirectory`. On it.

Issue #211, opened 2019-08-07
Parser does not error when there are duplicate keys


stephenmac7:
The TOML `"x = 1\nx = 2"` is parsed as `"x = 2"`, but it should error according to the TOML spec.

This is actually a pretty large issue because misspellings interact badly with this behavior. For example,
```
[[item]]
  [[item.sub]]
    a = 1
[[item]]
  [[item.sub]]
    a = 2
  [[ite.sub]]
    a=4
[[item]]
  [[item.sub]]
    a = 3
```
parses as
```
[[ite.sub]]
  a = 4

[[item]]
  [[item.sub]]
    a = 3
```

which is quite different from what we would like.

jiegillet:
Yes, this is basically #5, we haven't implemented any checks yet. If @chshersh is OK with it, I'm happy to get started on that. 

The examples you give are very relevant, thank you.

chshersh:
@stephenmac7 Thanks for providing a convincing example for this feature! As @jiegillet mentioned, this is indeed a part of TOML validation, which hasn't been implemented yet. Parsing TOML (as far as `tomland` concerned) is a multi-step process which includes lexing, parsing, validation and decoding. I see that @jiegillet already opened a PR. Will look into this soon.

chshersh:
I'm currently slowly working on refactoring the parser and implementing validation. But I want to point out that _validation_ is not the only solution to this particular problem. The alternative solution is to implement TOML difference (#186). So you can have warnings of unused fields. And the result of parsing TOML will be a data type like this:

```haskell
data Result a
    = Failure String
    | Warning [String] a
    | Success a
```

chshersh:
@stephenmac7 I've implemented validation for duplicate keys in #216. However, I realised that your problem is not in duplicating keys. And checking for duplicating keys won't help you to solve your problem. What you need is TOML difference implementation. So I'm closing this issue as a duplicate of #186.

PR #212, opened 2019-08-09
[#5] TOML Validation [WIP]

jiegillet:
This is a proof of concept to address #5 and #211 (it is already enough to fix 211 in its current form).

This is quite a big issue so I want to make sure this approach is OK before working too much on it. The basic idea is that I'm adding a `failure :: Maybe String` field to `TOML` and tweaks its `(<>)`
definition so that it detects invalid syntax there. If there is a failure, the parsing will fail and return an error.

What do you think?

It passes all test and here are benchmarks:

master:
```
benchmarked Parse/tomland
time                 354.0 μs   (349.8 μs .. 360.0 μs)
                     0.998 R²   (0.996 R² .. 1.000 R²)
mean                 358.5 μs   (355.7 μs .. 362.4 μs)
std dev              11.86 μs   (8.058 μs .. 17.37 μs)
variance introduced by outliers: 15% (moderately inflated)
```

with validation:
```
benchmarked Parse/tomland
time                 376.4 μs   (369.0 μs .. 386.3 μs)
                     0.996 R²   (0.992 R² .. 0.999 R²)
mean                 377.4 μs   (373.9 μs .. 383.4 μs)
std dev              15.26 μs   (11.03 μs .. 22.16 μs)
variance introduced by outliers: 20% (moderately inflated)
```

### DIFF
--- a/src/Toml/Parser/TOML.hs
+++ b/src/Toml/Parser/TOML.hs
@@ -97,7 +97,11 @@ tableArrayP key =
 
 -- | Parser for a '.toml' file
 tomlP :: Parser TOML
-tomlP = sc *> localTomlP Nothing <* eof
+tomlP = do
+    toml <- sc *> localTomlP Nothing <* eof
+    case failure toml of
+        Nothing  -> pure toml
+        Just err -> fail err
 
 -- | Parser for a toml under a certain key
 localTomlP :: Maybe Key -> Parser TOML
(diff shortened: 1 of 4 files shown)


chshersh:
@jiegillet Thanks for your effort on working on this issue! Validation is hard, and it's unfortunate, that it decreases the performance, but correctness is more important. Also, I'm not entirely happy with adding `failure` to the `TOML` data type because `TOML` in my vision should represent TOML AST, not the state of parsing.

I see that TOML parser in `tomland` becomes more complicated. For quite some time I was thinking about how to make it simpler, faster and support validation as well. I have some ideas now. Will try to put them in a separate PR.

jiegillet:
Alright, looking forward to seeing your ideas.

chshersh:
@jiegillet I've implemented an alternative solution in #216 (including some validation). So I'm closing this PR in favour of another one.

PR #220, opened 2019-10-01
[#219] Add inline pragma

### LINKED ISSUE #219: Add INLINE pragmas to code
It seems that the performance of the library can be improved by adding `{-# INLINE #-}` pragmas to various small functions. So it would be nice to walk through the whole library code and add these pragmas where they seem reasonable.

willbasky:
Resolves #219


chshersh:
@willbasky Thanks for cleaning some unused imports so far :slightly_smiling_face:

willbasky:
I try to add INLINE somewhere. It improved speed on 10-20 μs for Parse section at benchmark.

willbasky:
Now mean for Decode became as before and sometimes better.

src/Toml/Bi/Combinators.hs
```diff
@@ -1,6 +1,9 @@
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE GADTs               #-}
+-- {-# LANGUAGE RankNTypes          #-}
+-- {-# LANGUAGE LambdaCase          #-}
+-- {-# LANGUAGE RecordWildCards          #-}
```
chshersh: I think you can delete these extensions instead of commenting.

src/Toml/Bi/Map.hs
```diff
@@ -125,12 +124,14 @@ data BiMap e a b = BiMap
...
 
     (.) :: BiMap e b c -> BiMap e a b -> BiMap e a c
     bc . ab = BiMap
         { forward  =  forward ab >=>  forward bc
         , backward = backward bc >=> backward ab
         }
+    {-# INLINE (.) #-}
 
 -- | Inverts bidirectional mapping.
 invert :: BiMap e a b -> BiMap e b a
```
chshersh: I think `invert`, `iso` and `prism` functions should have INLINE pragmas as well.

src/Toml/Bi/Map.hs
```diff
@@ -417,24 +433,28 @@ instances.  Usually used as 'Toml.Bi.Combinators.enumBounded' combinator.
...
 
 {- | 'Float' bimap for 'AnyValue'. Usually used as
 'Toml.Bi.Combinators.float' combinator.
 -}
 _Float :: TomlBiMap Float AnyValue
 _Float = iso realToFrac realToFrac >>> _Double
+{-# INLINE _Float #-}
 
 -- | Helper bimap for 'Data.Text.Text' and strict 'ByteString'
 _ByteStringText :: TomlBiMap ByteString Text
```
chshersh: I think it's actually a good idea to inline `_ByteStringText` as well because optimizations on ByteString and Text functions are relying on inlining and can be huge in some cases.

src/Toml/Bi/Map.hs
```diff
@@ -506,3 +529,4 @@ _IntSet = iso IS.toList IS.fromList >>> _Array _Int
```
chshersh: I think `_Set`, `_HashSet` and `_IntSet` deserve the pragma as well :slightly_smiling_face:

src/Toml/Bi/Map.hs
```diff
@@ -492,17 +520,21 @@ _NonEmptyArray = BiMap
 -- as an array. Usually used as 'Toml.Bi.Combinators.arraySetOf' combinator.
```
vrom911: Should we inline `_NonEmptyArray` (above) as well?

PR #221, opened 2019-10-01
Cleanup some metadata

chshersh:
Cleanup some metadata:

1. Use `--system-ghc` for `stack` builds.
2. Missing deriving strategies now produce a compiletime error to make sure they are used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ script:
     if [ -z "$STACK_YAML" ]; then
       cabal test --enable-tests
     else
-      stack test --no-terminal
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -46,7 +46,7 @@ script:
     if [ -z "$STACK_YAML" ]; then
       cabal test --enable-tests
     else
-      stack test --no-terminal
+      stack test --no-terminal --system-ghc
```
willbasky: What `--system-ghc` gives?

willbasky: That is great!

Issue #201, opened 2019-07-08
Add combinators `set / hashSet` for array of tables


chshersh:
Currently, we have `list` and `nonEmpty` combinators for coding array of tables. It would be nice to have `set` and `hashSet` as well.

SanchayanMaity:
@chshersh I did like to help with this.

chshersh:
@SanchayanMaity Sure, go for it :+1:

SanchayanMaity:
@chshersh  Can you point me in the right direction a bit?

set :: forall a . TomlCodec a -> Key -> TomlCodec (Set a)
hashSet :: forall a . TomlCodec a -> Key -> TomlCodec (HashSet a)

Looking at the existing code, would the above be the correct type signatures? The initial thought I had for *set* was to mimic the *list* implementation but instead of fmapping toList, map (fromList . toList) however this fails the Ord constraint required for *Set*. Not sure if I am on the right train of thought.

chshersh:
@SanchayanMaity You're in the right direction :+1: The types of `set` and `hashSet` along with the implementation are correct. You only need to add required constraints as compiler tells you. So for `set` the type signature will look like this:

```haskell
set :: forall a . Ord a => TomlCodec a -> Key -> TomlCodec (Set a)
```

If you want to put elements in the `Set` container, you need to be able to compare them.

SanchayanMaity:
@chshersh 

```haskell
set :: forall a . Ord a => TomlCodec a -> Key -> TomlCodec (Set a)
set codec key = Codec
    { codecRead = ((Set.fromList . toList) <$> codecRead nonEmptyCodec) `catchError` \case
        TableNotFound errKey | errKey == key -> pure Set.empty
        err -> throwError err
    , codecWrite = \as -> do
        let tomls = (Set.fromList . toList) $ fmap (execTomlCodec codec) ((NE.fromList . Set.toList) as)
        mTables <- gets $ HashMap.lookup key . tomlTableArrays

        let newTomls = case mTables of
                Nothing       -> tomls
                Just oldTomls -> ((Set.fromList . toList) oldTomls) <> tomls

        as <$ modify (insertTableArrays key ((NE.fromList . Set.toList) newTomls))
    }
  where
    nonEmptyCodec :: TomlCodec (NonEmpty a)
    nonEmptyCodec = nonEmpty codec key
```

I came up with this which does not look clean and fails to compile. What stymies me is the error.

```haskell
[...]

PR #222, opened 2019-10-03
Add combinators set and hashSet for array of tables

SanchayanMaity:
Implements #201 

Signed-off-by: Sanchayan Maity <maitysanchayan@gmail.com>

### DIFF
--- a/examples/Playground.hs
+++ b/examples/Playground.hs
@@ -1,11 +1,16 @@
 {-# OPTIONS -Wno-unused-top-binds #-}
+{-# LANGUAGE DeriveAnyClass #-}
 
 module Main (main) where
 
 import Control.Applicative ((<|>))
 import Control.Arrow ((>>>))
+import Data.Hashable (Hashable)
+import Data.HashSet (HashSet)
+import Data.Set (Set)
 import Data.Text (Text)
 import Data.Time (fromGregorian)
+import GHC.Generics (Generic)
 
 import Toml (ParseException (..), TomlCodec, pretty, (.=), (<!>))
 import Toml.Edsl (mkToml, table, (=:))
@@ -22,7 +27,8 @@ insideCodec = Toml.dimap unInside TestInside $ Toml.text "inside"
 data User = User
     { userName :: !Text
     , userAge  :: !Int
-    }
+    } deriving stock (Eq, Ord, Generic)
+      deriving anyclass (Hashable)
 
 userCodec :: TomlCodec User
 userCodec = User
@@ -53,6 +59,8 @@ data Test = Test
     , testE1 :: !(Either Integer String)
     , testE2 :: !(Either String Double)
     , users  :: ![User]
+    , susers :: !(Set User)
+    , husers :: !(HashSet User)
     }
 
 
@@ -71,6 +79,8 @@ testT = Test
     <*> eitherT1 .= testE1
     <*> eitherT2 .= testE2
     <*> Toml.list userCodec "user" .= users
+    <*> Toml.set userCodec "suser" .= susers
+    <*> Toml.hashSet userCodec "huser" .= husers
   where
     -- different keys for sum type
     eitherT1 :: TomlCodec (Either Integer String)
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -305,7 +305,9 @@ list codec key = Codec
 -- | 'Codec' for set of values. Represented in TOML as array of tables.
 set :: forall a . Ord a => TomlCodec a -> Key -> TomlCodec (Set a)
 set codec key = dimap S.toList S.fromList (list codec key)
+{-# INLINE set #-}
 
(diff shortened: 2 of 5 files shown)


src/Toml/Bi/Combinators.hs
```diff
@@ -63,7 +65,7 @@ import Data.IntSet (IntSet)
 import Data.List.NonEmpty (NonEmpty (..), toList)
 import Data.Maybe (fromMaybe)
 import Data.Semigroup ((<>))
-import Data.Set (Set)
+import Data.Set (Set, fromList)
```
chshersh: Importing `fromList` seems redundant because you're using qualified `S.fromList`
```suggestion
import Data.Set (Set)
```

src/Toml/Bi/Combinators.hs
```diff
@@ -298,3 +302,15 @@ list codec key = Codec
   where
     nonEmptyCodec :: TomlCodec (NonEmpty a)
     nonEmptyCodec = nonEmpty codec key
+
+set :: forall a . Ord a => TomlCodec a -> Key -> TomlCodec (Set a)
+set codec key = dimap (NE.fromList . S.toList) (S.fromList . NE.toList) nonEmptyCodec
```
chshersh: You can use `list` codec instead of `nonEmpty` to avoid manual conversion from/to `NonEmpty`. Also, `fromList` is an unsafe function and can crash in runtime.

src/Toml/Bi/Combinators.hs
```diff
@@ -298,3 +302,15 @@ list codec key = Codec
...
+            where
+              nonEmptyCodec :: TomlCodec (NonEmpty a)
+              nonEmptyCodec = nonEmpty codec key
+
+hashSet :: forall a . (Hashable a, Eq a) => TomlCodec a -> Key -> TomlCodec (HashSet a)
```
chshersh: It would be also nice to add some documentation to these functions. At least in the spirit of the `list` function.

SanchayanMaity: Will go through the existing docs and add similar.

src/Toml/Bi/Combinators.hs
SanchayanMaity: You mean I shouldn't use the S.fromList function or was that for NE?

Ah I missed seeing that I could use list codec.

chshersh: You shouldn't use `NE.fromList`.

chshersh:
Oh, I almost forgot. Could you also please add two instances in the `Toml.Generic` module below lists for `Set` and `HashSet` as well since we now have codecs for them?

https://github.com/kowainik/tomland/blob/3c323e14e8b344c7cf2340b5ca32b44a1ab15925/src/Toml/Generic.hs#L340-L348

examples/Playground.hs
```diff
@@ -1,11 +1,17 @@
 {-# OPTIONS -Wno-unused-top-binds #-}
+{-# LANGUAGE DeriveGeneric  #-}
```
chshersh: `DeriveGeneric` is already enabled in the `default-extensions`, so need to enable it in this module :slightly_smiling_face:

examples/Playground.hs
```diff
@@ -71,6 +80,8 @@ testT = Test
     <*> eitherT1 .= testE1
     <*> eitherT2 .= testE2
     <*> Toml.list userCodec "user" .= users
+    <*> Toml.set userCodec "user" .= susers
+    <*> Toml.hashSet userCodec "user" .= husers
```
chshersh: Thanks for updating the example! But I believe that this codec is not completely valid because it populates multiple fields from a single key but they all output to the same key which breaks roundtrip property. You can just change key names to `suser` and `huser` here as well to be more correct.

src/Toml/Bi/Combinators.hs
```diff
@@ -298,3 +301,17 @@ list codec key = Codec
   where
     nonEmptyCodec :: TomlCodec (NonEmpty a)
     nonEmptyCodec = nonEmpty codec key
+
+-- | 'Codec' for set of values. Represented in TOML as array of tables.
+set :: forall a . Ord a => TomlCodec a -> Key -> TomlCodec (Set a)
+set codec key = dimap S.toList S.fromList listCodec
+            where
+              listCodec :: TomlCodec [a]
+              listCodec = list codec key
```
chshersh: I think that having a separate variable under `where` is not neccessary here, and you can write the implementation in a single line:

```haskell
set codec key = dimap S.toList S.fromList (list codec key)
```

SanchayanMaity:
> add `{-# INLINE #-}` pragmas to `set` and `hashSet` combinators

I thought you meant the inline pragma work got merge in master but I do not see that. I added a separate commit for adding inline pragma. Assuming that's what you meant.

chshersh:
@SanchayanMaity yes, that's exactly what I meant, you understood me 100% correctly :slightly_smiling_face: Sorry for being unclear.

Issue #204, opened 2019-07-08
Implement `_ByteStringArray` bimap


chshersh:
To decode/encode `ByteString` as array of non-negative integers. Same for `LByteString`.

PR #224, opened 2019-10-04
[#204] Add _ByteStringArray bimap

crtschin:
Should satisfy #204

### DIFF
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -21,6 +21,8 @@ module Toml.Bi.Combinators
        , lazyText
        , byteString
        , lazyByteString
+       , byteStringArray
+       , lazyByteStringArray
        , string
          -- ** Time types
        , zonedTime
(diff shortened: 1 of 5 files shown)


src/Toml/Bi/Map.hs
```diff
@@ -460,6 +462,29 @@ _LByteStringText = prism (TL.encodeUtf8 . TL.fromStrict) eitherText
...
+    eitherNonNeg xs
+      | all (>= 0) xs = Right $ map fromIntegral xs
+      | otherwise     = Left $
+          ArbitraryError ("Negative numbers not allowed when converting "
+              <> tShow xs
+              <> " to a ByteString")
+
+-- | Helper bimap for a 'Integer' list and strict 'ByteString'
+_ByteStringArray :: TomlBiMap ByteString [Int]
+_ByteStringArray = iso unpack pack >>> _Word8ArrIntArr
```
chshersh: Let's instead add `import qualified Data.ByteString as BS` and use qualified `unpack` and `pack` here. For code readability.

src/Toml/Bi/Map.hs
```diff
@@ -460,6 +462,29 @@ _LByteStringText = prism (TL.encodeUtf8 . TL.fromStrict) eitherText
...
+    eitherNonNeg :: [Int] -> Either TomlBiMapError [Word8]
+    eitherNonNeg xs
+      | all (>= 0) xs = Right $ map fromIntegral xs
+      | otherwise     = Left $
+          ArbitraryError ("Negative numbers not allowed when converting "
+              <> tShow xs
+              <> " to a ByteString")
+
+-- | Helper bimap for a 'Integer' list and strict 'ByteString'
+_ByteStringArray :: TomlBiMap ByteString [Int]
```
chshersh: This bidirectional converter should have type `TomlBiMap ByteString AnyValue` so it can decode TOML values. This is the final goal :slightly_smiling_face: So in the TOML file you can write:

```toml
bytes = [ 96, 20, 40, 50 ]
```

src/Toml/Bi/Map.hs
```diff
@@ -460,6 +462,29 @@ _LByteStringText = prism (TL.encodeUtf8 . TL.fromStrict) eitherText
...
+_ByteStringArray = iso unpack pack >>> _Word8ArrIntArr
+
+-- | Helper bimap for a 'Integer' list and lazy 'ByteString'
+_LByteStringArray :: TomlBiMap BL.ByteString [Int]
+_LByteStringArray = iso BL.unpack BL.pack >>> _Word8ArrIntArr
```
chshersh: The current implementation of `_Word8ArrIntArr` is okay :+1: But I propose to do it differentl:

1. Implement `_Word8` bidirectional map.
2. Create ByteString array mapping by composing current `iso` and `_Array _Word8`

This solution has the following benefits:

1. It reuses more existing code
2. In implements new primitive codec which can be useful for somebody in the future

crtschin: That's a much cleaner approach. Thanks!

test/Test/Toml/BiMap/Property.hs
```diff
@@ -39,6 +39,7 @@ test_BiMaps = pure $ testGroup "BiMap roundtrip tests" $ concat
     , prop "Natural" (testBiMap B._Natural G.genNatural)
     , prop "Int" (testBiMap B._Int G.genInt)
     , prop "Word" (testBiMap B._Word G.genWord)
+    , prop "Word8" (testBiMap B._Word8 G.genWord8)
```
chshersh: Thank you for adding tests! :heart:

src/Toml/Bi/Combinators.hs
```diff
@@ -143,6 +146,11 @@ word :: Key -> TomlCodec Word
 word = match _Word
 {-# INLINE word #-}
 
+-- | Codec for word8 values.
+word8 :: Key -> TomlCodec Word8
+word8 = match _Word8
+{-# INLINE word8 #-}
```
vrom911: I guess this should be reexported too :slightly_smiling_face:

crtschin: Right, missed that one

PR #225, opened 2019-10-05
Avoid to generate text with ending backslash

gabrielelana:
Fixes half of #198 

We were wondering why you didn't use `genPrefixMap` in `genToml`, seems like the second problem stated in #198 it might be related to `genPrefixMap`, are we on the right path?

Proudly made by @sphaso and @gabrielelana at "Open Source Saturday Milano"

[Open Source Saturday](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/)

### DIFF
--- a/test/Test/Toml/Gen.hs
+++ b/test/Test/Toml/Gen.hs
@@ -102,7 +102,7 @@ genVal :: MonadGen m => m V
 genVal = Gen.int (Range.constant 0 256)
 
 -- | Generates random value of 'AnyValue' type.
-genAnyValue :: MonadGen m => m AnyValue
+genAnyValue :: (MonadGen m, GenBase m ~ Identity) => m AnyValue
 genAnyValue = Gen.choice $
     (AnyValue <$> genArray) : noneArrayList
 
@@ -126,7 +126,7 @@ genPiece = Piece <$> Gen.choice [bare, quoted]
     quotedWith c = wrapChar c <$> Gen.text (Range.constant 1 10) (Gen.filter (/= c) notControl)
 
     quoted :: m Text
-    quoted = Gen.choice [quotedWith '"', quotedWith '\'']
+    quoted = Gen.filter invalidEscape $ Gen.choice [quotedWith '"', quotedWith '\'']
 
 genKey :: (MonadGen m, GenBase m ~ Identity) => m Key
 genKey = Key <$> Gen.nonEmpty (Range.constant 1 10) genPiece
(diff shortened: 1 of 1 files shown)


test/Test/Toml/Gen.hs
```diff
@@ -285,8 +285,8 @@ genUniHex8Color = do
     pure . Text.pack $ "\\U" ++ hex
 
 -- | Generates text from different symbols.
-genText :: MonadGen m => m Text
-genText =  fmap Text.concat $ Gen.list (Range.constant 0 256) $ Gen.choice
+genText :: (MonadGen m, GenBase m ~ Identity) => m Text
+genText = Gen.filter invalidEscape $ fmap Text.concat $ Gen.list (Range.constant 0 256) $ Gen.choice
```
chshersh: Issue #198 is about not generating `\\` inside TOML keys. However, I believe that `genText` is used not only for keys but for something else. So filtering here is not exactly correct. Instead, you should filter only when generating TOML keys.

test/Test/Toml/Gen.hs
```diff
@@ -332,7 +332,15 @@ Array [Double (-563397.0197456297),Double (-308866.62837749254),Double (-29555.3
...
     [Gen.choice $ map genArrayFrom noneArrayList]
     [Array <$> Gen.list (Range.constant 0 5) genArray]
+
+-- filters
+
+invalidEscape :: Text -> Bool
+invalidEscape t
+    | t == Text.empty = True
+    | Text.last t == '\\' = False
+    | otherwise = True
```
chshersh: Can you please some documentation to this function and some examples of valid and invalid strings? And ideally with the link to TOML spec where it says that `"\"` cannot be a valid key. I'm afraid that this this solution solves the problem partially. Is the following key valid?

```toml
"\x"
```

I think that no, but it still can be genarated. I'm okay to accept this PR as a partial solution, but let's document some cases of what this function covers and what it doesn't cover :slightly_smiling_face:

test/Test/Toml/Gen.hs
gabrielelana: I think `genText` is not used to generate keys but I also thought that a generic text should not end with a single escape like `"aaa\"`

test/Test/Toml/Gen.hs
gabrielelana: Look at `validText` in the last commit if it's what you asked

test/Test/Toml/Gen.hs
```diff
@@ -332,7 +332,28 @@ Array [Double (-563397.0197456297),Double (-308866.62837749254),Double (-29555.3
...
+    validString ['\\'] = False
+    validString ('\\':'b':s) = validString s
+    validString ('\\':'t':s) = validString s
+    validString ('\\':'n':s) = validString s
+    validString ('\\':'f':s) = validString s
+    validString ('\\':'r':s) = validString s
+    validString ('\\':'"':s) = validString s
+    validString ('\\':'\'':t) = validString t
+    validString ('\\':'\\':t) = validString t
+    validString ('\\':'u':n1:n2:n3:n4:t) = (all Char.isHexDigit [n1, n2, n3, n4]) && validString t
```
chshersh: This function is becoming too complicated... I'm afraid it will require some time to understand what's happening here and make sure that there's no better approach... At this point the previous solution doesn't look that bad :slightly_smiling_face:

gabrielelana: Bummer, I thought it was pretty slick 😂 Anyway, no problem to revert it

gabrielelana:
@chshersh let me know what do you think

PR #229, opened 2019-10-08
[#226] Add dimatch combinator for sum types

### LINKED ISSUE #226: Add `disum` combinator for sum types
Connected to [crocodealer/#9](https://github.com/kowainik/crocodealer/pull/9).

Since the `prism` approach for sum types works with single non-nested values, a new combinator is needed to handle sum types, similar to dimap but accounting for the fact that the value may not be there ( doesn't have the right constructor we need ). So a combinator `disum` could be added to handle that with a type of:

```haskell
disum
    :: (c -> Maybe d) -- ^ Mapper for consumer
    -> (a -> b)       -- ^ Mapper for producer
    -> Codec r w d a  -- ^ Source 'Codec' object
    -> Codec r w c b  -- ^ Target 'Codec' object
```

It's implemented in the pull request above but it would be better here. I'm willing to contribute it.

Nimor111:
Fixes #226 

Hope the example is sensible enough. And if some more documentation should be added, do tell. :)

### DIFF
--- a/src/Toml/Bi/Monad.hs
+++ b/src/Toml/Bi/Monad.hs
@@ -6,6 +6,7 @@ module Toml.Bi.Monad
        , dimap
        , dioptional
        , diwrap
+       , disum
        , (<!>)
        , (.=)
        ) where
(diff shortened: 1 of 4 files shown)


src/Toml/Bi/Monad.hs
```diff
@@ -219,6 +220,51 @@ diwrap
...
+    \<|\> disum matchBar (uncurry Bar) (Toml.table barCodec "bar")
+@
+-}
+disum
+    :: (Functor r, Alternative w)
+    => (c -> Maybe d) -- ^ Mapper for consumer
+    -> (a -> b)       -- ^ Mapper for producer
+    -> Codec r w d a  -- ^ Source 'Codec' object
+    -> Codec r w c b  -- ^ Target 'Codec' object
+disum match ctor codec = Codec
```
chshersh: Excellent implementation :+1: 

I have only one minor request: let's rename `disum` to `dimatch`. After thinking a bit more about this name, I feel like `disum` looks more like some arithmetic operations (because of the popular `sum` function). `dimatch` is a bit more clear name, since we actually perform bidireaction matching on the constructors of the sum type.

PR #232, opened 2019-10-10
Add GitHub Actions using cachix

chshersh:
Let's see how it goes!


chshersh:
@vrom911 It works! I've also added a badge. But it doesn't render properly now. Probably requires this to be merged first...

PR #234, opened 2019-10-12
Prepare for 1.2.0.0 release

chshersh:
* Changelog updated
* Added `@since` annotations
* Documentation is fixed in some places
* Updated to the latest LTS snapshot

### DIFF
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -147,7 +147,10 @@ word :: Key -> TomlCodec Word
 word = match _Word
 {-# INLINE word #-}
 
--- | Codec for word8 values.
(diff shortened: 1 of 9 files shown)


src/Toml/Bi/Combinators.hs
```diff
@@ -207,12 +210,18 @@ lazyByteString :: Key -> TomlCodec BL.ByteString
 lazyByteString = match _LByteString
 {-# INLINE lazyByteString #-}
 
--- | Codec for positive integer array values as 'ByteString'.
+{- | Codec for positive integer array values as 'ByteString'.
+
+@since 1.2.0.0
```
vrom911: This is actually pretty useful! Great idea! Need to make it from now on :+1:

PR #236, opened 2019-10-21
[#217] Use TOML Edsl, remove redundant utility functions in tests

### LINKED ISSUE #217: Refactor `Test.Toml.Parsing.Unit`
587 lines for a testing file is insane. Also it's a bid hard to work with the existing tests. The following refactoring should be performed to make existing and new tests nicer:

- [ ] Use eDSL to construct test `TOML` values (we've implemented it for this purpose)
- [ ] Use `neat-interpolation` to construct testing strings where possible (or at least `unlines`/`intercalate`)
- [x] Split into multiple modules: `Test.Toml.Parser.{Unit, Unit.Boo, Unit.Integer, ...}`

Better names for testing modules are appreciated!

Nimor111:
Fixes #217. 

This PR is a refactoring of the `Test.Toml.Parsing.Unit.Toml` tests to use the TOML Edsl in the library. 
Sadly I couldn't figure out how to construct the array of arrays as written in the TODO in the code. A lot of time went into wondering whether the Edsl allows array values with more than one property and I couldn't see how. :/

### DIFF
--- a/test/Test/Toml/Parsing/Unit/Common.hs
+++ b/test/Test/Toml/Parsing/Unit/Common.hs
@@ -1,8 +1,42 @@
 {-# LANGUAGE FlexibleContexts #-}
 
-module Test.Toml.Parsing.Unit.Common where
+module Test.Toml.Parsing.Unit.Common
+       ( parseX
+       , failOn
+       , parseArray
+       , parseBool
+       , parseDouble
+       , parseInteger
+       , parseKey
+       , parseText
+       , parseToml
+       , parseDateTime
+       , arrayFailOn
+       , boolFailOn
+       , integerFailOn
+       , dateTimeFailOn
+       , doubleFailOn
+       , textFailOn
+       , tomlFailOn
+       , quoteWith
+       , squote
+       , dquote
+       , squote3
+       , dquote3
+       , makeKey
+       , makeOffset
+       , makeZoned
+       , int1
+       , int2
+       , int3
+       , int4
+       , offset710
+       , offset0
+       , day1
+       , day2
+       , hours1
+       ) where
 
-import Data.List.NonEmpty (NonEmpty (..))
 import Data.Semigroup ((<>))
 import Data.Text (Text)
 import Data.Time (Day, LocalTime (..), TimeOfDay (..), TimeZone, ZonedTime (..), fromGregorian,
@@ -16,10 +50,9 @@ import Toml.Parser.Key (keyP)
 import Toml.Parser.String (textP)
 import Toml.Parser.Validate (validateItems)
 import Toml.Parser.Value (arrayP, boolP, dateTimeP, doubleP, integerP)
-import Toml.PrefixTree (Key (..), Piece (..), fromList)
-import Toml.Type (AnyValue (..), TOML (..), UValue (..), Value (..))
+import Toml.PrefixTree (Key (..), Piece (..))
+import Toml.Type (TOML (..), UValue (..))
 
-import qualified Data.HashMap.Lazy as HashMap
 import qualified Data.List.NonEmpty as NE
 
 ----------------------------------------------------------------------------
@@ -93,25 +126,7 @@ makeOffset hours mins = minutesToTimeZone (hours * 60 + mins * signum hours)
 makeKey :: [Text] -> Key
 makeKey = Key . NE.fromList . map Piece
 
-tomlFromKeyVal :: [(Key, AnyValue)] -> TOML
-tomlFromKeyVal kv = TOML (HashMap.fromList kv) mempty mempty
-
-tomlFromTable :: [(Key, TOML)] -> TOML
-tomlFromTable t = TOML mempty (fromList t) mempty
-
-tomlFromArray :: [(Key, NonEmpty TOML)] -> TOML
-tomlFromArray = TOML mempty mempty . HashMap.fromList
-
 -- Test Data
-
--- Key-Value pairs
-str, int :: (Key, AnyValue)
-str = (makeKey ["key1"], AnyValue (Text "some string"))
-int = (makeKey ["key2"], AnyValue (Integer 123))
-
-strT, intT :: TOML
-strT = tomlFromKeyVal [str]
-intT = tomlFromKeyVal [int]
 
 int1, int2, int3, int4 :: UValue
 int1 = UInteger 1
--- a/test/Test/Toml/Parsing/Unit/Common.hs
+++ b/test/Test/Toml/Parsing/Unit/Common.hs
@@ -23,7 +23,6 @@ module Test.Toml.Parsing.Unit.Common
        , dquote
        , squote3
        , dquote3
-       , makeKey
        , makeOffset
        , makeZoned
        , int1
@@ -50,10 +49,8 @@ import Toml.Parser.Key (keyP)
 import Toml.Parser.String (textP)
 import Toml.Parser.Validate (validateItems)
 import Toml.Parser.Value (arrayP, boolP, dateTimeP, doubleP, integerP)
-import Toml.PrefixTree (Key (..), Piece (..))
+import Toml.PrefixTree (Key (..))
 import Toml.Type (TOML (..), UValue (..))
-
-import qualified Data.List.NonEmpty as NE
 
 ----------------------------------------------------------------------------
 -- Utilities
@@ -122,9 +119,6 @@ makeZoned d h offset = UZoned $ ZonedTime (LocalTime d h) offset
 
 makeOffset :: Int -> Int -> TimeZone
 makeOffset hours mins = minutesToTimeZone (hours * 60 + mins * signum hours)
-
-makeKey :: [Text] -> Key
-makeKey = Key . NE.fromList . map Piece
 
 -- Test Data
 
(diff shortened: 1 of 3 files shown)


test/Test/Toml/Parsing/Unit/Toml.hs
```diff
@@ -8,109 +9,120 @@ import Data.Text (Text)
...
 tomlSpecs = do
     describe "Key/values" $ do
         it "can parse key/value pairs" $ do
-            parseToml "x='abcdef'" (tomlFromKeyVal [(makeKey ["x"], AnyValue (Text "abcdef"))])
-            parseToml "x=1"        (tomlFromKeyVal [(makeKey ["x"], AnyValue (Integer 1))])
-            parseToml "x=5.2"      (tomlFromKeyVal [(makeKey ["x"], AnyValue (Double 5.2))])
-            parseToml "x=true"     (tomlFromKeyVal [(makeKey ["x"], AnyValue (Bool True))])
-            parseToml "x=[1, 2, 3]" (tomlFromKeyVal [(makeKey ["x"] , AnyValue (Array [Integer 1, Integer 2, Integer 3]))])
-            parseToml "x = 1920-12-10" (tomlFromKeyVal [(makeKey ["x"], AnyValue (Day day2))])
+            parseToml "x = 'abcdef'" $ mkToml ("x" =: Text "abcdef")
```
chshersh: There's `IsString` instance for `Value` so you can actually write such tests simpler
```suggestion
            parseToml "x = 'abcdef'" $ mkToml ("x" =: "abcdef")
```

test/Test/Toml/Parsing/Unit/Toml.hs
```diff
@@ -8,109 +9,120 @@ import Data.Text (Text)
...
+            parseToml "x = 1"  $ mkToml ("x" =: Integer 1)
```
chshersh: Thanks to the `Num` instance, you can also omit `Integer` part :slightly_smiling_face: 
```suggestion
            parseToml "x = 1"  $ mkToml ("x" =: 1)
```

test/Test/Toml/Parsing/Unit/Toml.hs
```diff
@@ -8,109 +9,120 @@ import Data.Text (Text)
...
+            parseToml "x = 5.2" $ mkToml ("x" =: Double 5.2)
```
chshersh: Hmm, for some reasons there's no `Fractional` instance so you can't just write `"x" =: 5.2` :thinking: Probably just an oversight and nothing stops from having `Fractional` instance for `Value`. But this can be done in a separate issue :slightly_smiling_face:

test/Test/Toml/Parsing/Unit/Toml.hs
```diff
@@ -8,109 +9,120 @@ import Data.Text (Text)
...
+            parseToml "x = true" $ mkToml ("x" =: Bool True)
```
chshersh: Let's also remove spaces around `=` in some tests. We also want to test that this parses correctly in different situations:

```haskell
x=true
x= true
x =true
x = true
```

I'm not sure whether we have tests for all these things, but we can just have different space placement in some of the existing tests.

test/Test/Toml/Parsing/Unit/Toml.hs
```diff
@@ -8,109 +9,120 @@ import Data.Text (Text)
...
+            parseToml "x = [1, 2, 3]" $ mkToml ("x" =: Array [Integer 1, Integer 2, Integer 3])
```
chshersh: Just one more simplification here
```suggestion
            parseToml "x = [1, 2, 3]" $ mkToml ("x" =: Array [1, 2, 3])
```

test/Test/Toml/Parsing/Unit/Toml.hs
```diff
@@ -8,109 +9,120 @@ import Data.Text (Text)
...
+                       \[[array.subarray]] \nkey2 = 123" array
+        -- TODO this does not work currently, Edsl does not seem to be able to generate arrays of tables with more than one different property
+        --it "can parse an array of arrays" $ do
+            --let
+                --arr1 = tableArray "table-1" ("key1" =: Text "some string" :| [])
+                --arr2 = tableArray "table-2" ("key2" =: Integer 123 :| [])
+                --array = mkToml $ tableArray "array" $ arr1 :| [arr2]
+
+            --parseToml "[[array]]\n [[array.table-1]] \nkey1 = \"some string\"\n \
+                                    -- \[[array.table-2]] \nkey2 = 123" array
```
chshersh: @Nimor111 Could you please clarify this issue a bit more? It's not immediately clear what's the problem here... Could the following examples of using `tableArray` help?

https://github.com/kowainik/tomland/blob/5ad363a001b47fd30e81971457312233c7b8612c/test/Test/Toml/Printer/Golden.hs#L29-L34

I understand that the documentation always can be improved (even if we already have a lot of it!). Maybe I'm just missing something here :thinking:

Nimor111: Well, in these examples the structure of the toml is the following: 

```toml
[[deeper]]
ping = "pong"

[[deeper]]

[[deeper]]
green = true

[[deeper]]
[deeper.blue]
red = 255
```

And the expected string here is 
```toml
[[array]]
[[array.table-1]]
key1 = "some string"

[[array.table-2]]
key2 = 123
```

In both of the examples from the golden tests there is only a single thing as a property of the array.
What I got to work was this:

```toml
[[array]]
[[array.table-1]]
key1 = "some string"

[[array]]
[[array.table-2]]
key2 = 123
```

chshersh: I tried in GHCi and the following works for me:

```haskell
mkToml $
    tableArray "array" $
        tableArray "table-1" ("key1" =: "some string" :| [])
        :| [ Edsl.tableArray "table-2" ("key2" =: 123 :| []) ]
```

Nimor111: This generates this: 

```toml
[[array]]
[[array.table-1]]
key1 = "some string"

[[array]]
[[array.table-2]]
key2 = 123
```
And the expected string is 

```toml
[[array]]
[[array.table-1]]
key1 = "some string"

[[array.table-2]]
key2 = 123
```
Which sadly is not exactly the same. :/

chshersh: @Nimor111 Ah, I see the difference now. In that case the following should work:

```haskell
mkToml $ tableArray "array" $
    (tableArray "table-1" ("key1" =: "some string" :| []) >> tableArray "table-2" ("key2" =: 123 :| []))
    :| []
```

Nimor111: Works great, thanks!

test/Test/Toml/Parsing/Unit/Toml.hs
```diff
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE PatternSynonyms  #-}
 
 module Test.Toml.Parsing.Unit.Toml where
```
chshersh: I believe that a lot of functions like `makeKey` are redundant now. Could you please add the minimal required explicit export list to this module so it becomes apparent which old utility functions can be removed now?

Nimor111: Hey, just to clarify, you mean add an export list to `Toml.Toml.Parsing.Unit.Common`?

chshersh: @Nimor111 Yes, you're right. I missed the fact that now all functions are in the `Common` module. I think `makeKey` function can be removed as well since there's `IsString` instance for `Key` and all usages like `makeKey ["foo", "bar"]` can be replaced with just `"foo.bar"`.

test/Test/Toml/Parsing/Unit/Common.hs
```diff
@@ -1,8 +1,42 @@
 {-# LANGUAGE FlexibleContexts #-}
 
-module Test.Toml.Parsing.Unit.Common where
+module Test.Toml.Parsing.Unit.Common
+       ( parseX
+       , failOn
+       , parseArray
+       , parseBool
+       , parseDouble
+       , parseInteger
```
chshersh: Wow, thatt's a huge list :open_mouth: Thank you for doing this careful and necessary work! :+1:

Nimor111: This should wrap it up, i'll do the `neat-interpolation` package refactoring in another PR. :+1:

test/Test/Toml/Parsing/Unit/Toml.hs
Nimor111: This should wrap it up, i'll do the neat-interpolation package refactoring in another PR. :+1:

chshersh: @Nimor111 Thanks for your work! I think that at this stage `neat-interpolation` is not really required. New tests are already clean enough :+1: Also, since we want to test that everything works with trailing whitespaces, there're always will be tests written in the old style, like `"foo   "` (because editors usually remove trailing whitespaces). So no need to use two ways to describe the same thing, we can live with raw strings for now :slightly_smiling_face:

Nimor111: Fair enough. :slightly_smiling_face:

PR #244, opened 2019-11-06
[#241] Implement codecs for some common Monoid types

### LINKED ISSUE #241: Implement codecs for some common Monoid types
- [ ] `Last`
- [ ] `First`
- [ ] `Any`
- [ ] `All`

Including generic codecs.

These data types are very common and useful, so why not implement codecs for them and remove some boilerplate in other projects.

vrom911:
Resolves #241

I think that iit probably should be a major version update as the name clash could harm when updating...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,11 +3,14 @@
 tomland uses [PVP Versioning][1].
 The changelog is available [on GitHub][2].
 
-## 1.2.1.0 — Nov 3, 2019
+## 1.2.1.0 — Nov 6, 2019
 
 * [#203](https://github.com/kowainik/tomland/issues/203):
   Implement codecs for `Map`-like data structures.
   (by [@chshersh](https://github.com/chshersh))
+* [#241](https://github.com/kowainik/tomland/issues/241):
+  Implement codecs for `Monoid` wrappers:
+  `all`, `any`, `sum`, `product`, `first`, `last`.
 
 ## 1.2.0.0 — Oct 12, 2019
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The changelog is available [on GitHub][2].
 * [#241](https://github.com/kowainik/tomland/issues/241):
   Implement codecs for `Monoid` wrappers:
   `all`, `any`, `sum`, `product`, `first`, `last`.
+  (by [@vrom911](https://github.com/vrom911))
 
 ## 1.2.0.0 — Oct 12, 2019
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,11 +3,14 @@
...
 
-## 1.2.1.0 — Nov 3, 2019
+## 1.2.1.0 — Nov 6, 2019
 
 * [#203](https://github.com/kowainik/tomland/issues/203):
   Implement codecs for `Map`-like data structures.
   (by [@chshersh](https://github.com/chshersh))
+* [#241](https://github.com/kowainik/tomland/issues/241):
+  Implement codecs for `Monoid` wrappers:
+  `all`, `any`, `sum`, `product`, `first`, `last`.
```
chshersh: It's great and useful feature, you deserve credit!
```suggestion
  (by [@vrom911](https://github.com/vrom911))

```

src/Toml/Bi/Combinators.hs
```diff
@@ -293,6 +305,58 @@ arrayNonEmptyOf :: TomlBiMap a AnyValue -> Key -> TomlCodec (NonEmpty a)
...
+all :: Key -> TomlCodec All
+all = dimap (Just . getAll) (All . fromMaybe False) . dioptional . bool
+{-# INLINE all #-}
+
+{- | Codec for 'Any' wrapper for boolean values.
+
+@since 1.2.1.0
+-}
+any :: Key -> TomlCodec Any
+any = dimap (Just . getAny) (Any . fromMaybe False) . dioptional . bool
```
chshersh: That's an interesting idea! My initial thought was that this codec should always require a boolean field. But actually quite often I want to have missing fields that defaults to `False` (especially in `summoner`). And since `any` is a `Monoid`, it's actually quite fine to return `False` on missing fields, great idea :clap: 

Should we add to documentation that returns `False` on missing fields?

src/Toml/Bi/Combinators.hs
```diff
@@ -293,6 +305,58 @@ arrayNonEmptyOf :: TomlBiMap a AnyValue -> Key -> TomlCodec (NonEmpty a)
...
+----------------------------------------------------------------------------
+-- Monoid codecs
+----------------------------------------------------------------------------
+
+{- | Codec for 'All' wrapper for boolean values.
+
+@since 1.2.1.0
+-}
+all :: Key -> TomlCodec All
+all = dimap (Just . getAll) (All . fromMaybe False) . dioptional . bool
```
chshersh: Maybe here should be `True` as an argument to `False`? Because `mempty` for `All` is `All True` but it's hard to guess for me what will be more useful :thinking:

vrom911: Ah, good catch! It could actually spoil smth.

Issue #254, opened 2020-04-26
Dynamic Keys


everythingfunctional:
I'm trying to write something to parse into a Map, where the keys in the toml file are the keys in the map. I can't seem to find any way of doing this.

Basically I need something like the following.

```
someFunc :: Codec a -> Codec (Map String a)
```

These keys are going to be in a table, so if that makes things more concrete, what I really want would be more like

```
genericTable :: Key -> Codec a -> Codec (Map String a)
```

An example of the toml file would be something like

```
[dependencies]
thing1 = { ... }
thing2 = "1.2.3"
...
```

So I know I'll need a Codec to match the values, since those aren't a simple type, and I know the table the keys will be directly under, but I need to parse each of the keys into my data structure.

Sorry if this was a bit disorganized. Hopefully that's enough info to see what I'm after.

chshersh:
Hi @everythingfunctional, thanks for using `tomland` and opening an issue! Unfortunately, `tomland` doesn't have out-of-the-box solution for this case. However, I can see how this can be useful in some cases. So I don't mind having such function implemented in `tomland` :slightly_smiling_face: 

A few notes about the implementation:

1. I agree with the idea that it should work with the table and should take a table name as an argument.
2. The `Key` type is represented as a list of `Text` pieces, so the returning type will be something like `Map Key a` or `Map Text a`. Libraries like `aeson` provide a custom abstraction like [`FromJSONKey`](http://hackage.haskell.org/package/aeson-1.4.7.1/docs/Data-Aeson.html#t:FromJSONKey) so you can have your custom `Map` keys. But `tomland` uses combinator-based approach instead of typeclasses-based.
https://github.com/kowainik/tomland/blob/6ca6ed3a3d6e1bbcd2b7b06ac82238ef56b00ccc/src/Toml/PrefixTree.hs#L78-L81

So I think that we can have a function like:

```haskell
tableMapCodec
    :: Ord key
    => (BiMap Key key)  -- ^ Bidirectional converter between TOML and Map keys
    -> Key  -- ^ Table name key
    -> TomlCodec val  -- ^ Codec for map values
    -> TomlCodec (Map key val)
```

everythingfunctional:
I'm still a bit new to doing stuff this advanced in Haskell. I'd be happy to take a shot at implementing such a function, but I'll need some guidance on where to start, where to put it, and what might be involved.

chshersh:
[...]

PR #255, opened 2020-04-26
Dynamic keys

everythingfunctional:
Implement `tableMapCodec` and necessary supporting functions as outlined in #254. Includes `TomlBiMap`s from Key to Text and String.

### DIFF
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -513,7 +513,7 @@ map keyCodec valCodec key = Codec input output
                     insertTableArrays key $ t :| (ts ++ tomls)
 
         dict <$ modify updateAction
-
+-- TODO: add docs and tests
 tableMapCodec
     :: forall key val . Ord key
     => TomlBiMap Key key  -- ^ Bidirectional converter between TOML and Map keys
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -544,4 +544,4 @@ tableMapCodec keyBiMap valBiMap key = Codec input output
     update toml (k, v) = do
         tomlKey <- MaybeT $ pure $ either (const Nothing) Just $ backward keyBiMap k
         anyVal <- MaybeT $ pure $ either (const Nothing) Just $ forward valBiMap v
-        toml <$ modify (insertKeyAnyVal tomlKey anyVal)
+        return $ insertKeyAnyVal tomlKey anyVal toml
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -515,14 +515,14 @@ map keyCodec valCodec key = Codec input output
         dict <$ modify updateAction
 
 tableMapCodec
-    :: Ord key
+    :: forall key val . Ord key
     => TomlBiMap Key key  -- ^ Bidirectional converter between TOML and Map keys
     -> TomlBiMap val AnyValue  -- ^ Bidirectional converter between TOML and Map values
     -> Key  -- ^ Table name key
     -> TomlCodec (Map key val)
 tableMapCodec keyBiMap valBiMap key = Codec input output
   where
-    -- input :: Env (Map key val)
+    input :: Env (Map key val)
     input = do
         mTable <- asks $ Prefix.lookup key . tomlTables
         case mTable of
@@ -534,11 +534,12 @@ tableMapCodec keyBiMap valBiMap key = Codec input output
                         Left err -> throwError $ BiMapError err
                     Left err -> throwError $ BiMapError err
 
-    -- output :: Map key val -> St (Map key val)
+    output :: Map key val -> St (Map key val)
     output dict = do
         newToml <- foldM update mempty (Map.toList dict)
         dict <$ modify (insertTable key newToml)
 
+    update :: TOML -> (key, val) -> St TOML
     update toml (k, v) = do
         tomlKey <- MaybeT $ pure $ either (const Nothing) Just $ backward keyBiMap k
         anyVal <- MaybeT $ pure $ either (const Nothing) Just $ forward valBiMap v
--- a/src/Toml/Bi/Combinators.hs
+++ b/src/Toml/Bi/Combinators.hs
@@ -76,7 +76,7 @@ module Toml.Bi.Combinators
 
 import Prelude hiding (all, any, last, map, product, read, sum)
 
-import Control.Monad (forM)
+import Control.Monad (foldM, forM)
 import Control.Monad.Except (catchError, throwError)
 import Control.Monad.Reader (asks, local)
 import Control.Monad.State (execState, gets, modify)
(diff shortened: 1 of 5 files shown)


everythingfunctional:
This is still work in progress, but I wanted to make sure what I've done so far was going to work.

everythingfunctional:
I've got a function I think will work. I had to use a `TomlBiMap` instead of `TomlCodec`, because I couldn't figure out how to use the codec without it already having a `Key`. This really needs somebody with more experience than me to look it over and give me feedback though.

src/Toml/Bi/Map.hs
```diff
@@ -572,6 +578,26 @@ _IntSet :: TomlBiMap IS.IntSet AnyValue
...
+    , backward = textToKey . T.pack
+    }
+
+keyToText :: Key -> Text
+keyToText = (T.intercalate ".") . (map unPiece) . toList . unKey
+
+textToKey :: Text -> Either TomlBiMapError Key
+textToKey t = case P.parse keyP "" t of
+    Left err  -> Left $ ArbitraryError $ T.pack $ P.errorBundlePretty err
+    Right key -> Right key
```
chshersh, line 599: Nice idea to reuse parser for `Key` to convert from `Text`!

src/Toml/Bi/Map.hs
```diff
@@ -52,6 +52,8 @@ module Toml.Bi.Map
        , _Set
        , _IntSet
        , _HashSet
+       , _KeyText
+       , _KeyString
```
chshersh, line 56: Let's add a separate section in the documentation for `Key` mappings. We will probably have more of them in the future.
```suggestion

         -- * Bidirectional mappings for 'Key'
       , _KeyText
       , _KeyString
```

src/Toml/Bi/Map.hs
```diff
@@ -572,6 +578,26 @@ _IntSet :: TomlBiMap IS.IntSet AnyValue
...
+    }
+
+_KeyString :: TomlBiMap Key String
+_KeyString = BiMap
+    { forward = Right . T.unpack . keyToText
+    , backward = textToKey . T.pack
+    }
+
+keyToText :: Key -> Text
+keyToText = (T.intercalate ".") . (map unPiece) . toList . unKey
```
chshersh, line 594: This looks like a useful function in general. Could you, please move into the `PrefixTree` module after this line?

https://github.com/kowainik/tomland/blob/6ca6ed3a3d6e1bbcd2b7b06ac82238ef56b00ccc/src/Toml/PrefixTree.hs#L105

Also, I think it can be written in a slightly shorter and more performant way, using `coerce` trick:

```suggestion
keyToText :: Key -> Text
keyToText = T.intercalate "." . toList . coerce
```

`coerce` here will automatically convert `Key` to `NonEmpty Text` with zero runtime cost.

src/Toml/Bi/Map.hs
```diff
@@ -572,6 +578,26 @@ _IntSet :: TomlBiMap IS.IntSet AnyValue
...
+_KeyText = BiMap
+    { forward = Right . keyToText
+    , backward = textToKey
+    }
+
+_KeyString :: TomlBiMap Key String
+_KeyString = BiMap
+    { forward = Right . T.unpack . keyToText
+    , backward = textToKey . T.pack
+    }
```
chshersh, line 591: Can you add some documentation to these functions and `@since` annotation? I'm not sure at the moment, what the next version will be, but having some `@since` annotation will save some time in future. Something like this should work for now:

```haskell
{- | Bidirectional converter between 'Key' and 'Text'.

@since x.x.x.x 
-}
_KeyText :: ....
```

src/Toml/Bi/Combinators.hs
```diff
@@ -513,3 +513,33 @@ map keyCodec valCodec key = Codec input output
                     insertTableArrays key $ t :| (ts ++ tomls)
 
         dict <$ modify updateAction
+
+tableMapCodec
+    :: Ord key
+    => TomlBiMap Key key  -- ^ Bidirectional converter between TOML and Map keys
+    -> TomlBiMap val AnyValue  -- ^ Bidirectional converter between TOML and Map values
```
chshersh, line 520: Yeah, that's true that this function need to take a `TomlBiMap` instead of `Codec` since it works on a level of a single _key-value_ pair, instead of multiple keys. Good catch!

src/Toml/Bi/Combinators.hs
```diff
@@ -513,3 +513,33 @@ map keyCodec valCodec key = Codec input output
...
+    -> Key  -- ^ Table name key
+    -> TomlCodec (Map key val)
+tableMapCodec keyBiMap valBiMap key = Codec input output
+  where
+    -- input :: Env (Map key val)
```
chshersh, line 525: According to our style guide, it's recommended to write explicit type signatures for functions inside `where`. I believe, you tried doing this but got a compiler error. To fix this error, you need to write type signatlure with explicit `forall`, like this:

```haskell
:: forall key val . Ord key => ...
```

```suggestion
    input :: Env (Map key val)
```

src/Toml/Bi/Combinators.hs
```diff
@@ -513,3 +513,33 @@ map keyCodec valCodec key = Codec input output
...
+                    Left err -> throwError $ BiMapError err
+
+    -- output :: Map key val -> St (Map key val)
+    output dict = do
+        newToml <- foldM update mempty (Map.toList dict)
+        dict <$ modify (insertTable key newToml)
+
+    update toml (k, v) = do
+        tomlKey <- MaybeT $ pure $ either (const Nothing) Just $ backward keyBiMap k
+        anyVal <- MaybeT $ pure $ either (const Nothing) Just $ forward valBiMap v
```
chshersh, line 544: Yeah, this looks a bit verbose... It could be written shorter, if a library provided a function like:

```haskell
eitherToMaybeT :: Applicative m => Either e a -> MaybeT m a
```

But [Hoogle](https://hoogle.haskell.org/?hoogle=Either+e+a+-%3E+MaybeT+m+a&scope=set%3Astackage) doesn't show anything 🤔

src/Toml/Bi/Combinators.hs
```diff
@@ -513,3 +513,33 @@ map keyCodec valCodec key = Codec input output
                     insertTableArrays key $ t :| (ts ++ tomls)
 
         dict <$ modify updateAction
+
```
chshersh, line 516: Note to myself to add tests and documentation in the future
```suggestion
-- TODO: add docs and tests
```

src/Toml/Bi/Combinators.hs
everythingfunctional: I basically lifted that form directly from the `match` function. So if there is a solution it would be useful there as well.

everythingfunctional:
I (think) I've addressed all your comments. I think I might try adding a section to the example file to make sure it works.

everythingfunctional:
I tried it out in the example, but it doesn't include the contents of the map. I'm pretty sure I know what the problem is, but I'm not sure how to actually correct it. The `update` function just keeps returning the same empty `TOML` value every time.

https://github.com/kowainik/tomland/pull/255/files#diff-7656782eac820c0e6e7c9bfd084c85d1R517-R547

everythingfunctional:
Figured out how to fix the update function. It works in the example now.

everythingfunctional:
No problem. Just glad I could make it work.

PR #257, opened 2020-05-06
[#252] Switch testing from 'tasty-*' libraries to 'hspec-*' libraries

### LINKED ISSUE #252: Switch testing from 'tasty-*' libraries to 'hspec-*' libraries
In all our projects we're using `hspec` family of Haskell testing libraries for testing our code. `tomland` is the only exception, probably for some historical reasons. I propose to switch from `tasty` to `hspec` because of the following reasons:

1. **Faster local build times.** I don't need to build extra new libraries now and then, since `hspec` most likely will be already built on my machine after `cabal update`.
2. **Code consistency.** Don't need to recall how a different library works each time we return back to `tomland`.

The plan is to:

- [ ] Remove `tasty-hspec`
- [ ] Remove `tasty-discover`
- [ ] Replace `tasty` with `hspec`
- [ ] Replace `tasty-hedgehog` with `hspec-hedgehog`
- [ ] Replace `tasty-silver` with `hspec-golden`

vrom911:
Resolves #252

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,5 @@ TAGS
 
 # other
 .DS_Store
+
+.golden/
--- a/test/Spec.hs
+++ b/test/Spec.hs
@@ -1,1 +1,36 @@
-{-# OPTIONS_GHC -F -pgmF tasty-discover #-}
+module Main (main) where
+
+import Test.Hspec (hspec)
+import Test.Hspec.Hedgehog (modifyMaxDiscardRatio)
+
+import Test.Toml.BiCode.Property (biCodePropertySpec)
+import Test.Toml.BiMap.Property (biMapPropertySpec)
+import Test.Toml.Parsing.Examples (parsingExamplesSpec)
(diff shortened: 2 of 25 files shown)


.gitignore
```diff
@@ -56,3 +56,5 @@ TAGS
 
 # other
 .DS_Store
+
+.golden/
```
chshersh, line 60: Hmm, why this was needed? :thinking:

test/Spec.hs
```diff
@@ -1 +1,36 @@
...
+    parsingExamplesSpec
+    parsingPropertySpec
+    parsingUnitSpec
+    -- prefix tree
+    prefixTreePropertySpec
+    prefixTreeUnitSpec
+    -- Printer
+    prettyPrinterGoldenSpec
+    -- toml
+    tomlLawsSpec
```
chshersh, line 36: I like more explicit tests much better! It's clear what tests we have with what structure :+1:

Issue #261, opened 2020-05-07
Table Maps Codec


everythingfunctional:
This is a follow up to #254. #255 Solved the problem of dynamic keys if the values are simple values. But I need the values to be more complex than that. I need a way of specifying a codec for the values.

everythingfunctional:
I came up with an approach in #262, but I think it's still not quite sufficient. What you'd really like to be able to specify is a codec that will work for simple values, tables or arrays, but this function so far only covers tables. I don't think the array case would be too difficult, but I'm not quite sure how to deal with the simple value case.

chshersh:
Sharing a link to a comment that explains use-case in detail:

* https://github.com/kowainik/tomland/pull/262#issuecomment-625302500

PR #262, opened 2020-05-07
[#261] Dynamic keys with Codec values

everythingfunctional:
I managed to write a function that takes a key, BiMap for the keys, and a codec for the values.

### DIFF
--- a/src/Toml/Codec/Combinators.hs
+++ b/src/Toml/Codec/Combinators.hs
@@ -66,7 +66,7 @@ module Toml.Codec.Combinators
 
       -- * Combinators for Maps
     , map
-    , tableMapCodec
+    , dynamicMapCodec
 
       -- * General construction of codecs
     , match
@@ -103,7 +103,7 @@ import Toml.Codec.Di (dimap, dioptional)
 import Toml.Codec.Error (TomlDecodeError (..))
 import Toml.Codec.Types (Codec (..), TomlCodec, TomlEnv, TomlState)
 import Toml.Type.AnyValue (AnyValue (..))
-import Toml.Type.Key (Key)
+import Toml.Type.Key (Key (..))
 import Toml.Type.TOML (TOML (..), insertKeyAnyVal, insertTable, insertTableArrays)
 
 import qualified Data.ByteString.Lazy as BL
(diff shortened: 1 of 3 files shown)


everythingfunctional:
I'm writing a [package manager](https://github.com/fortran-lang/fpm) and I'm trying to use the same [dependency specification](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html) that is possible in cargo. So I want to be able to parse something like the following into a `Map String VersionSpec` where `Data VersionSpec = Simple String | Git {..} | Path {..} | ..`.

```toml
[dependencies]
thing1 = "^1.0"
thing2 = { git = "https://gitlab.com/whoknows/thing2.git" }
thing3 = { path = "sub-dir" }
```

The previous one does work, but only for the simple case. This new one also works, but only for the complex cases. I'm trying to work towards having something that works for both at the same time, and figured doing them individually to start with would be useful.

chshersh:
@everythingfunctional, I see. However, as you noticed, the new codec won't solve your problem completely. You still need some way to specify different types of codecs (either `TolmBiMap` or `TomlCodec`) while parsing `Map` values. Instead of having 2 codecs for dynamic `Map` in `tomland`, I would like to have a single `dynamicMapCodec` that will solve this problem automatically. It will be a more general solution, and it will cover such cases.

I imagine that you might be not the only person who wants such a configuration, so I would like to have a proper solution in the library. I need to think about the most convenient design for this feature and will implement it in the library but feel free to share your ideas if you come up with something while looking into this issue :slightly_smiling_face:

everythingfunctional:
I came up with a solution for the arrays case. But I think you are correct about trying to come up with a more general solution.

My initial thought is to utilize the idea that you've already got a way of specifying codecs for sum types, and so just try the codec for each of the simple, table and array cases, and the first one that works use that. Maybe instead of the function taking a `TomlCodec a` it should take a `Key -> TomlCodec a` and then it can just use each of the keys it pulls out of each of the `tomlPairs`, `tomlTables` and `tomlTableArrays` to create the codec to use to decode a standalone `TOML` value created with just that key and value?

chshersh:
@everythingfunctional I think it will be a good idea to start with writing a codec for your `VersionSpec` type using combinators for sum types. After that, you can see what type the resulting codec has, and you can implement the universal codec for all cases in one place.

everythingfunctional:
I think I've come up with a working solution.

src/Toml/Codec/Combinators.hs
```diff
@@ -67,6 +67,7 @@ module Toml.Codec.Combinators
       -- * Combinators for Maps
     , map
     , tableMapCodec
+    , dynamicMapCodec
```
chshersh, line 70: I think we don't need `tableMapCodec`, since `dynamicMapCodec` is a more general and powerful function. So we can remove `tableMapCodec`, since it's only a special case. But it would be a good idea to keep a `TODO` near the `dynamicMapCodec` function about adding docs

src/Toml/Codec/Combinators.hs
```diff
@@ -517,33 +518,86 @@ map keyCodec valCodec key = Codec input output
...
+      tomlKey <- MaybeT $ pure $ either (const Nothing) Just $ backward keyBiMap k
+      anyVal <- MaybeT $ pure $ either (const Nothing) Just $ forward valBiMap v
+      return $ insertKeyAnyVal tomlKey anyVal toml
+
+dynamicMapCodec :: forall key val .
+                   Ord key
+                => TomlBiMap Key key
+                -> (Key -> TomlCodec val)
+                -> Key
+                -> TomlCodec (Map key val)
```
chshersh, line 557: According to our style guide, the indentation should not depend on the length of the identifier, so this type signature should be formatted differently:

```haskell
dynamicMapCodec
    :: forall
    .  Ord
    => ...
    -> ...
```

src/Toml/Codec/Combinators.hs
```diff
@@ -517,33 +518,86 @@ map keyCodec valCodec key = Codec input output
...
+dynamicMapCodec keyBiMap valCodec key = Codec input output
+  where
+    input :: TomlEnv (Map key val)
+    input = do
+      mtable <- asks $ Prefix.lookup key . tomlTables
+      case mtable of
+        Nothing   -> pure Map.empty
+        Just toml -> do
+          simpleVals <- fmap Map.fromList $ forM (HashMap.toList (tomlPairs toml)) $ \(thisKey, anyVal) ->
+            let thisToml = TOML {tomlPairs = HashMap.singleton thisKey anyVal, tomlTables = mempty, tomlTableArrays = mempty}
```
chshersh, line 567: We use 4 spaces for indenting code blocks

src/Toml/Codec/Combinators.hs
```diff
@@ -517,33 +518,86 @@ map keyCodec valCodec key = Codec input output
...
+                thisCodec = valCodec thisKey in
+              case forward keyBiMap thisKey of
+                Right k  -> case runTomlCodec thisCodec thisToml of
```
chshersh, line 570: Converting each `key-value` pair to a separate TOML and running the codec each time seems a bit suboptimal. Need to think about how to do this properly :thinking:

vrom911: Agree that this solution is a no-go at the moment.
@everythingfunctional why exactly `runTomlCodec` is needed here? Seems like a hack, probably there are already functions that are required here in `tomland`?

everythingfunctional:
I've fixed the stylistic comments. As to the rigmarole of constructing a new `TOML` and `TomlCodec` to run for every key, I do agree it is probably not optimal. I will stare at it a bit more and see if I can come up with a better solution. But at first glance I couldn't come up with a way to "run" the codec without giving it the key (which it needs the specific key every time), and then I didn't know what to run it on without constructing a `TOML` value.

vrom911:
@everythingfunctional But are you sure that you definitely need to "run" it? I guess that this is a key problem in here. We need to look at the solution without running codec. I don't quite see why this is the necessary step :thinking:

everythingfunctional:
I'm not certain, but it was the obvious choice, and I haven't yet found an alternative. You need a way to convert the value associated with each key into the desired type. The value associated with the key could have 3 different types; `AnyValue`, `PrefixTree TOML`, or `NonEmpty TOML`. You need to unify those back into a single type and `TOML` is the obvious solution. Especially given that you have a codec, and it knows how to convert `TOML` to the desired type.

everythingfunctional:
I saw the `codecReadTOML` function, so I switched to using that. Not sure if that truly addresses the concerns you guys had.

vrom911:
Resolved in #269

PR #265, opened 2020-05-07
[#263] Simplify 'Codec' abstraction

### LINKED ISSUE #263: Simplify 'Codec' abstraction
Currently, we have the following `Codec` abstract data type:

```haskell
data Codec r w c a = Codec
    { codecRead :: r a
    , codecWrite :: c -> w a
    }
```

But in this library it's used only for `TomlCodec`:

```haskell
type TomlCodec a = Codec Env St a a
```

It could be a good simplification to combine two and have

```haskell
data Codec in out = Codec
    { codecRead :: TomlEnv out
    , codecWrite :: in -> TomlState out
    }

type TomlCodec a = Codec a a
```

What do you think? With this approach, we can have a single module `Toml.Codec` with all general functions, encoding and decoding functions instead of two modules `Toml.Monad` and `Toml.Code`

vrom911:
Resolves #263


src/Toml/Codec.hs
```diff
@@ -0,0 +1,57 @@
...
+
+Reexports functions under @Toml.Codec.*@.
+
+* __"Toml.Bi.Map"__: contains implementation of tagged bidirectional
+  isomorphisms.
+* __"Toml.Bi.Monad"__: core abstraction for bidirectional conversion based on
+  profunctor monads.
+* __"Toml.Codec.Combinators"__: TOML specific combinators based on 'Codec' type
+  from "Toml.Bi.Monad".
+* __"Toml.Codec.Code"__: conversion functions like 'decode' and 'encode'.
```
chshersh, line 14: I think this part can be removed. You already provided a nice description of these modules

src/Toml/Codec/Code.hs
```diff
@@ -0,0 +1,95 @@
...
+-}
+decode :: TomlCodec a -> Text -> Either TomlDecodeError a
+decode codec text = do
+    toml <- first ParseError (parse text)
+    runTomlCodec codec toml
+
+{- | Similar to 'decode', but takes a path to a file with textual @TOML@
+values from which it decodes them with the provided codec.
+
+@since 1.3.0.0
```
chshersh, line 54: Thanks for adding `@since` annotations!

src/Toml/Codec/Types.hs
```diff
@@ -0,0 +1,132 @@
...
+type TomlState = MaybeT (State TOML)
+
+{- | Specialied 'Codec' type alias for bidirectional TOML serialization. Keeps
+'TOML' object as both environment and state.
+-}
+type TomlCodec a = Codec a a
+
+{- | Monad for bidirectional conversion. Contains pair of functions:
+
+1. How to read value of type @a@ from immutable environment context ('TomlEnv')?
```
chshersh, line 50: To be consistent with the type definition :slightly_smiling_face: 
```suggestion
1. How to read value of type @outT@ from immutable environment context ('TomlEnv')?
```

src/Toml/Codec/Types.hs
```diff
@@ -0,0 +1,132 @@
...
+2. How to store value of type @a@ in stateful context ('TomlState')?
```
chshersh, line 51:
```suggestion
2. How to store a value of type @inT@ in stateful context ('TomlState') and return a value of type @outT@?
```

src/Toml/Codec/Types.hs
```diff
@@ -0,0 +1,132 @@
...
+This approach with the bunch of utility functions allows to
+have single description for from/to @TOML@ conversion.
+
+In practice this type will always be used in the following way:
+
+@
+type 'TomlCodec' a = 'Codec' a a
+@
+
+Type parameter @c@ if fictional. Here some trick is used. This trick is
```
chshersh, line 62: It was called `@c@` previously
```suggestion
Type parameter @inT@ if fictional. Here some trick is used. This trick is
```

src/Toml/Codec/Types.hs
```diff
@@ -0,0 +1,132 @@
...
+implemented in the [codec](http://hackage.haskell.org/package/codec) package and
+described in more details in related blog post:
+<https://blog.poisson.chat/posts/2016-10-12-bidirectional-serialization.html>.
+-}
+data Codec inT outT = Codec
```
chshersh, line 67: Maybe we can use `i` and `o` type variables in here?

src/Toml/Codec/Di.hs
```diff
@@ -0,0 +1,210 @@
...
+exampleCodec = Example
+    \<$\> Toml.bool "foo" '.=' foo
+    \<*\> 'dimap' unEmail Email (Toml.text "bar") '.=' bar
+@
+-}
+dimap
+    :: (inC -> inD)    -- ^ Mapper for consumer
+    -> (outA -> outB)  -- ^ Mapper for producer
+    -> Codec inD outA  -- ^ Source 'Codec' object
+    -> Codec inC outB  -- ^ Target 'Codec' object
```
chshersh, line 73: While working with `tomland`, I noticed that it's confusing to have these functions work with `Codec` when in reality they work with `TomlCodec` all the time. I used the wrong functions and got scary error messages. So I would propose to specialize types of all functions in here for `TomlCodec` to also have a more readable type signatures.

```suggestion
dimap
    :: (b -> a)    -- ^ Mapper for consumer
    -> (a -> b)  -- ^ Mapper for producer
    -> TomlCodec a  -- ^ Source 'Codec' object
    -> TomlCodec b  -- ^ Target 'Codec' object
```

PR #267, opened 2020-05-08
[#256] Simpler module structure

### LINKED ISSUE #256: Simpler module structure
After working with the `tomland` code for a while, it was discovered that its codebase is not optimal, and sometimes confusing to navigate. Its is a proposal to change the existing module structure. This is a breaking change, but since we recommend to import the top-level module `qualified`, almost all users of `tomland` shouldn't even notice the change.

My current proposal is the following:

- [ ] **Rename:** `Toml.Bi.Combinators` to `Toml.Codec`
    * This module contains all codecs , so it name could help find them
- [ ] **Rename:** `Toml.Bi.Monad` to `Toml.Monad`
- [ ] **Rename:** `Toml.Bi.Map` to `Toml.BiMap`
- [ ] **Rename:** `Toml.Bi.Code` to `Toml.Decoding`
    * It has both functions `encode` and `decode`, so `Decoding` is not precise enough... Maybe we should have two modules: `Toml.Encode` and `Toml.Decode`? Or maybe we can even combine `Encode` with `Toml.Printer`?
- [ ] **Delete:** `Toml.Bi`
    * Module structure is becoming flatter
- [ ] **New**: `Toml.Key`
    * This data is inside the `Toml.PrefixTree` module, and keep forgetting about that. Probably we can move it out to help navigation

Open questions:

1. The core polymorphic monad is called `Codec`. Wouldn't it be confusing to have `Toml.Monad` have the `Codec` type and the `Toml.Codec` module to contain all combinators? Maybe names should be `Toml.Codec` and `Toml.Combinators` instead?
2. Maybe we can even remove the `Toml.Type` prefix and remove the `.Type` part from all `Toml.Type.*` modules?
3. What to do with `Toml.Bi.Code`? :slightly_smiling_face: 

Any ideas of other improvements are appreciated :slightly_smiling_face:

chshersh:
Resolves #256

* Rename modules
* Add docs for sections
* Update copyright years
* Use more granular imports

Short recap of renaming:

* `Toml.Generic -> Toml.Codec.Generic`
* `Toml.Bi.Map -> Toml.Codec.BiMap`
* `Toml.Printer -> Toml.Type.Printer`
* `Toml.Edsl -> Toml.Type.Edsl`
* `Toml.PrefixTree -> Toml.Type.PrefixTree`
* **New**: `Toml.Type.Key`


.github/workflows/ci.yml
```diff
@@ -53,11 +53,11 @@ jobs:
     - name: Build
       run: |
         cabal update
-        cabal build --enable-tests --enable-benchmarks --write-ghc-environment-files=always
+        cabal build all --enable-tests --enable-benchmarks --write-ghc-environment-files=always
 
     - name: Test
       run: |
-        cabal test --enable-tests
+        cabal test all --enable-tests --test-show-details=direct
```
vrom911, line 60: :+1: Good one!

CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ The changelog is available [on GitHub][2].
   Drop support of GHC-8.2.2.
 * [#252](https://github.com/kowainik/tomland/issues/252):
   Move to `hspec-*` family of libraries from `tasty-*`.
+* [#256](https://github.com/kowainik/tomland/issues/256):
+  Rename modules to simplify module structure.
+  __TODO__: write migration guide
```
vrom911, line 15: I think you can add that you renamed `ParseException` to `TomlParseError` in the CHANGELOG as well :slightly_smiling_face:

src/Toml.hs
```diff
@@ -31,24 +32,36 @@ For more detailed examples see README.md in the repository:
...
-    , module Toml.Generic
-    , module Toml.Parser
-    , module Toml.PrefixTree
-    , module Toml.Printer
+    ( -- $codec
+      module Toml.Codec
+      -- $type
     , module Toml.Type
+      -- $parser
+    , module Toml.Parser
```
vrom911, line 44: I like the new simplified and more structured system! :bricks:

src/Toml.hs
```diff
@@ -31,24 +32,36 @@ For more detailed examples see README.md in the repository:
...
+
+{- $codec
+Main types and functions to implement TOML codecs. This module
+provides high-level API of @tomland@ library.
+-}
+
+{- $type
+Low-level implementation details of types that power @tomland@. The
+"Toml.Type" module contains types to repesent TOML AST and basic
+functions to work with it.
```
vrom911, line 59: Maybe mention the EDSL?

PR #269, opened 2020-05-14
[#261] Table MapsCodec

### LINKED ISSUE #261: Table Maps Codec

chshersh:
Resolves #261

@vrom911 and I brainstormed the solution for this problem and came up with a nice implementation. I'm quite happy that we could avoid full-batteries TOML codec running. @everythingfunctional you can check it out, it should address your use case that you've attempted to solve in #262.

### DIFF
--- a/src/Toml/Codec/Combinators.hs
+++ b/src/Toml/Codec/Combinators.hs
@@ -546,8 +546,8 @@ __data__ Rgb = Rgb
     }
 
 __data__ Colour
-   = Hex Text
-   | RGB Rgb
+    = Hex Text
+    | RGB Rgb
 
 colourCodec :: 'TomlCodec' Colour
 colourCodec = ...
@@ -562,7 +562,7 @@ And you want in the result to have a 'Map' like this:
 @
 'Map.fromList'
     [ "yellow" -> Hex "#FFFF00"
-    , "ping"   -> Hex "#FFC0CB"
+    , "pink"   -> Hex "#FFC0CB"
     , "red"    -> Rgb 255 0 0
     ]
 @
--- a/src/Toml/Codec/Combinators.hs
+++ b/src/Toml/Codec/Combinators.hs
@@ -1,3 +1,6 @@
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE TupleSections   #-}
+
 {- |
 Copyright: (c) 2018-2020 Kowainik
 SPDX-License-Identifier: MPL-2.0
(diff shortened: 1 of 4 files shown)


src/Toml/Codec/Combinators.hs
```diff
@@ -516,34 +520,112 @@ map keyCodec valCodec key = Codec input output
...
-                case backward valBiMap anyVal of
-                    Right v  -> case forward keyBiMap tomlKey of
-                        Right k  -> pure (k, v)
-                        Left err -> throwError $ BiMapError err
+    input :: TomlEnv (Map k v)
+    input = asks (Prefix.lookup tableName . tomlTables) >>= \case
+        Nothing -> pure Map.empty
+        Just toml -> local (const toml) $ do
+            valKeys <- asks (HashMap.keys . tomlPairs)
+            tableKeys <- asks (fmap (:|| []) . HashMap.keys . tomlTables)
```
everythingfunctional, line 615: Do you also need to add this?
```haskell
tableArrayKeys <- asks (HashMap.keys . tomlTableArrays)
```

src/Toml/Codec/Combinators.hs
```diff
@@ -516,34 +520,112 @@ map keyCodec valCodec key = Codec input output
...
+    { configColours :: 'Map' 'Text' Colour
+    }
+@
+
+And you want in the result to have a 'Map' like this:
+
+@
+'Map.fromList'
+    [ "yellow" -> Hex "#FFFF00"
+    , "ping"   -> Hex "#FFC0CB"
```
everythingfunctional, line 565: typo here. "ping" should be "pink"

everythingfunctional:
This does look much better. I clearly still have a lot to learn about Haskell programming. And I'm pretty sure it does solve my problem. Thanks.

src/Toml/Codec/Combinators.hs
vrom911: Don't think that we should support such structures.

everythingfunctional: Why not? I can't think of a use case off the top of my head, but I'm just curious.

chshersh: By design, TOML the language discourages nested data structures. If there are compelling use-cases, we can consider adjusting the implementation.

everythingfunctional: Ok, fair enough. Thanks.

PR #285, opened 2020-05-16
[#277] Fill missing parts in tests

### LINKED ISSUE #277: Fill missing parts in tests
I believe the following tests can be added:

- [x] Property tests on `Category` laws for `BiMap`
- [x] Property roundtrip tests for each `BiMap` convered (some of them are not covered)
- [x] Unit tests on `KeysDiff` in `Test.Toml.Type.Key`

chshersh:
Resolves #277

For `Category` laws: ideally, those tests should generate random `BiMap` but this is too complicated. So I'm just checking the laws for the specific `TomlBiMap`. I guess this should be enough for now.

### DIFF
--- a/test/Test/Toml/Type/Key.hs
+++ b/test/Test/Toml/Type/Key.hs
@@ -40,3 +40,5 @@ keysDiffSpec = describe "Key difference" $ do
         keysDiff "key.foo" "key.bar" `shouldBe` Diff "key" "foo" "bar"
     it "Diff: Two components" $
         keysDiff "key.nest.foo" "key.nest.bar" `shouldBe` Diff "key.nest" "foo" "bar"
+    it "Diff: Two diff components" $
+        keysDiff "key.foo.nest" "key.bar.nest" `shouldBe` Diff "key" "foo.nest" "bar.nest"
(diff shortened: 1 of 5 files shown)


test/Test/Toml/Type/Key.hs
```diff
@@ -4,20 +4,39 @@ module Test.Toml.Type.Key
...
+    it "NoPrefix" $
+        keysDiff "foo" "bar" `shouldBe` NoPrefix
+    it "FstIsPref" $
+        keysDiff "key" "key.nest" `shouldBe` FstIsPref "nest"
+    it "SndIsPref" $
+        keysDiff "key.nest" "key" `shouldBe` SndIsPref "nest"
+    it "Diff: Simple" $
+        keysDiff "key.foo" "key.bar" `shouldBe` Diff "key" "foo" "bar"
+    it "Diff: Two components" $
+        keysDiff "key.nest.foo" "key.nest.bar" `shouldBe` Diff "key.nest" "foo" "bar"
```
vrom911, line 42:
```suggestion
        keysDiff "key.nest.foo" "key.nest.bar" `shouldBe` Diff "key.nest" "foo" "bar"
    it "Diff: Two diff components" $
        keysDiff "key.foo.nest" "key.bar.nest" `shouldBe` Diff "key" "foo.nest" "bar.nest"
```

PR #286, opened 2020-05-16
[#237] Combinator with predicate

### LINKED ISSUE #237: Combinator with predicate
Sometimes you need something to be valid only if satisfy the predicate. It would be nice to have such a function in `tomland`.

Recently I've done the one for `Text` using `textBy` function:

https://github.com/kowainik/summoner/blob/2ce94372c0bccaf4adf377814324f6d4fd751366/summoner-cli/src/Summoner/CustomPrelude.hs#L23-L32

However, I don't really see how this could be generalised. Do you have any idea? And do you think that this is generally useful for the library?

vrom911:
Resolves #237

### DIFF
--- a/src/Toml/Codec/BiMap/Conversion.hs
+++ b/src/Toml/Codec/BiMap/Conversion.hs
@@ -56,6 +56,7 @@ module Toml.Codec.BiMap.Conversion
     , _EnumBounded
     , _Read
     , _TextBy
+    , _Validate
 
       -- * 'Key's
     , _KeyText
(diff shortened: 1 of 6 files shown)


src/Toml/Codec/BiMap/Conversion.hs
```diff
@@ -403,6 +404,21 @@ _TextBy toText parseText = BiMap toAnyValue fromAnyValue
...
+Usually used as the 'Toml.Codec.Combinator.Custom.validate' or
+'Toml.Codec.Combinator.Custom.validateIf' combinator.
+
+@since 1.3.0.0
+-}
+_Validate :: forall a . (a -> Either Text a) -> TomlBiMap a AnyValue -> TomlBiMap a AnyValue
+_Validate p BiMap{..} = BiMap forward backwardWithValidation
+  where
+    backwardWithValidation :: AnyValue -> Either TomlBiMapError a
+    backwardWithValidation anyVal = backward anyVal >>= first ArbitraryError . p
```
chshersh, line 420: Neat implementation! :100:

src/Toml/Codec/Combinator/Custom.hs
```diff
@@ -46,3 +52,89 @@ nullary sum types.
...
+@
+
+@
+tomland decode error:  Value does not pass the validation for key: myKeyName
+@
+
+@since 1.3.0.0
+-}
+validateIf :: forall a . (a -> Bool) -> TomlBiMap a AnyValue -> Key -> TomlCodec a
+validateIf p bimap k = validate validateEither bimap k
```
chshersh, line 134: Let's name the variable `biMap` not to confuse with `bimap` from `Data.Bifunctor`
```suggestion
validateIf p biMap k = validate validateEither biMap k
```

src/Toml/Codec/Combinator/Custom.hs
```diff
@@ -46,3 +52,89 @@ nullary sum types.
 enumBounded :: (Bounded a, Enum a, Show a) => Key -> TomlCodec a
 enumBounded = match _EnumBounded
 {-# INLINE enumBounded #-}
+
+{- | Codec that checks the 'BiMap' on the given predicate.
+The predicate function returns the value, if the validation is successful, or
+the 'Text' of the error that should be returned in case of validation failure.
+
+__Example:__
```
chshersh, line 60: Do you think it's worth it to make this section expandable? I'm afraid, nobody will see it if we make it expandable, and the doc size for this module is still not that big, but maybe you have a different idea

vrom911: I also had the same idea but was convinced that nobody will ever notice the example then 😄 Let's consider this option when the module becomes too long.

PR #296, opened 2020-05-17
[#280] More granular `Codec.Combinators` testing

### LINKED ISSUE #280: More granular `Codec.Combinators` testing
Currently, to test `Combinators`, we create huge `BigType`, write a single codec for it perform roundtrip tests. This solution has several problems:

1. Tests are slow. it takes some time to generate a huge data type hundreads of times and decode and encode it.
2. Compilation is slow: the module with `BigType` is big.
3. We don't test all combinators, because we need to have `Generic` instances for everything, but it's not easy to have ordinary and generic instances for everything.

The proposal is to refactor tests for codecs in the following way:

1. Test each manual codec on `roundtrip` tests individually.
2. Change `BigType` to `SmallType` with 5-6 primitive fields + 1 table to test roundtrip properties for it
3. Create `SmallSumType` with 2-4 constructors to test codecs for sum types

Tasks:

- [x] Replace `BigType` with `SmallType`
- [x] Create `SmallSumType`
- [x] Roundtrip tests for `Combinator.`:
    - [x] `Custom`
    - [x] `List`
    - [x] `Map`
    - [x] `Monoid`
    - [x] `Primitive`
    - [x] `Set`
    - [x] `Table`
    - [x] `Time`
    - [x] `Tuple`

chshersh:
Resolves #280


test/Test/Toml/Codec/Combinator/Custom.hs
```diff
@@ -0,0 +1,21 @@
...
+
+import qualified Hedgehog.Gen as Gen
+import qualified Test.Toml.Gen as Gen
+import qualified Toml.Codec.BiMap.Conversion as Toml
+import qualified Toml.Codec.Combinator.Custom as Toml
+
+
+customSpec :: Spec
+customSpec = describe "Combinator.Custom: Roundtrip tests" $ do
+    codecRoundtrip "Read (Int)      " (Toml.read @Int) Gen.genInt
```
vrom911, line 17: Maybe we can add "dummy" `textBy` check as well?
Something like `textBy show (maybeToEither (?) . readMaybe)`?

PR #309, opened 2020-05-18
[#271] Use 'Validation' from 'validation-selective' in 'TomlEnv'

### LINKED ISSUE #271: Use 'Validation' from 'validation-selective' in 'TomlEnv'
`TomlEnv` is 

```haskell
ExceptT TomlDecodeError (Reader TOML)
```

which is the same as

```haskell
ReaderT TOML (Either TomlDecodeError)
```

But what if we use `Validation` instead of `Either`?

```haskell
ReaderT TOML (Validation (NonEmpty TomlDecodeError))
```

With this approach, we will be able to report all errors, not only the first one. But this also will mean that we won't have a `Monad` instance for codec, which is probably fine. But we also need to rewrite all functions to use. Maybe will be easier with `ApplicativeDo` :thinking:

vrom911:
Resolves #271

image

### DIFF
--- a/src/Toml/Codec/Types.hs
+++ b/src/Toml/Codec/Types.hs
@@ -16,14 +16,15 @@ module Toml.Codec.Types
 
          -- * Codec
        , Codec (..)
+
+         -- * Function alternative
+       , (<!>)
        ) where
 
-import Control.Applicative (Alternative (..))
-import Control.Monad (MonadPlus (..))
-import Control.Monad.Except (ExceptT)
-import Control.Monad.Reader (Reader)
+import Control.Applicative (Alternative (..), liftA2)
 import Control.Monad.State (State)
 import Control.Monad.Trans.Maybe (MaybeT (..))
+import Validation (Validation (..))
 
 import Toml.Codec.Error (TomlDecodeError)
 import Toml.Type (TOML (..))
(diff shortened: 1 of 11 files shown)


src/Toml/Codec/Types.hs
```diff
@@ -102,47 +103,35 @@ instance Functor (Codec i) where
...
-        { codecRead  = pure a
+        { codecRead  = \_ -> Success a
         , codecWrite = \_ -> pure a
         }
     {-# INLINE pure #-}
 
     (<*>) :: Codec i (oA -> oB) -> Codec i oA -> Codec i oB
     codecf <*> codeca = Codec
-        { codecRead  = codecRead codecf <*> codecRead codeca
+        { codecRead  = liftA2 (<*>) (codecRead codecf) (codecRead codeca)
```
chshersh, line 113: That was clever! :brain: :nerd_face:

src/Toml/Codec/Types.hs
```diff
@@ -102,47 +103,35 @@ instance Functor (Codec i) where
...
 
--- | @since 0.0.0
-instance Monad (Codec i) where
-    (>>=) :: Codec i oA -> (oA -> Codec i oB) -> Codec i oB
-    codec >>= f = Codec
-        { codecRead  = codecRead codec >>= \a -> codecRead (f a)
-        , codecWrite = \c -> codecWrite codec c >>= \a -> codecWrite (f a) c
-        }
-    {-# INLINE (>>=) #-}
-
```
chshersh, line 125: Goodbye, `Monad`! :wave:

PR #316, opened 2020-05-19
[#311] Reimplement custom State

### LINKED ISSUE #311: Reimplement custom State
Now we don't have `Reader` anymore and the only thing that is blocking `DerivingVia` feature for generics is `State`. We can try to reimplement the custom one and see how it works for us.

vrom911:
Resolves #311

### DIFF
--- a/src/Toml/Codec/Di.hs
+++ b/src/Toml/Codec/Di.hs
@@ -136,7 +136,7 @@ diwrap
        (Coercible a b)
     => TomlCodec a
     -> TomlCodec b
-diwrap = dimap coerce coerce
+diwrap = coerce
 {-# INLINE diwrap #-}
 
 {- | Bidirectional converter for @sum types@. For example, given the data
--- a/src/Toml/Codec/Types.hs
+++ b/src/Toml/Codec/Types.hs
@@ -1,3 +1,9 @@
+{-# LANGUAGE FlexibleInstances     #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE TupleSections         #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE UndecidableInstances  #-}
+
 {- |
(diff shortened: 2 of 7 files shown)


src/Toml/Codec/Types.hs
```diff
@@ -135,3 +133,79 @@ infixl 3 <!>
...
+    {-# INLINE fmap #-}
+
+    (<$) :: a -> TomlState b -> TomlState a
+    a <$ TomlState{..} = TomlState (first (fmap (const a)) . unTomlState)
+    {-# INLINE (<$) #-}
+
+-- | @since 1.3.0.0
+instance Applicative TomlState where
+    pure :: a -> TomlState a
+    pure a = TomlState (Just a,)
```
chshersh, line 158: That's a very elegant implementation!

src/Toml/Codec/Types.hs
```diff
@@ -135,3 +133,79 @@ infixl 3 <!>
 (<!>) :: Alternative f => (a -> f x) -> (a -> f x) -> (a -> f x)
 f <!> g = \a -> f a <|> g a
 {-# INLINE (<!>) #-}
+
+{- | Mutable context for TOML conversion.
```
chshersh, line 137: Should we write that this monad is equivalent to `@MaybeT (State TOML)@`? Maybe it will be helpful for someone who knows monad transformers...

src/Toml/Codec/Di.hs
```diff
@@ -136,7 +136,7 @@ diwrap
        (Coercible a b)
     => TomlCodec a
     -> TomlCodec b
-diwrap = dimap coerce coerce
+diwrap = coerce
```
chshersh, line 139: Current mood: reimplementing `mtl` just to use `coerce` :rofl:

vrom911: :bad_guys: :rofl:

PR #320, opened 2020-05-19
[#272] DerivingVia for Generic table codecs

### LINKED ISSUE #272: DerivingVia for Generic table codecs
I've tried to use fancy tricks for table codecs with generics. You can find my commented code here:

https://github.com/kowainik/tomland/blob/536a5e6ffb148d0dd4e4c4b120913a6744097676/src/Toml/Codec/Generic.hs#L406-L449

It wasn't possible to implement this feature back then. But thanks to @vrom911 nice work on simplifying the `Codec` type, I think we should be able to implement that feature again.

vrom911:
Resolves #272

Thanks a lot, @chshersh for such a great job on this already! I just needed to test it and it all worked :mage:

### DIFF
--- a/examples/generic.toml
+++ b/examples/generic.toml
@@ -1,0 +1,5 @@
+name = "foo"
+
+[address]
+    addressStreet = "Bar"
+    addressHouse = 42
--- a/src/Toml/Codec/Generic.hs
+++ b/src/Toml/Codec/Generic.hs
@@ -125,6 +125,9 @@ module Toml.Codec.Generic
        , ByteStringAsBytes (..)
        , LByteStringAsText (..)
        , LByteStringAsBytes (..)
+
+         -- * Deriving Via
+       , TomlTable (..)
        ) where
 
 import Data.ByteString (ByteString)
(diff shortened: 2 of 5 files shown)


src/Toml/Codec/Generic.hs
```diff
@@ -669,10 +672,6 @@ instance HasCodec a => HasCodec (Last a) where
     hasCodec = Toml.last (hasCodec @a)
     {-# INLINE hasCodec #-}
 
-{-
```
chshersh, line 672: Should we remove field prefixes from the example?

src/Toml/Codec/Generic.hs
```diff
@@ -712,7 +711,6 @@ instance (Generic a, GenericCodec (Rep a)) => HasCodec (TomlTable a) where
 
 instance (Generic a, GenericCodec (Rep a)) => HasItemCodec (TomlTable a) where
     hasItemCodec = Right $ Toml.diwrap $ genericCodec @a
```
chshersh, line 713: Let's add `INLINE` pragmas for this codec to have slightly better chances of having slightly better performance...

examples/generic.toml
```diff
@@ -0,0 +1,5 @@
+name = "foo"
+
+[address]
+    addressStreet = "Bar"
+    addressHouse = 42
```
chshersh, line 5: I guess you can create an issue to start thinking about a possible interface for improving DerivingVia experience and stripping options :slightly_smiling_face:

Issue #331, opened 2020-08-01
Hexidecimal value with underscore


dariodsa:
Hi,
according to the TOML-0.5.0 version ([link](https://toml.io/en/v0.5.0#integer)) hexadecimal, binary and octal values were introduced.  I checked your test modules and you are checking for case insensitivity, negative, non-negative, etc but you missed one sentence in the specification.    

``Underscores are allowed between digits (but not between the prefix and the value).``   

So, for example, the following code snippet should pass:
```haskell
*> Toml.Parser.parse $ "hex3 = 0xdead_beef"
*> Left (TomlParseError {unTomlParseError = "1:19:\n  |\n1 | hex3 = 0xdead_beef\n  |                   ^\nunexpected end of input\nexpecting '-', '.', '=', '_', or alphanumeric character\n"})
```
The same problem is with binary and octal numbers, classic decimal numbers support underscores. 

I noticed the issue #199 is still open. If you want I can look into that and made test cases which will check all changes made from version 0.4.0 to 0.5.0.

chshersh:
Hi @dariodsa, thanks for reporting the issue! This indeed looks like a bug. It would be nice to add test-cases for such situations and to fix the problem itself, of course 🙂

Regarding issue #199: I've opened issue in the TOML repo itself a while ago about providing testable example:

* https://github.com/toml-lang/toml/issues/643

But there will be no example in the TOML repo. Instead, the implementation needs to be verified with the official compliance test-suite:

* https://github.com/toml-lang/compliance

So instead of walking through the examples in documentation, it's much better to perform automatic testing, as described in #167. But this also requires more work. I've also closed #199 as a duplicate of #167 due to some recent changes in the `toml-lang` org.

dariodsa:
Ok, I will add test cases for such situations and try to fix the problem.

chshersh:
@dariodsa No worries and thanks for the help!

Just to give some more context, here is the parser for all integer numbers:

https://github.com/kowainik/tomland/blob/316c893a10d1dc652bbcc46f7f8862974941244e/src/Toml/Parser/Value.hs#L44-L52

For binary, octal and hexadecimal numbers the primitive parsers from `megaparsec` are reused. That's why they don't support underscores. The implementation needs to do something similar to the `decimal` parser:

https://github.com/kowainik/tomland/blob/316c893a10d1dc652bbcc46f7f8862974941244e/src/Toml/Parser/Value.hs#L33-L42

[...]

PR #332, opened 2020-08-05
[#331] Hexidecimal value with underscore

dariodsa:
Issue #331 
Instead of classic functions like ``hexadecimal``, ``octal`` and ``binary`` from ``megaparsec`` which doesn't support underscore, new functions were implemented ( ``hexadecimalP``, ``octalP`` and ``binaryP``). 

In order for those functions to work, I needed to implement a function which will convert a string representation of number in base N into base 10. 

New tests were added to test this bug fix.


src/Toml/Parser/Value.hs
```diff
@@ -41,14 +59,47 @@ decimalP = zero <|> more
...
+
+    check :: Maybe Integer -> Parser Integer
+    check = maybe (fail "Not hexadecimal") pure
+
+-- | Parser for octal numbers : included parsing of underscore.
+octalP :: Parser Integer
+octalP = zero <|> more
+  where
+    zero, more :: Parser Integer
+    zero = 0 <$ char '_'
```
chshersh, line 78: I think that the `zero` parser is redundant in `octalP`, `binaryP` and `hexadecimalP`. I'm not even sure why it's introduced in `decimalP` 🤔 Would be nice to investigate what tests fail when we remove `zero` from `decimalP`. But `zero` definitely looks redundant in such parsers.

src/Toml/Parser/Value.hs
```diff
@@ -41,14 +59,47 @@ decimalP = zero <|> more
     check :: Maybe Integer -> Parser Integer
     check = maybe (fail "Not an integer") pure
 
+-- | Parser for hexadecimal numbers : included parsing of underscore.
+hexadecimalP :: Parser Integer
+hexadecimalP = zero <|> more
```
chshersh, line 64: All 3 parsers have a similar structure. I propose to remove code duplication by introducing a separate function that takes a parser for digit, string tag and a conversion function. If `zero` is redundant in `decimal` as well, then we might implement all 4 functions using the same common parsing function.

src/Toml/Parser/Value.hs
```diff
@@ -22,13 +22,31 @@ import Control.Applicative.Combinators (between, count, option, optional, sepBy1
...
+  where
+    getValue [] = 0
+    getValue (x:xs) = charValue x + base * getValue xs
+
+-- | Returns value of a single char
+charValue :: Char -> Integer
+charValue c
+    | num >= 48 && num <= 57 = num - 48
+    | num >= 97 && num <= 102 = num - 87
+    | num >= 65 && num <= 70 = num - 55
```
chshersh, line 46: I feel like `charValue` function does some redundant work. When using the `hexDigitChar` parser and similar, you already compare the char value with the number. I propose to implement custom parsers that do parsing and conversion. So you will be able to remove the `charValue` function completely. Something like this:

```haskell
hexDigitP :: Parser Integer
hexDigitP = hexCharToDigit <$> hexDigitChar

octalDigitP :: ... similar  goes here
```

And then conversion function will work with `[Integer]` instead of `String` (aka `[Char]`).

test/Test/Toml/Parser/Integer.hs
```diff
@@ -81,3 +81,27 @@ integerSpecs = describe "integerP" $ do
...
             parseInteger "0xAbCdEf" 0xAbCdEf
             parseInteger "0xaBcDeF" 0xaBcDeF
+        it "can parse numbers with underscore in hexadecimal representation" $ do
+            parseInteger "0xAb_Cd_Ef" 0xabcdef
+            parseInteger "0xA_bcd_ef" 0xabcdef
+            parseInteger "0x123_abc" 0x123abc
+            parseInteger "0xa_b_c_1_2_3" 0xabc123
+        it "can't parse when underscore is between hexadecimal prefix and suffix" $ do
+            parseInteger "0x_Abab_ca" 0
+            parseInteger "0x_ababbac" 0
```
chshersh, line 91: I believe these tests should fail, not return `0`

test/Test/Toml/Parser/Integer.hs
```diff
@@ -81,3 +81,27 @@ integerSpecs = describe "integerP" $ do
         it "can parse numbers when hex digits are in both lowercase and uppercase" $ do
             parseInteger "0xAbCdEf" 0xAbCdEf
             parseInteger "0xaBcDeF" 0xaBcDeF
+        it "can parse numbers with underscore in hexadecimal representation" $ do
```
chshersh, line 84: Great job on the tests! It's nice to have more cases checked 🙂

test/Test/Toml/Parser/Integer.hs
```diff
@@ -81,3 +81,27 @@ integerSpecs = describe "integerP" $ do
...
+            parseInteger "0x_Abab_ca" 0
+            parseInteger "0x_ababbac" 0
+        it "can parse numbers with underscore in octal representation" $ do
+            parseInteger "0o12_34_56" 0o123456
+            parseInteger "0o1_2345_6" 0o123456
+            parseInteger "0o76_54_21" 0o765421
+            parseInteger "0o4_5_3_2_6" 0o45326
+        it "can't parse when underscore is between octal prefix and suffix" $ do
+            parseInteger "0o_123_4567" 0
+            parseInteger "0o_1234567" 0
```
chshersh, line 99: Additionally, for each such failure case, you can check `_` not followed by any numbers. Just `0b_`, `0x_` and `0o_`

test/Test/Toml/Parser/Integer.hs
dariodsa: You are right.

src/Toml/Parser/Value.hs
dariodsa: Zero in ``decimalP`` is used for leading zeros. Even though I don't understand why number with leading zero is parsed into zero, when official documentation of TOML says ``Leading zeros are not allowed.``. But that could be discussed in a new issue.

src/Toml/Parser/Value.hs
```diff
@@ -22,13 +22,31 @@ import Control.Applicative.Combinators (between, count, option, optional, sepBy1
...
 
-import Toml.Parser.Core (Parser, binary, char, digitChar, hexadecimal, lexeme, octal, sc, signed,
+import Toml.Parser.Core (Parser, char, digitChar, hexDigitChar, octDigitChar, binDigitChar, lexeme, sc, signed,
                          string, text, try, (<?>))
 import Toml.Parser.String (textP)
 import Toml.Type (AnyValue, UValue (..), typeCheck)
 
+-- | Returns decimal value from base N representation
+baseStringToInt :: Integer -> String -> Maybe Integer
+baseStringToInt base = Just . getValue . reverse
```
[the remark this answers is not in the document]
dariodsa: Now I implemented only one function for hex, oct and bin numbers but without ``foldl'`` because I used ``megaparsec`` parser to convert in from ``[Char]`` or ``String`` into ``Integer``.

src/Toml/Parser/Value.hs
chshersh: > I don't understand why number with leading zero is parsed into zero

Yeah, if the current `tomland` implementation allows numbers with leading zeroes, then this problems deserves a separate issue and a solution...

src/Toml/Parser/Value.hs
```diff
@@ -22,9 +22,13 @@ import Control.Applicative.Combinators (between, count, option, optional, sepBy1
 import Data.Fixed (Pico)
 import Data.Time (Day, LocalTime (..), TimeOfDay, ZonedTime (..), fromGregorianValid,
                   makeTimeOfDayValid, minutesToTimeZone)
+
+import Data.String (fromString)
+
```
chshersh, line 27: Empty lines are redundant around this import 🙂  According to our style guide, imports should be grouped into external libraries, from the package itself, and `qualified` imports. And empty lines are only between those groups.

```suggestion
import Data.String (fromString)
```

src/Toml/Parser/Value.hs
```diff
@@ -41,15 +45,33 @@ decimalP = zero <|> more
...
+numberP parser parseDigit = more
+  where
+    more :: Parser Integer
+    more = check =<< intValueMaybe . concat <$> sepBy1 (some parseDigit) (char '_')
+
+    intValueMaybe :: String -> Maybe Integer
+    intValueMaybe = parseMaybe (parser :: Parser Integer) . fromString
+
+    check :: Maybe Integer -> Parser Integer
+    check = maybe (fail "Not valid number") pure
```
chshersh, line 59: You can pass an additional `String` to `numberP` parse to improve this error message, so it will become `"Invalid hexadecimal number"` or `"Invalid binary number"`

src/Toml/Parser/Value.hs
```diff
@@ -41,15 +45,33 @@ decimalP = zero <|> more
...
 
+-- | Parser for hexadecimal, octal and binary numbers : included parsing
+numberP :: Parser Integer -> Parser Char -> Parser Integer
+numberP parser parseDigit = more
+  where
+    more :: Parser Integer
+    more = check =<< intValueMaybe . concat <$> sepBy1 (some parseDigit) (char '_')
+
+    intValueMaybe :: String -> Maybe Integer
+    intValueMaybe = parseMaybe (parser :: Parser Integer) . fromString
```
chshersh, line 56: Let's rename `parser` to `parseInteger`. Also, the type annotation here is not needed, since the type of `parse` is already specified in the top-level signature.

Issue #333, opened 2020-08-12
Numbers with leading zero


dariodsa:
Currently ``tomland`` interprets decimal numbers with leading zeros into zero. That is not really in the spirit of a TOML-0.5.0. According to the [documentation](https://toml.io/en/v0.5.0#integer):
```
Leading zeros are not allowed. Integer values -0 and +0 are valid and identical to an unprefixed zero.
```
For this behaviour is responsible ``zero`` parser in function ``decimalP``. 
https://github.com/kowainik/tomland/blob/316c893a10d1dc652bbcc46f7f8862974941244e/src/Toml/Parser/Value.hs#L37-L38
So the type of parser should allow leading zero only if it is the only digit in the number, otherwise it shouldn't be allowed to come first.

chshersh:
Resolved by #344

Issue #147, opened 2018-11-13
Parsing of values doesn't require spacing


jiegillet:
I realized this issue working on #128 and I have discussed in with @willbasky in private.
Basically tests like these:
```haskell
it "can parse true" $ parseX boolP "trueXXX" True
```
pass, and it confused me for a bit, but then I realized that it's fine, that's how parsers work. You might say that testing `boolP <* eol` instead of `boolP` would be better, but it's not a big problem. 

Instead, what is a problem is that we are wrapping most parsers with `lexeme` which allows for any space characters and comments, but also accepts 0 spaces.

That means `a = true \n XXX = 1` is a valid `TOML`, but so is `a = trueXXX = 1`, which is bad in my opinion. Also, `a = trueXXX` would also be valid `TOML` since the last `XXX` is simply ignored (this is easily fixed by adding a `eol` in `tomlP` though). In other words, the current parser accepts all valid `.toml` files, but also accepts wrong ones. 

This could probably be fixed by making sure that every value is followed by a space, an EOL or one of these: `,]}`. It would be a bit less performant but we can run benchmarks now.

ChShersh:
@jiegillet This is indeed an issue. I guess what we can do is to have two combinators:

```
lexeme
lexemeNoLn
```

And use `lexemeNoLn` explicitly in places with `eol` combinator to check on end of line. I don't think this will make parsing noticeably slower. But this should be addressed in some future.

PR #344, opened 2020-10-04
Leading zero

dariodsa:
According to the issue [#333], tomland has problem with parsing the numbers with leading zero. That problem is some sort of connected with #147 because parser should consume data and throw an error without recovering. 
That feature is now implemented thanks to the ``observing`` function. So if number has leading zero, an error will be thrown, otherwise normal ``more`` parser will be called. It took me some time to figure it out, so any advice how to make this approach better would be appreciated. 
Dario


chshersh (approved):
@dariodsa Sorry for the delayed review! :pray: 

The implementation looks good :+1: And thanks for writing tests :100:

Issue #326, opened 2020-05-24
Add 'either' and 'validation' codecs for returning TOML errors


chshersh:
The following codecs can be useful for cases when we want to print pretty error messages:

```haskell
either :: TomlCodec a -> TomlCodec (Either [TomlDecodeError] a)
validation :: TomlCodec a -> TomlCodec (Validation [TomlDecodeError] a)
```

PR #346, opened 2020-10-17
Add 'either' combinator for TomlCodec

markus1189:
Implement the `either` combinator from #326

### DIFF
--- a/src/Toml/Codec/Combinator/Custom.hs
+++ b/src/Toml/Codec/Combinator/Custom.hs
@@ -250,10 +250,10 @@ __Example:__
 either (bool "mykey") :: Codec Bool (Either [TomlDecodeError] Bool)
 @
 -}
-either :: forall i o. Codec i o -> Codec i (Either [TomlDecodeError] o)
+either :: forall a. TomlCodec a -> TomlCodec (Either [TomlDecodeError] a)
 either (Codec oldCodecRead oldCodecWrite) = Codec newCodecRead newCodecWrite
   where
-    newCodecRead :: TomlEnv (Either [TomlDecodeError] o)
+    newCodecRead :: TomlEnv (Either [TomlDecodeError] a)
     newCodecRead = Success . validationToEither . oldCodecRead
-    newCodecWrite :: i -> TomlState (Either [TomlDecodeError] o)
-    newCodecWrite = fmap Right . oldCodecWrite
+    newCodecWrite :: Either [TomlDecodeError] a -> TomlState (Either [TomlDecodeError] a)
+    newCodecWrite = traverse oldCodecWrite
--- a/src/Toml/Codec/Combinator/Custom.hs
+++ b/src/Toml/Codec/Combinator/Custom.hs
@@ -10,7 +10,6 @@ See examples below of the situations you may need the following combinators.
 
 @since 1.3.0.0
 -}
-{-# LANGUAGE LambdaCase #-}
 
 module Toml.Codec.Combinator.Custom
     ( -- * 'Text' combinators
@@ -22,20 +21,20 @@ module Toml.Codec.Combinator.Custom
     , validate
     , validateIf
 
-      -- * Transform
+      -- * With the errors reporting
     , either
     ) where
 
 import Prelude hiding (read, either)
 
 import Data.Text (Text)
-import Validation (validationToEither)
+import Validation (validationToEither, Validation(Success))
 
 import Toml.Codec.BiMap (TomlBiMap)
 import Toml.Codec.BiMap.Conversion (_EnumBounded, _Read, _TextBy, _Validate)
 import Toml.Codec.Combinator.Common (match)
 import Toml.Codec.Error (TomlDecodeError)
-import Toml.Codec.Types (TomlCodec, Codec(..))
+import Toml.Codec.Types (TomlCodec, Codec(..), TomlEnv, TomlState)
 import Toml.Type.AnyValue (AnyValue)
 import Toml.Type.Key (Key)
 import Toml.Type.Printer (prettyKey)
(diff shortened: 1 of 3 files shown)


src/Toml/Codec/Combinator/Custom.hs
```diff
@@ -10,6 +10,7 @@ See examples below of the situations you may need the following combinators.
 
 @since 1.3.0.0
 -}
+{-# LANGUAGE LambdaCase #-}
```
chshersh, line 13: This extension is enabled in the `tomland.cabal` file by default in all modules, so no need to add it here :slightly_smiling_face: 
```suggestion
```

src/Toml/Codec/Combinator/Custom.hs
```diff
@@ -20,16 +21,21 @@ module Toml.Codec.Combinator.Custom
       -- * Validation
     , validate
     , validateIf
+
+      -- * Transform
```
chshersh, line 25: Let's rename the section to give some more meaning to it :slightly_smiling_face: 
```suggestion
      -- * With the errors reporting
```

src/Toml/Codec/Combinator/Custom.hs
```diff
@@ -235,3 +241,18 @@ validateIf p biMap k = validate validateEither biMap k
...
+__Example:__
+
+@
+either (bool "mykey") :: TomlCodec (Either [TomlDecodeError] Bool)
+@
+-}
+either :: TomlCodec a -> TomlCodec (Either [TomlDecodeError] a)
+either (Codec r w) = Codec r' $ \case
+  Left x' -> pure . Left $ x'
+  Right x' -> fmap Right . w $ x'
```
chshersh, line 257: This case looks like the `Traversable` instance for `Either`, so you can use a single function `traverse` instead of a manual case:

* https://hackage.haskell.org/package/base-4.14.0.0/docs/src/Data.Traversable.html#line-240

src/Toml/Codec/Combinator/Custom.hs
```diff
@@ -235,3 +241,18 @@ validateIf p biMap k = validate validateEither biMap k
...
+  where r' = pure . validationToEither . r
```
chshersh, line 258: Let's rename variables for clarity, and also use `Success` here for clarity. Just to make the code more maintainable :slightly_smiling_face: 
```suggestion
  where
    newCodecRead :: TomlEnv (Either [TomlDecodeError] a)
    newCodecRead = Success . validationToEither . oldCodecRead
```

test/Test/Toml/Codec/Combinator/Custom.hs
```diff
@@ -25,3 +26,6 @@ customSpec = describe "Combinator.Custom: Roundtrip tests" $ do
     codecRoundtrip "TextBy          "
         (Toml.textBy (Text.pack . show) (first Text.pack . readEither . Text.unpack))
         Gen.genInt
+    codecRoundtrip "either          "
+        (Toml.either . Toml.bool)
+        (Right <$> Gen.genBool)
```
chshersh, line 31: Here you are testing the `either` combinator only on values generated with the constructor `Right`, but we actually want to test on both `Left` and `Right`. So, we need to generate either a list of decode errors or a different value (in your case it's `Bool`)

src/Toml/Codec/Combinator/Custom.hs
```diff
@@ -235,3 +241,18 @@ validateIf p biMap k = validate validateEither biMap k
...
+
+{- | Converts a given @TomlCodec@ to give you access to the list of
+@TomlDecodeError@s as an @Either@
+
+__Example:__
+
+@
+either (bool "mykey") :: TomlCodec (Either [TomlDecodeError] Bool)
+@
+-}
```
chshersh, line 253: Very nice documentation!

chshersh (commented):
@markus1189 Thanks for your contribution! Much appreciated :slightly_smiling_face: 
I left some comments on possible code improvements and some suggestions on how to move this forward :fast_forward:

src/Toml/Codec/Combinator/Custom.hs
markus1189: Removed

markus1189: Removed

src/Toml/Codec/Combinator/Custom.hs
markus1189: Sure!

src/Toml/Codec/Combinator/Custom.hs
markus1189: You're totally right! :wink:

src/Toml/Codec/Combinator/Custom.hs
markus1189: Yes I agree, I moved out both `newCodecRead` and `newCodecWrite`

test/Test/Toml/Codec/Combinator/Custom.hs
markus1189: Hmmm I'm no longer sure if `TomlCodec (Either [TomlDecodeError] a)` is correct :thinking: 

We probably don't want to enode a `Left [TomlDecodeError]`... Maybe we should change `either` to:

```
either :: Codec i o -> Codec i (Either [TomlDecodeError] o)
```

and then we need to use something other than `codecRoundtrip` in the test :thinking: 
```

**Review by markus1189 — COMMENTED** [review]
Thanks @chshersh for the friendly review!  I left some comments in answer

File: test/Test/Toml/Codec/Combinator/Custom.hs (hunk 5 above)
Comment (vrom911) [reply test/Test/Toml/Codec/Combinator/Custom.hs]: @markus1189 I think it is okay to have `TomlCodec` in type as `TomlDecodeError` works with `TomlCodec` 🙂 
You can just change the generator to something with [`Gen.either`](https://hackage.haskell.org/package/hedgehog-1.0.3/docs/Hedgehog-Gen.html#v:either).

File: test/Test/Toml/Gen.hs [hunk 7]
```diff
@@ -187,6 +190,36 @@ genToml = Gen.recursive
...
+genTValue = Gen.element [ TBool
+                        , TInteger
+                        , TDouble
+                        , TText
+                        , TZoned
+                        , TLocal
+                        , TDay
+                        , THours
+                        , TArray
+                        ]
```
vrom911, line 203: I guess we can use [`enumBounded`](https://hackage.haskell.org/package/hedgehog-1.0.3/docs/Hedgehog-Internal-Gen.html#v:enumBounded) here, if we derive `Enum` for `TValue` 🙂

vrom911:
Closing due to the inactivity.
Feel free to reopen if you would like to continue working on it 🙂

PR #347, opened 2020-10-22
Fix doc links to Toml.Codec.Combinator.*

AlistairB:
I noticed these are broken links currently https://hackage.haskell.org/package/tomland-1.3.1.0/docs/Toml-Codec-Combinator.html , the module is `Combinator` not `Combinators`.


vrom911 (approved):
Great catch! Thank you 🤗

Issue #334, opened 2020-08-16
Unicode characters incorrectly escaped in encode


samhh:
At least, that's what I think is happening. In a REPL, the following will fail:

```haskell
Toml.decode (Toml.text "k") $ Toml.encode (Toml.text "k") "ü"
-- Left [ParseError (TomlParseError {unTomlParseError = "1:8:\n  |\n1 | k = \"\\252\"\n  |        ^\nInvalid escape sequence: \\2\n"})]
```

Looking at the encoding, here's what we're given:

```haskell
Toml.encode (Toml.text "k") "ü"
-- "k = \"\\252\"\n"
```

And passing that into decode will fail per the above:

```haskell
Toml.decode (Toml.text "k") "k = \"\\252\"\n"
-- Left [ParseError (TomlParseError {unTomlParseError = "1:8:\n  |\n1 | k = \"\\252\"\n  |        ^\nInvalid escape sequence: \\2\n"})]
```

Looking at the [TOML spec](https://toml.io/en/v0.5.0#string), it looks like Unicode characters should be encoded with a `\u` prefix. Modifying the string to contain an extra `u0` allows encoding to succeed, and I _think_ that's what we want given that's roughly the decimal output in an online Unicode converter:

```haskell
Toml.decode (Toml.text "k") "k = \"\\u0252\"\n"
-- Right "\594"
```

But I'm pretty ignorant about character encoding and am honestly not sure if that's the right output. :smile:

chshersh:
Hi @samhh, thanks for submitting the issue! This indeed looks like an unexpected behaviour :disappointed: 
`tomland` uses `Text` internally, and during encoding `Text` is printed using the `show` function. The `show` does the escaping of all characters. You can reproduce this behaviour even easier in GHCi:

```haskell
λ: show "ü"
"\"\\252\""
```

It looks like some smarter handling of Unicode characters is required to preserve TOML semantics.

Relevant code to change is here:

https://github.com/kowainik/tomland/blob/38d74d37efa876e57c01bcfecca56f032305d957/src/Toml/Type/Printer.hs#L141

A more interesting question is why such errors weren't caught by our property tests? :thinking: :thinking: :thinking:

dariodsa:
I implemented the requested feature, now I only have to implement tests and figure it out why it wasn't caught by our test cases.

dariodsa:
[...]

PR #354, opened 2020-12-09
Unicode escape

dariodsa:
Issue #334 

### File changed
 - ``src/Toml/Type/Printer.hs`` 
 - ``test/Test/Toml/Gen.hs``

Function in ``Printer.hs`` was changed in order to allow escaping unicode characters. Firstly escaping unicode characters are performed before calling function ``show`` which will escape others like ``\n``.
I implemented tests but they are not functional. The reason is that because currently char ``č`` is parsed into ``\U0000010d`` and it can only be unparsed into ``\U0000010d``. That is the reason why those tests will fail (starting char is not equal to end). I am still wondering should I add that tests with different printing without \u or \U or this is good enough. 

```haskell
*Toml> Toml.decode (Toml.text "a") $ Toml.encode (Toml.text "a") "č"
Right "\\U0000010d"
```
```haskell
*Toml> Toml.decode (Toml.text "a") $ Toml.encode (Toml.text "a") "\\u010d"
Right "\\u010d"
```

### DIFF
--- a/src/Toml/Type/Printer.hs
+++ b/src/Toml/Type/Printer.hs
@@ -176,15 +176,22 @@ prettyKeyValue options i = mapOrdered (\kv -> [kvText kv]) options . HashMap.toL
     showText = Text.pack . show
 
     showTextUnicode :: Text -> Text
-    showTextUnicode text = Text.pack $ show $ foldl (\acc (ch, asciiCh) -> acc ++ getCh ch asciiCh) "" asciiArr
+    showTextUnicode text = Text.pack quotedText
       where
-        xs = Text.unpack text
-        asciiArr = zip xs $ asciiStatus xs
+        quotedText = finalText
+        finalText = foldl (\acc (ch, asciiCh) -> acc ++ getCh ch asciiCh) "" asciiArr 
+        xss = unicodeEscape $ show $ Text.unpack text
+        asciiArr = zip xss $ asciiStatus xss
         getCh ch True  = [ch]
-        getCh ch False = printf "\\U%08x" ordChr :: String
+        getCh ch False = printf "\\\\U%08x" ordChr :: String
           where
             ordChr = ord ch
         asciiStatus = map isAscii
+
+        unicodeEscape "" = ""
+        unicodeEscape ('\\':'\\':'u':xs) = '\\':'\\':'u': unicodeEscape xs
+        unicodeEscape ('\\':'\\':'U':xs) = '\\':'\\':'U': unicodeEscape xs
+        unicodeEscape (x:xs) = x : unicodeEscape xs
 
     showDouble :: Double -> Text
     showDouble d | isInfinite d && d < 0 = "-inf"
--- a/src/Toml/Type/Printer.hs
+++ b/src/Toml/Type/Printer.hs
@@ -24,7 +24,7 @@ import Data.Char (isAscii, ord)
 import Data.Coerce (coerce)
 import Data.Function (on)
 import Data.HashMap.Strict (HashMap)
-import Data.List (sortBy)
+import Data.List (sortBy, foldl')
 import Data.List.NonEmpty (NonEmpty)
 import Data.Semigroup (stimes)
 import Data.Text (Text)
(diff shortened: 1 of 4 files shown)


chshersh:
Hi @dariodsa! Thanks for working on this 🤗 
It's a bit busy time, but we'll try to find time to review the PR.

src/Toml/Type/Printer.hs
```diff
@@ -171,6 +175,19 @@ prettyKeyValue options i = mapOrdered (\kv -> [kvText kv]) options . HashMap.toL
     showText :: Show a => a -> Text
     showText = Text.pack . show
 
+    showTextUnicode :: Text -> Text
+    showTextUnicode text = Text.pack quotedText
+      where
+        quotedText = show finalText
+        finalText = foldl (\acc (ch, asciiCh) -> acc ++ getCh ch asciiCh) "" asciiArr
```
chshersh, line 182: `foldl` leaks memory and it's discouraged to use it. Instead, it's better to use `foldl'`

src/Toml/Type/Printer.hs
```diff
@@ -171,6 +175,19 @@ prettyKeyValue options i = mapOrdered (\kv -> [kvText kv]) options . HashMap.toL
...
+      where
+        quotedText = show finalText
+        finalText = foldl (\acc (ch, asciiCh) -> acc ++ getCh ch asciiCh) "" asciiArr 
+        xss = Text.unpack text
+        asciiArr = zip xss $ asciiStatus xss
+        getCh ch True  = [ch]
+        getCh ch False = printf "\\U%08x" ordChr :: String
+          where
+            ordChr = ord ch
+        asciiStatus = map isAscii
```
chshersh, line 189: Could you please add types to all functions and values inside the `where` block? Otherwise, it's a bit hard to understand the whole context and what is going on here.

src/Toml/Type/Printer.hs
```diff
@@ -171,6 +175,19 @@ prettyKeyValue options i = mapOrdered (\kv -> [kvText kv]) options . HashMap.toL
...
 
+    showTextUnicode :: Text -> Text
+    showTextUnicode text = Text.pack quotedText
+      where
+        quotedText = show finalText
+        finalText = foldl (\acc (ch, asciiCh) -> acc ++ getCh ch asciiCh) "" asciiArr 
+        xss = Text.unpack text
+        asciiArr = zip xss $ asciiStatus xss
+        getCh ch True  = [ch]
+        getCh ch False = printf "\\U%08x" ordChr :: String
```
chshersh, line 186: Hmm, I'm curious, what does the `Bool` mean here? 🤔

src/Toml/Type/Printer.hs
```diff
@@ -171,6 +175,19 @@ prettyKeyValue options i = mapOrdered (\kv -> [kvText kv]) options . HashMap.toL
     showText :: Show a => a -> Text
     showText = Text.pack . show
 
+    showTextUnicode :: Text -> Text
```
chshersh, line 178: I think, adding a few lines of documentation to this function would help to understand it better 🙂 Something like "this function converts `Text` to `String`, processes each character to decide how to pretty-print it and converts back to `Text`"

src/Toml/Type/Printer.hs
```diff
@@ -171,6 +175,19 @@ prettyKeyValue options i = mapOrdered (\kv -> [kvText kv]) options . HashMap.toL
...
+    showTextUnicode text = Text.pack quotedText
+      where
+        quotedText = show finalText
+        finalText = foldl (\acc (ch, asciiCh) -> acc ++ getCh ch asciiCh) "" asciiArr 
+        xss = Text.unpack text
+        asciiArr = zip xss $ asciiStatus xss
+        getCh ch True  = [ch]
+        getCh ch False = printf "\\U%08x" ordChr :: String
+          where
+            ordChr = ord ch
```
chshersh, line 188: I think it's okay to simplify this function a little bit 
```suggestion
        getCh ch False = printf "\\U%08x" (ord char) :: String
```

test/Test/Toml/Gen.hs
```diff
@@ -307,8 +314,10 @@ genText = genNotEscape $ fmap Text.concat $ Gen.list (Range.constant 0 256) $ Ge
     , genPunctuation
     , genUniHex4Color
     , genUniHex8Color
+    --, genUnicodeChar
```
chshersh, line 317: Should this be uncommented back?

chshersh (approved):
@dariodsa Awesome job! Thanks for working on it, really appreciate that 👍🏻 
And sorry again for taking too long to review. I left some comments and suggestions, but it already looks great 🙂

test/Test/Toml/Gen.hs
dariodsa: ``genUnicodeChar`` is the function that I wrote myself thinking that I might be helpful to add it into the testing phase. But if I enable it, the test will fail. 
The reason why it fails is mentioned above but I can repeat it. Character ``č`` is transformed into ``\U0000010d`` but decoding it back will give ``\U0000010d`` again, not ``č`` so the ``encode . decode`` tests will fail. 
Currently, tests are only include escaped unicode characters in ``encode . decode`` phase.

chshersh (approved):
@dariodsa Great work! I'm going to merge this PR 👍🏻 
And I propose to create a separate issue regarding `encode . decode` property-based tests for unicode characters to track this problem and figure out a fix for it eventually 🙂

PR #363, opened 2021-02-12
[#186] Implement TOML difference

### LINKED ISSUE #186: Implement TOML difference
Currently if you implement `Codec` for file that contains more fields that you need, `tomland` will use only those fields that you specified. However, sometimes it's desired to tell about redundant or unused fields. In JSON it is completely okay to ignore some fields, but in TOML static configuration having some fields in the config may lead to frustration in the future. It would be nice to have some way of telling which fields were not used.

vrom911:
Resolves #186

Shout out to @chshersh for pair programming 🙌🏼 ❤️

### DIFF
--- a/src/Toml/Type/PrefixTree.hs
+++ b/src/Toml/Type/PrefixTree.hs
@@ -22,6 +22,8 @@ module Toml.Type.PrefixTree
     , insertT
     , lookupT
     , toListT
+    , addPrefixT
+    , differenceWithT
 
       -- * Prefix map that stores roots of 'PrefixTree'
     , PrefixMap
@@ -30,6 +32,7 @@ module Toml.Type.PrefixTree
     , lookup
     , fromList
     , toList
+    , differenceWith
     ) where
 
 import Prelude hiding (lookup)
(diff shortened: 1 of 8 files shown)


src/Toml/Type/PrefixTree.hs
```diff
@@ -168,3 +193,53 @@ toListT (Branch pref ma prefMap) = case ma of
...
+-}
+differenceWith :: (a -> b -> Maybe a) -> PrefixMap a -> PrefixMap b -> PrefixMap a
+differenceWith f = HashMap.differenceWith (differenceWithT f)
+
+{- | Difference of two 'PrefixTree's. Returns elements of the first 'PrefixTree'
+that are not existing in the second one.
+
+@since x.x.x.x
+-}
+differenceWithT :: (a -> b -> Maybe a) -> PrefixTree a -> PrefixTree b -> Maybe (PrefixTree a)
```
chshersh, line 210: This was a tough function to implement 😰

chshersh (approved):
Looks great! It was a pleasure working on this together 🤗 

Looking forward to using this function!

PR #366, opened 2021-02-12
[#322] Strip Prefix with DerivingVia TomlTable

### LINKED ISSUE #322: Strip Prefix with DerivingVia TomlTable
Currently, the prefix of data fields appears as it is in the TOML keys. Maybe we can add some newtype that could handle the stripping as well?

https://github.com/kowainik/tomland/pull/320#discussion_r427483318

chshersh:
Resolves #322

### DIFF
--- a/src/Toml/Codec/Generic.hs
+++ b/src/Toml/Codec/Generic.hs
@@ -723,7 +723,7 @@ instance (Generic a, GenericCodec (Rep a)) => HasItemCodec (TomlTable a) where
     {-# INLINE hasItemCodec #-}
 
 {- | @newtype@ for generic deriving of 'HasCodec' typeclass for custom data
-types that should we wrapped into separate table.
+types that should be wrapped into a separate table.
 
 Similar to 'TomlTable' but also strips the data type name prefix from
 TOML keys.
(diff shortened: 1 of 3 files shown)


src/Toml/Codec/Generic.hs
```diff
@@ -721,6 +722,39 @@ instance (Generic a, GenericCodec (Rep a)) => HasItemCodec (TomlTable a) where
     hasItemCodec = Right $ Toml.diwrap $ genericCodec @a
     {-# INLINE hasItemCodec #-}
 
+{- | @newtype@ for generic deriving of 'HasCodec' typeclass for custom data
+types that should we wrapped into separate table.
```
vrom911, line 726:
```suggestion
types that should be wrapped into a separate table.
```

src/Toml/Codec/Generic.hs
```diff
@@ -721,6 +722,39 @@ instance (Generic a, GenericCodec (Rep a)) => HasItemCodec (TomlTable a) where
...
+@personCodec@ from the 'TomlTable' comment corresponds to the TOML of
+the following structure:
+
+@
+name = "foo"
+[address]
+    street = \"Bar\"
+    house = 42
+@
+
```
vrom911, line 740: Awesome documentation 💯 🎞️

vrom911 (approved):
Great one 🥌

PR #386, opened 2021-07-15
Allow megaparsec 9.1

felixonmars:
https://hackage.haskell.org/package/megaparsec-9.1.0/changelog

### DIFF
--- a/tomland.cabal
+++ b/tomland.cabal
@@ -120,7 +120,7 @@ library
                      , containers >= 0.5.7 && < 0.7
                      , deepseq ^>= 1.4
                      , hashable >= 1.3.1.0 && < 1.4
-                     , megaparsec >= 7.0.5 && < 9.1
+                     , megaparsec >= 7.0.5 && < 9.2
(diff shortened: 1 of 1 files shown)


vrom911 (approved):
Thanks!

bergmark:
could we get a revision / upload please? Last thing remaining to close https://github.com/commercialhaskell/stackage/issues/6116

chshersh:
@bergmark Why is this blocking Stackage? We removed all Kowainik packages from Stackage previously exactly because we don't want to block Stackage:

* https://github.com/commercialhaskell/stackage/pull/5965

bergmark:
Ah, i wasn't aware of that. We should probably have moved those into the "abandoned [from stackage] packages" section rather than grandfathered so that they would not block us.

bergmark:
there we go https://github.com/commercialhaskell/stackage/commit/4d1760aa36180af9c83ba4e82364b2b9e2100eb5, this should make it clearer for us in the future.

PR #414, opened 2022-10-01
Add HLint

blackheaven:
I'm sorry for the unsolicited PR but I planned to do a hlint contribution on another repository and another contribution for this one and I mixed my folders :s

Anyway, i push it so it could be useful


README.md
```diff
@@ -230,6 +230,14 @@ make it unique:
    Haskell types and `htoml*` libraries use typeclasses-based approach
    via `aeson` library.
 
+## Development
+
+`iris` uses `hlint`, you can run it through:
```
vrom911, line 235: I guess this name is from other PR 🙂

vrom911 (commented):
Thanks for a nice contribution!
Only small thing to fix at the moment 👌🏼

README.md
blackheaven: I'm confused :x

I've pushed a fix.

.hlint.yaml
```diff
@@ -0,0 +1,76 @@
+# HLint configuration file
```
vrom911, line 1: Is this the default file or some additional configurations were added here?
I think it is okay if we use the default one without adding the config file at all  🙂

blackheaven:
@vrom911 is there anything I can do/improve to get it merged?

PR #421, opened 2022-11-29
Allow megaparsec 9.3

felixonmars:
Builds fine and all tests pass.


vrom911 (approved):
Looking at the [CHANGELOG](https://github.com/mrkkrp/megaparsec/blob/master/CHANGELOG.md#megaparsec-930), non of the changes should affect us, so happy to apply. Thanks a lot!

PR #441, opened 2024-05-11
Bump base and containers (to support GHC 9.10)

tomjaguarpaw:


tomjaguarpaw:
Revised at (-r4): https://hackage.haskell.org/package/tomland-1.3.3.2/revisions/

PR #448, opened 2024-12-11
Bump upper bounds for hashable, megaparsec, hedgehog, and hspec-hedgehog

telser:
While here also added GHC 9.10.1 to tested-with as it is in CI.

Would be great if we can get a revision or release with this to get `tomland` back in the stackage nightly.

### DIFF
--- a/tomland.cabal
+++ b/tomland.cabal
@@ -43,6 +43,7 @@ tested-with:         GHC == 8.4.4
                      GHC == 9.4.4
                      GHC == 9.6.3
                      GHC == 9.8.1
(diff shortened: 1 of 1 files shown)


tomjaguarpaw (approved):

tomjaguarpaw:
Thanks! Revision published at: https://hackage.haskell.org/package/tomland-1.3.3.3/revisions/

Please take a look and let me know if there's anything else you need.

telser:
Thank you so much @tomjaguarpaw!

PR #449, opened 2024-12-11
Bump upper bounds for hashable, megaparsec, hedgehog, and hspec-hedgehog

tomjaguarpaw:
While here also added GHC 9.10.1 to tested-with as it is in CI.


tomjaguarpaw:
Alternative to https://github.com/tomjaguarpaw/tomland/pull/new/telser/bump-bounds, because CI was disabled and I don't know how to retrigger it.

tomjaguarpaw:
Making this pull request seems to have retriggered CI on the other pull request, so we can close this one.

